PR #10, opened 2026-07-27
Add direct Chrome DevTools transport and reliable opt-in Ultra workers

floze-the-genius:
## Summary

- replace the Playwright runtime dependency with a small direct Chrome DevTools Protocol client controlling one dedicated ordinary-Chrome profile on a fixed loopback port
- verify login through the current Temporary Chat composer plus a boolean `/api/auth/session` probe, without exporting cookies or logging account/session data
- complete the full native harness for current macOS Seatbelt envelopes, exact text-file reads, and deferred tools loaded through tool search
- add the full-harness-only `chatgpt-web/ultra` route with Extra High coordination, up to three isolated native Codex worker tasks, and a bounded four-lane pool for independently routed Ultra browser turns
- keep normal routes exclusive, Pro read-only, model selection fail-closed, and the configured service tier unchanged
- document the behavior and security boundary in English and Simplified Chinese

## Why

Three concrete failures motivated the transport work:

1. Current Temporary Chat can omit the visible profile/account button even while the session and composer are valid, so profile-control-only authentication checks incorrectly report a signed-out account.
2. Tools returned by Codex tool search were routed through the generic `exec` gateway and could fail as missing JavaScript members instead of reaching their native wire names.
3. The single serialized browser lane cannot support an explicit orchestration mode while preserving normal-turn isolation.

The direct CDP transport also removes the bundled Playwright and Chromium BiDi dependency graph. Chrome keeps its own cookies inside a dedicated profile; the bridge stores only a verification record.

## Ultra worker boundary

Live validation exposed two distinct failures in the original direct-collaboration design:

1. A child without an explicit model recursively inherited `chatgpt-web/ultra` and failed because it had no trusted child browser environment.
2. Pinning the child model was still insufficient. Native `spawn_agent` requires an encrypted collaboration message (`gAAAAA...` on the wire), while the browser bridge can only provide plaintext. All three pinned children therefore failed with an encrypted-content decode error.

The PR now uses the path that completed successfully in live macOS validation:

- Ultra sends a reserved `[[CODEX_ULTRA_WORKER]]` envelope through the existing `codex_exec` tool, so upgraded installs work with the connector's cached MCP schema and require no manual Rescan
- the bridge intercepts the envelope and launches a real native `codex exec --json` process through the outer Codex command tool
- the configured native model is enforced and cannot be overridden by the browser coordinator
- every worker is forced into a read-only sandbox
- no `service_tier` override is sent, so the user's selected/default tier remains authoritative
- each worker call uses a one-second yield so up to three workers start before polling
- verbose child command/reasoning events are kept out of the parent browser context; only `worker_task_id`, exit status, errors, and the final report are returned
- temporary event files are removed inside the Bun filter rather than by a shell cleanup command that Codex safety hooks may reject
- a zero exit without both task ID and final report fails closed
- direct Ultra `spawn_agent` calls are rejected before reaching the incompatible native collaboration boundary

The coordinator contract forbids treating process start as completed work and requires polling every returned session until exit. Fast workers that complete inline are accepted only with the same task-ID, exit-code, and final-report checks.

## Ultra safety boundary

> **Ultra is experimental and explicitly opt-in.** It may fan out to three native Codex workers and permit up to four simultaneous ChatGPT Web turns against one account. Users are responsible for account limits, throttling, workspace policy, and compliance with OpenAI terms.

Ultra does not rotate accounts, change service tiers, retry to evade product limits, or bypass workspace controls. Browser concurrency and native worker fan-out are separate mechanisms. Worker tasks are read-only; the coordinator remains responsible for mutations and verification.

## Compatibility and failure behavior

- Ultra is exposed only in full-harness mode and only when selected manually.
- Browser-only mode continues to expose Instant through Extra High plus account-gated Pro.
- Existing v1 Playwright storage-state files are not imported. Users must complete `login` once to create the dedicated Chrome profile and v2 verification record.
- Normal routes remain serialized and take an exclusive browser lane.
- Every routed turn owns a new Temporary Chat target and closes it in `finally`.
- UI/model drift, unavailable effort levels, invalid session state, missing worker reports, incompatible direct collaboration calls, and blocked command wrappers fail closed.

## Observed UI evidence

On the current Temporary Chat surface, the visible profile/account control can be absent while the composer is visible, the page is the `temporary-chat=true` surface, and `/api/auth/session` reports an authenticated session. The new probe consumes only that boolean result and does not return or log the token or account payload.

The current effort menu and response completion selectors were exercised against visible ordinary Chrome. Selector failure does not fall back to another effort or transport.

## Validation

- `bun run verify`
- TypeScript: clean
- dependency audit: no vulnerabilities
- tests: 108 passed, 0 failed, 459 expectations
- relocatable runtime smoke: passed
- wrapper smoke covers shell quoting, task-ID extraction, final-report extraction, suppression of noisy child output, and fail-closed missing-report behavior
- live macOS orchestration proof: one Ultra coordinator launched three native `gpt-5.6-sol` read-only worker tasks through the cached `codex_exec` schema; all three returned unique task IDs, required markers, compact final reports, and exit code `0`; the parent synthesized all results and completed
- the compact worker path reduced the observed parent accounting from the earlier verbose multi-million-token run to about 230k tokens for the three-worker live smoke
- the original worktree and active account were unchanged after the run

$'GitHub Actions may require maintainer approval because the head branch belongs to an external contributor.'


miuuyy:
Thank you for the substantial work here, and sorry for closing it. The current branch has moved to an integrated desktop launcher and native Codex-harness architecture; merging the custom CDP and Ultra orchestration approach would reduce reliability and duplicate capabilities that Codex already owns natively. I have nearly finished the launcher with macOS, Windows, and Linux support, embedded browser lifecycle management, guided setup and diagnostics, and native MCP integration. I plan to publish it tomorrow. I did identify a few useful isolated ideas in this PR and may reimplement them against the current architecture.

PR #24, opened 2026-08-03
Detect Codex 0.146 sandbox permission profiles

Albro3459:
## Symptom

On Codex CLI 0.146, full harness fails for every sandbox except `danger-full-access`:

```
ChatGPT web turn requires one explicit trusted Codex sandbox mode
ChatGPT web turn is missing cwd in trusted Codex environment context
```

Both are the same bug. Which one shows up depends on whether the fallback text happens to contain a `<cwd>`.

## Cause

`sandboxTypeFromEnvironment` reads `workspace-write` and `read-only` only from the old `<sandbox_mode>` tag. Codex 0.146 dropped that tag in favor of `<permission_profile>`, so the function returns `undefined` and the turn fails.

## Evidence

Captured on codex-cli 0.146.0 with `codex debug prompt-input`, which renders the real prompt without an API call.

`workspace-write`:

```xml
<permission_profile type="managed"><file_system type="restricted"><entry access="read"><special>:root</special></entry><entry access="write"><path>/workspace</path></entry><entry access="write"><special>:slash_tmp</special></entry><entry access="write"><special>:tmpdir</special></entry><entry access="read"><path>/workspace/.git</path></entry></file_system></permission_profile>
```

`read-only`:

```xml
<permission_profile type="managed"><file_system type="restricted"><entry access="read"><special>:root</special></entry></file_system></permission_profile>
```

`danger-full-access` is unchanged: `<permission_profile type="disabled"><file_system type="unrestricted" /></permission_profile>`

## Mapping

| Sandbox | Profile | Signal |
| --- | --- | --- |
| `danger-full-access` | `disabled` + `unrestricted` | unchanged |
| `workspace-write` | `managed` + `restricted` | has `<entry access="write">` |
| `read-only` | `managed` + `restricted` | read entries only |
| external | `external` + `external` | unmapped, fails closed |

The write check only looks inside the matched `<file_system type="restricted">` block, so a stray `access="write"` elsewhere can't make a read-only session writable. The exactly-one check and the old `<sandbox_mode>` path are unchanged.

## Tests

Four cases in `tests/environment.test.ts`: the three modes above, plus `external` failing closed. Fixtures are the captures above, using the existing `root` helper instead of real paths.

`bun run verify` passes.

## Known gap

The binary also contains a self-closing `<file_system type="restricted" />`. I couldn't reproduce it in any config, so I left it unmatched rather than guessing. It fails closed if it does happen.

### DIFF
--- a/launcher/electron/browser-host.cjs
+++ b/launcher/electron/browser-host.cjs
@@ -458,22 +458,26 @@ class BrowserHost {
     return this.snapshot();
   }
 
-  closeTab(tabId) {
-    const tab = this.turnTabs.get(tabId);
-    if (!tab) throw new Error("Browser tab does not exist");
-    this.turnTabs.delete(tabId);
-    if (tab.status === "running") {
+  removeTurnTab(tab, abortRunning) {
+    this.turnTabs.delete(tab.id);
+    if (abortRunning && tab.status === "running") {
       this.closedTurnOwners.set(tab.traceId, tab.helperPid);
       tab.status = "aborted";
     }
     try { this.window.contentView.removeChildView(tab.view); } catch {}
     if (!tab.view.webContents.isDestroyed()) tab.view.webContents.close();
-    if (this.selectedTabId === tabId) {
+    if (this.selectedTabId === tab.id) {
       this.selectedTabId = [...this.turnTabs.keys()].at(-1) || "home";
     }
     this.syncViewVisibility();
     this.publishState?.(this.snapshot());
     this.writeDescriptor();
+  }
+
+  closeTab(tabId) {
+    const tab = this.turnTabs.get(tabId);
+    if (!tab) throw new Error("Browser tab does not exist");
+    this.removeTurnTab(tab, true);
     this.logger.info("browser.tab_closed", { tabId, traceId: tab.traceId, status: tab.status });
     return this.snapshot();
   }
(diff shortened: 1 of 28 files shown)


kvsentry:
@miuuyy +1

I ran into the same failure and traced it to the same function before I found this PR, and the fix looks right to me.

Two things that might be useful:

Your captures came from codex-cli 0.146.0. I got the same three permission_profile shapes on 0.146.0-alpha.9.2 on Windows 11, and they match your test fixtures exactly. So the format looks steady across 0.146 builds rather than being specific to one version.

This is also still broken in the released 1.1.0. The environment.ts there is identical to the 1.0.1 one, and the old code is in the runtime that actually ships. Worth mentioning because the PR now shows a conflict with 1.1.0, so it could look like the release already handled it.

Albro3459:
> @miuuyy +1
> 
> I ran into the same failure and traced it to the same function before I found this PR, and the fix looks right to me.
> 
> Two things that might be useful:
> 
> Your captures came from codex-cli 0.146.0. I got the same three permission_profile shapes on 0.146.0-alpha.9.2 on Windows 11, and they match your test fixtures exactly. So the format looks steady across 0.146 builds rather than being specific to one version.
> 
> This is also still broken in the released 1.1.0. The environment.ts there is identical to the 1.0.1 one, and the old code is in the runtime that actually ships. Worth mentioning because the PR now shows a conflict with 1.1.0, so it could look like the release already handled it.

It appears that @miuuyy fixed this issue in the 1.1.1 release. I merged in the updated main and I am now closing this PR.

PR #26, opened 2026-08-03
Withdrawn

ch9ddar:


miuuyy:
Thanks for the thoughtful work. This is slightly outside the architecture of this bridge.

Codex is the source of truth for task state and history. Every routed ChatGPT Web turn opens a fresh Temporary Chat and receives the complete accumulated Codex context, including system/developer instructions, prior messages, tool results, and attachments. The browser chat is only the transport for that turn; it is not a second memory layer.

Reusing a persistent Project conversation would introduce another mutable source of state and weaken task isolation, deterministic replay/retry behavior, model switching, and subagent independence. If persistent Project conversations are added in the future, they would fit better as a separate, explicitly stateful integration or tool rather than as another Codex model route.

So we do not plan to merge this direction into the current architecture, but thank you for exploring it and for the careful fail-closed implementation work.

PR #34, opened 2026-08-04
Add Install Models (Force) for existing Codex routes

jxnding:
## Summary
- Adds an **Install Models (Force)** button under Setup step 3 that passes `--replace-codex-route`.
- Keeps the normal **Install models** button fail-closed when Codex already has a model route.
- Wires the flag through preload → main → `setupCore({ replace: true })`, with launcher tests.

## Test plan
- [ ] Open Setup with `model_provider` already set in `~/.codex/config.toml`
- [ ] Confirm **Install models** still errors without replace
- [ ] Click **Install Models (Force)** and confirm install succeeds
- [ ] Restart Codex and confirm ChatGPT Web models appear
- [ ] `cd launcher && node --test tests/runtime-host.test.cjs`


miuuyy:
Thanks for putting this together. I am closing this because the reported model_provider="openai" conflict was a false positive: openai is the built-in provider, and the integration now accepts and removes that explicit default without replacing the route. An always-visible Force action is too broad because it can replace a genuinely foreign route without first detecting, showing, and confirming the exact values. The existing --replace-codex-route CLI flag remains available for that deliberate advanced case; a future launcher flow should be conflict-specific and confirmed rather than a permanent Force button.

PR #35, opened 2026-08-04
Back up Codex config.toml before route install

jxnding:
## Summary
- Adds a DeepSeek-style full-file backup of `~/.codex/config.toml` under `~/.codex/backup-codex-chatgpt-web/` (`config.toml` + `manifest.txt`) before the first route install.
- Later setup updates keep the original snapshot instead of overwriting it.
- Verified journal uninstall still fail-closes on drift; the backup directory is discarded after a successful restore. If no journal remains, uninstall can restore from the full-file backup.

## Test plan
- [ ] Fresh install creates `~/.codex/backup-codex-chatgpt-web/config.toml` matching the pre-install file
- [ ] Re-running setup does not overwrite that backup
- [ ] Uninstall restores the prior route and deletes the backup directory
- [ ] Drift fail-closed uninstall tests still pass: `bun test tests/codex-integration.test.ts`

### DIFF
--- a/src/codex-config-backup.ts
+++ b/src/codex-config-backup.ts
@@ -1,0 +1,193 @@
+import { copyFileSync, existsSync, mkdirSync, readFileSync, rmSync, writeFileSync, chmodSync } from "node:fs";
+import { homedir } from "node:os";
+import { join } from "node:path";
+import { expandUserPath } from "./config";
+
+export const CODEX_CONFIG_BACKUP_DIRNAME = "backup-codex-chatgpt-web";
+
+export interface CodexConfigBackupManifest {
+  scriptVersion: string;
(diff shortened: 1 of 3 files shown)


miuuyy:
Thanks for the safety-focused contribution. I cannot accept this full-file restore path because it conflicts with the integration journal architecture and can destroy user changes made after installation. I reproduced the failure: install, add model_reasoning_effort to config.toml, remove the journal, then uninstall; the fallback reports success but restores the old snapshot and silently removes the later setting. The journal intentionally restores only fields owned by this project and fails closed on drift. A full-file backup could be offered as a manual recovery artifact, but it must not become an automatic uninstall source of truth when the journal is absent.

Issue #51, opened 2026-08-06
Stream disconnected bfore completition still persist 2


ForthtuN:
Still geting the same error in 2.0 app

```
{"at":"2026-08-06T00:05:24.827Z","level":"info","event":"launcher.window_created","detail":{"platform":"win32","cdpPort":63881}}
{"at":"2026-08-06T00:05:24.833Z","level":"info","event":"browser.control_started","detail":{"port":63882}}
{"at":"2026-08-06T00:05:24.883Z","level":"error","event":"browser.initialization_failed","detail":{"message":"ERR_ABORTED (-3) loading 'https://chatgpt.com/?temporary-chat=true'"}}
{"at":"2026-08-06T00:05:25.003Z","level":"info","event":"runtime.operation_started","detail":{"name":"bridge-status","args":["route","status"]}}
{"at":"2026-08-06T00:05:25.632Z","level":"info","event":"runtime.stdout","detail":{"operation":"bridge-status","line":"{"}}
{"at":"2026-08-06T00:05:25.634Z","level":"info","event":"runtime.stdout","detail":{"operation":"bridge-status","line":"  \"installed\": true,"}}
{"at":"2026-08-06T00:05:25.637Z","level":"info","event":"runtime.stdout","detail":{"operation":"bridge-status","line":"  \"active\": true,"}}
{"at":"2026-08-06T00:05:25.640Z","level":"info","event":"runtime.stdout","detail":{"operation":"bridge-status","line":"  \"routeUrl\": \"http://127.0.0.1:17841/v1\","}}
{"at":"2026-08-06T00:05:25.642Z","level":"info","event":"runtime.stdout","detail":{"operation":"bridge-status","line":"  \"errors\": []"}}
{"at":"2026-08-06T00:05:25.645Z","level":"info","event":"runtime.stdout","detail":{"operation":"bridge-status","line":"}"}}
{"at":"2026-08-06T00:05:25.653Z","level":"info","event":"runtime.operation_completed","detail":{"name":"bridge-status"}}
{"at":"2026-08-06T00:05:25.816Z","level":"info","event":"browser.authenticated","detail":{"url":"https://chatgpt.com/?temporary-chat=true"}}
{"at":"2026-08-06T00:05:26.752Z","level":"info","event":"runtime.daemon_started","detail":{"pid":424}}
[...]

Issue #46, opened 2026-08-05
Stream disconnected error still persist in new version


ForthtuN:
I don't know if i need make new issue since the other is closed but just to be sure (in case i don't you cn just delete this one) im still getting the issues described here with new version

https://github.com/miuuyy/codex-chatgpt-web/issues/40

miuuyy:
Thanks. The v1.1.3 log you posted stops immediately after startup, before any Project request or “Stream disconnected” error is recorded.  
Please reproduce the failure once, then export the complete Launcher Activity log from startup through the failed Project message. You can redact private paths and project data.

ForthtuN:
Strange, here is full log from start > after error

```
{"at":"2026-08-05T02:13:37.033Z","level":"info","event":"launcher.window_created","detail":{"platform":"win32","cdpPort":59877}}
{"at":"2026-08-05T02:13:37.040Z","level":"info","event":"browser.control_started","detail":{"port":59878}}
{"at":"2026-08-05T02:13:37.097Z","level":"error","event":"browser.initialization_failed","detail":{"message":"ERR_ABORTED (-3) loading 'https://chatgpt.com/?temporary-chat=true'"}}
{"at":"2026-08-05T02:13:37.207Z","level":"info","event":"runtime.operation_started","detail":{"name":"bridge-status","args":["route","status"]}}
{"at":"2026-08-05T02:13:37.900Z","level":"info","event":"runtime.stdout","detail":{"operation":"bridge-status","line":"{"}}
{"at":"2026-08-05T02:13:37.904Z","level":"info","event":"runtime.stdout","detail":{"operation":"bridge-status","line":"  \"installed\": true,"}}
{"at":"2026-08-05T02:13:37.906Z","level":"info","event":"runtime.stdout","detail":{"operation":"bridge-status","line":"  \"active\": true,"}}
{"at":"2026-08-05T02:13:37.909Z","level":"info","event":"runtime.stdout","detail":{"operation":"bridge-status","line":"  \"routeUrl\": \"http://127.0.0.1:17841/v1\","}}
{"at":"2026-08-05T02:13:37.911Z","level":"info","event":"runtime.stdout","detail":{"operation":"bridge-status","line":"  \"errors\": []"}}
{"at":"2026-08-05T02:13:37.914Z","level":"info","event":"runtime.stdout","detail":{"operation":"bridge-status","line":"}"}}
[...]

PR #54, opened 2026-08-06
Retire failed turn sessions so native retries start fresh

DevMello:
A browser turn that fails with anything other than a classified ChatGPT failure (client disconnect, stage timeout, Playwright error) stays cached in the turn registry. Every Codex retry replays the cached error without starting a new browser turn, and each retry resets the TTL, so the turn stays dead until the daemon restarts. This is one mechanism behind "stream disconnected before completion" persisting across retries (#46, #51).

v1.1.3 already retires sessions for retryable ChatGPT failures. This change inverts the condition so that only a non-retryable failure (like an exceeded context window) stays cached and a replayed turn cannot burn another browser attempt. Everything else retires the session so the next retry starts a fresh Temporary Chat.

Added tests for a client disconnect then retry and an unclassified failure then retry (both fail without the fix), plus one pinning that a non-retryable failure stays cached. `bun run verify` green on Windows.

### DIFF
--- a/src/adapters/chatgpt-web/index.ts
+++ b/src/adapters/chatgpt-web/index.ts
@@ -424,13 +424,15 @@ export function createChatGptWebAdapter(provider: CodexProviderConfig): Provider
           }
         });
       } catch (error) {
-        if (error instanceof ChatGptWebAdapterError && error.retryable) {
-          // Reconnects must replay an active/successful browser turn, but retryable terminal
-          // ChatGPT failures need a genuinely new Temporary Chat. Retaining a failed session here
-          // made every native retry replay the same cached error for the registry's full TTL.
-          chatGptTurnSessions.retire(executionKey, session);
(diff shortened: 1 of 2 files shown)


miuuyy:
Thank you for the careful diagnosis and regression cases. I had already started the broader v2.1.0 browser-session lifecycle rework before this PR arrived. Your analysis caught a remaining edge case during the final release audit, and the released implementation now applies the invariant in the new direct-turn architecture: deterministic non-retryable failures remain replayable, while disconnects, unclassified failures, and retryable ChatGPT errors retire the session. Because that implementation landed through #75 and this patch no longer applies cleanly to the v2.1.0 harness, I am closing this PR as superseded rather than merging it. I genuinely appreciate the contribution. Release: https://github.com/miuuyy/codex-chatgpt-web/releases/tag/v2.1.0

Issue #60, opened 2026-08-06
ChatGPT Web thinking-level selector changed to a slider and is no longer detected


cXSHoz:
## Summary

ChatGPT Web appears to have changed the thinking-level selector UI.

Previously, the thinking level was selected from the older menu/dropdown interface. It is now displayed as a horizontal slider inside a popover, and Codex Web GPT no longer appears to recognize the selected thinking level correctly.

In my case, ChatGPT Web is set to **High**, and the composer still displays `High` / `高い`, but the launcher does not correctly detect the setting after this UI change.

## Steps to reproduce

1. Open ChatGPT Web through Codex Web GPT.
2. Click the thinking-level control next to the message composer.
3. Observe that ChatGPT now displays a horizontal slider instead of the previous selection menu.
4. Set the thinking level to **High**.
5. Start or reconnect a Codex Web GPT session.

## Actual behavior

Codex Web GPT does not correctly recognize the currently selected thinking level after the selector was changed to the new slider-based UI.

## Expected behavior

Codex Web GPT should detect the selected thinking level from the new ChatGPT Web UI and continue to recognize values such as:

- Instant
- Medium
- High
- Extra High
- Pro, where available

## Additional information

The visible label next to the composer still shows `High` / `高い`, but the selector popover and its DOM structure appear to have changed.

The current detection logic may still be looking for the previous dropdown or menu-item structure, so the selector parsing may need to be updated for the new slider control.

A screenshot of the new selector UI is attached.

<img width="416" height="243" alt="Image" src="https://github.com/user-attachments/assets/f044b7c5-b322-4812-8392-4f8713b716bf" />

kjverse:
@cXSHoz @miuuyy Same change on Windows with 2.0.0 and a Plus account.

Two things that might save some work:

**The old tier list has not been removed — it is nested.** Clicking `Advanced` in the new popover
gives:

```
Model    GPT-5.6 Sol  ▸
Effort   High         ▸
```

and `Effort ▸` opens the original `Instant 5.5 / Medium / High`, with a check on the active tier —
same entries, same order. So this may be a navigation fix rather than a slider parser:
`CHATGPT_EFFORT_ITEM_SELECTOR` and the `uiEffortIndex` mapping look unchanged, just two clicks
deeper. Caveat: I read that from the rendered UI, not the DOM, so "still `menuitemradio`" is an
inference.

**The symptom is harder than "not detected" — turns fail outright** at the timeout:

```
[...]

Issue #53, opened 2026-08-06
Smoke Test Fail


Compterra:
I keep getting a "Error invoking remote method 'launcher:browser-smoke': Error: ChatGPT effort menu did not expose item index 2 (open=false; itemCount=0)" during the smoke test. i think it is because GPT updated the browser.

mihailmariusiondev:
Same root cause reproduced on 2.0.0 (Linux). The launcher tries to click the composer tone pill (`button[aria-haspopup="menu"][data-tone="neutral"]`) before sending and Playwright fails with `html intercepts pointer events` for the whole 30s timeout (chatgpt-web/light through extra-high, 3/3 attempts, with the launcher window both visible and hidden).

Full error:
```
locator.click: Timeout 30000ms exceeded.
  - locator resolved to <button data-tone="neutral" data-state="closed" aria-haspopup="menu" class="__composer-pill __composer-pill--neutral">…</button>
  - <html dir="ltr" … data-codex-web-gpt-surface="…">…</html> intercepts pointer events
  - element was detached from the DOM, retrying  (ChatGPT re-renders the composer mid-retry)
```
So the new ChatGPT composer (with the tone/intelligence pill) blocks the click with an overlay and re-renders the DOM while retrying. `chatgpt-web/pro` completes the turn only because it skips the tone-pill step (no local tools), and it then replies with the no-workspace warning.

Additional observation: after the failed click the daemon leaves the tab/surface in a bad state and subsequent turns fail the same way.

Compterra:
This is only an issue with Pro Users

PR #61, opened 2026-08-06
Fix ChatGPT effort slider selection

leonardocandiani:
## Summary

* Support ChatGPT's current reasoning effort slider in the Electron launcher smoke test and the Playwright browser worker.
* Preserve the legacy `menuitemradio` path for accounts that still receive the previous UI.
* Fail closed when slider ARIA attributes are missing, malformed, outside the advertised range, or expose more options than the supported model routes.

## Root cause

ChatGPT replaced the reasoning effort radio menu with a slider. The existing selectors waited only for `role="menuitemradio"`, so launcher setup failed with `itemCount=0` before it could write `config.json` or install the model route.

The observed control exposes:

```text
role="slider"
aria-valuemin="0"
aria-valuemax="4"
aria-valuenow="3"
```

The slider is contained by a focusable `role="menuitem"` whose `ArrowLeft` and `ArrowRight` shortcuts update the selected effort.

## Implementation

The patch maps the existing structural effort index to `aria-valuemin + index`, focuses the owning menu item, sends trusted arrow key input, and confirms the final value from the live ARIA state. Pro capability detection uses the same bounded slider range. The old radio menu behavior is unchanged.

Related to #53 and #60.

## Validation

* `bun run verify`
  * 243 runtime tests passed
  * 168 launcher tests passed
  * TypeScript checks, dependency audit, renderer build, runtime bundle build, license generation, and relocatable runtime smoke passed
* The packaged macOS launcher completed its real browser smoke test and received `CODEX WEB GPT READY`.
* A real `chatgpt-web/high` turn completed through the patched runtime and received `CODEX WEB MODEL OK`.

### DIFF
--- a/launcher/electron/browser-host.cjs
+++ b/launcher/electron/browser-host.cjs
@@ -1542,7 +1542,37 @@ class BrowserHost {
   }
 
   async inspectSession(detectPro = false) {
-    return await this.withManualOperation("session inspection", () => this.runSessionInspection(detectPro));
+    return await this.withManualOperation("session inspection", () => detectPro
+      ? this.withInteractiveHomeView(() => this.runSessionInspection(true))
+      : this.runSessionInspection(false));
+  }
+
+  async withInteractiveHomeView(action) {
+    if (browserViewVisible(this.visible, this.surfaceActive, this.boundsReady)) {
+      return await action();
+    }
+    const originalBounds = { ...this.bounds };
+    const [contentWidth, contentHeight] = this.boundsReady
+      ? [originalBounds.width, originalBounds.height]
+      : this.window.getContentSize();
+    const interactiveBounds = this.boundsReady
+      ? originalBounds
+      : { x: 0, y: 0, width: contentWidth, height: contentHeight };
+    const resize = () => this.view.webContents
+      .executeJavaScript("window.dispatchEvent(new Event('resize'))", true)
+      .catch(() => {});
+    this.view.setBounds(interactiveBounds);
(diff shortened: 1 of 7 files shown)


princepal9120:
Reproduced this exactly on macOS (darwin). Login succeeds (browser.authenticated), but the browser smoke test fails every run at the same spot:

```
ChatGPT effort menu did not expose item index 2 (open=false; itemCount=0)
```

Confirmed it is the ChatGPT effort-slider regression, not a config issue. This PR's approach (map structural index to aria-valuemin+index, drive trusted Arrow keys, confirm via live ARIA state, keep the legacy menuitemradio path) matches what the live control exposes. +1 to merge.

For anyone blocked now: build from this branch instead of the packaged release.

```bash
git clone https://github.com/leonardocandiani/codex-chatgpt-web.git
cd codex-chatgpt-web
git checkout codex/fix-chatgpt-effort-slider
bun install   # Bun 1.3.14
bun run app
```

miuuyy:
Thank you for the detailed DOM analysis, implementation, and validation. I had already started the v2.1.0 adaptation for ChatGPT's new effort selector before this PR arrived. The released implementation uses the shared Playwright browser worker for both launcher verification and real turns, supports the ARIA slider and legacy radio path, adds Luna-only handling, and removes the separate CDP activation path. Since that implementation landed through #75, I am closing this PR as superseded. Your work was valuable independent validation, and I appreciate the contribution. Release: https://github.com/miuuyy/codex-chatgpt-web/releases/tag/v2.1.0

Issue #85, opened 2026-08-08
[macOS ARM64][v2.1.0] System Chrome login succeeds but session import fails: no visible composer


Ares960826:
## Summary

On macOS ARM64 with the current v2.1.0 package, authentication succeeds in the dedicated system Chrome window and the normal ChatGPT composer is visibly available, but the launcher rejects the transferred session:

```
Error invoking remote method 'launcher:browser-login': Error:
codex-chatgpt-web: The authenticated ChatGPT page did not produce a visible composer
```

The core setup never advances past **Sign in to ChatGPT**. This appears related to #84, but this report provides a macOS reproduction and package/hash verification.

## Environment

- OS: macOS
- Architecture: ARM64
- Codex Web GPT: v2.1.0
- Browser selected by the launcher: `/Applications/Google Chrome.app/Contents/MacOS/Google Chrome`
- ChatGPT works normally in the user's regular Chrome profile
- API/MCP setup has not been attempted and is not involved

## Package verification

The installed application was compared with the current macOS ARM64 DMG from the v2.1.0 GitHub release.

- Current release DMG SHA-256:
  `d06032b4a47b5d40c2f1b661798fb345708e9f720fa2053795397c0ce455e4f4`
- Installed `app.asar` SHA-256:
  `e5186c1c53461dc8b8ed07e756ed3cfaacfb7fe4721d22ff4544e416a72bb0a3`
- `app.asar` from the current release DMG:
  `e5186c1c53461dc8b8ed07e756ed3cfaacfb7fe4721d22ff4544e416a72bb0a3`

The installed application therefore matches the currently published refreshed v2.1.0 package exactly.

## Steps to reproduce

1. Open Codex Web GPT.
2. Go to **Setup** and click **Open sign in**.
3. The launcher opens its dedicated system Chrome window.
4. Complete ChatGPT authentication.
5. Confirm that ChatGPT is loaded and the message composer is visible.
6. Fully quit the dedicated Chrome instance as instructed.
7. In one attempt, the dedicated Chrome window reopened automatically; quitting it completely again allowed the login operation to finish.
8. The launcher then reports that the authenticated page did not produce a visible composer.

The same result occurs after repeated attempts.

## Observed behavior

The launcher log records:

```
browser.system_login_started
runtime.operation_started name=browser-login
A system Chrome/Chromium window is open. Sign in to ChatGPT, confirm that the composer is visible, then quit this dedicated browser instance completely.
[...]

PR #92, opened 2026-08-09
fix(launcher): make Windows system-browser login compatible with Chrome 151

NcaMoq:
## Summary

- run only the Windows system-browser login flow with Node instead of Bun
- include and validate a Node-compatible CLI bundle in packaged launchers
- reuse the locale-independent ChatGPT composer selector during login verification
- exclude partitioned cookies rather than importing them without their partition key

## Safety

Electron's public cookie API cannot preserve a CHIPS partition key. The launcher never converts a partitioned cookie into an unpartitioned cookie. The imported session must still pass the existing authenticated-composer verification.

## Validation

- `bun run verify`
- generated runtime bundle contains and runs `app/cli-node.cjs`
- Electron Node-mode launch verified on Windows
- manually verified the Launcher login flow on Windows 11 with Chrome 151

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -17,6 +17,7 @@
       "devDependencies": {
         "@types/bun": "1.3.14",
         "@types/turndown": "5.0.5",
+        "tsx": "^4.23.11",
         "typescript": "5.9.3",
       },
     },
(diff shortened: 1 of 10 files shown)


miuuyy:
Thank you for investigating this and for providing the Windows validation details. I am closing this PR because the current implementation is not safe to ship as written.

The release-blocking problem is the generated `app/cli-node.cjs`: `Bun.build({ target: "node", format: "cjs" })` emits a `// @bun @bun-cjs` wrapper rather than a directly executable Node entrypoint. Running the generated artifact with either Node or Electron under `ELECTRON_RUN_AS_NODE=1` exits successfully but does not execute the CLI; even `cli-node.cjs --version` produces no output. The added test verifies only the selected executable, arguments, and environment, so it does not exercise this failure.

The selector portion is directionally correct—login verification should use one locale-independent shared composer contract—but this PR does not add the DOM variants from the active #85 reproduction to `CHATGPT_COMPOSER_SELECTOR`, so it does not cover that reported failure. It also combines the selector change, a second runtime/bundle, and partitioned-cookie policy in one patch, although those are separate fault domains.

A replacement PR would be welcome if it is split and evidence-backed: use one bounded shared composer detector with a regression fixture for the observed DOM, and, if a separate Node login runtime is still necessary, make the packaged entrypoint genuinely Node-executable and add an integration test that invokes the built artifact through Electron Node mode and asserts real CLI output/session-transfer evidence. The partitioned-cookie change should be justified and reviewed separately.

Thanks again for the contribution and the time spent reproducing the Windows behavior.

PR #96, opened 2026-08-09
fix(launcher): repair Electron runtime before startup

jooyal:
## Summary

Electron can report as installed while its platform runtime is incomplete, causing the launcher to fail during startup with:

> Electron failed to install correctly

This PR adds a validated Electron bootstrap and repair flow.

## Changes

- Add `ensure-electron.cjs` to validate:
  - Electron install markers
  - Installed Electron version
  - Platform-specific executable path
  - Actual runtime executable presence
- Automatically rerun Electron’s installer when the runtime is incomplete.
- Run Electron validation before development, standalone start, and packaging.
- Add a reusable launcher setup script that:
  - Installs root dependencies with the frozen lockfile
  - Installs launcher dependencies with the frozen lockfile
  - Repairs/verifies Electron before startup
- Mark Electron as a trusted dependency for Bun installation.
- Add regression tests for missing markers, valid installations, missing executables, and packaging contracts.
- Preserve explicit failure behavior for custom `ELECTRON_OVERRIDE_DIST_PATH` installations.

## Scope

This PR is limited to Electron installation and launcher bootstrap reliability. Default-browser resolution and ChatGPT session-transfer fixes are included in the subsequent PRs.

## Validation

- `bun run typecheck`
- `bun run launcher:typecheck`
- `bun run launcher:test`
- `bun test tests/*.test.ts`
- `git diff --check`

### DIFF
--- a/launcher/package.json
+++ b/launcher/package.json
@@ -8,6 +8,7 @@
   "main": "electron/main.cjs",
   "scripts": {
     "dev": "bun run scripts/dev.cjs",
+    "ensure-electron": "bun run scripts/ensure-electron.cjs",
     "typecheck": "tsc --noEmit",
     "build:renderer": "vite build",
     "build": "bun run typecheck && bun run build:renderer",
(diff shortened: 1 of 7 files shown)


miuuyy:
Thank you for the work and the tests. I’m closing this because the launcher must not repair a development dependency by executing Electron’s installer and downloading into `node_modules` during startup or packaging. Release packages already embed their Electron runtime, while source builds are expected to come from the frozen dependency install and fail explicitly when that toolchain is incomplete. Making startup mutate the build toolchain and depend on a network download would weaken reproducibility and blur the source/package ownership boundary.

The later PRs in this stack also no longer match the current login architecture, so I’m not going to merge this base change.

PR #97, opened 2026-08-09
fix(browser): use the system default browser for login

jooyal:
## Why

The login flow only looked for Chrome and Chromium at fixed paths. On my system, Brave was installed and set as the default browser, but Chrome was not. As a result, login failed even though a compatible Chromium-based browser was available.

## What changed

- Resolve the browser declared as the system’s default HTTP browser.
- Support default-browser discovery on macOS, Windows, and Linux.
- Resolve macOS application bundles when LaunchServices does not provide a direct path.
- Use the same browser resolver in the launcher and release smoke tests.
- Keep the Chromium compatibility check and fail clearly when the default browser is unsupported.
- Improve ChatGPT composer and authentication detection.
- Add regression fixtures for authenticated and logged-out ChatGPT pages.
- Update the relevant documentation and launcher messages.

## Observed DOM Evidence

The authenticated Temporary Chat surface includes:

````html
<textarea aria-label="Chat with ChatGPT" placeholder="Temporary chat">
````

The logged-out surface exposes:

- `Log in`
- `Sign up for free`

## Validation

- `bun run verify`
- Root tests: 305 passed
- Launcher tests: 182 passed
- Root and launcher typechecks passed
- Renderer build passed
- Relocatable runtime smoke test passed

## Relationship to [PR #96](https://github.com/miuuyy/codex-chatgpt-web/pull/96)

This is the second PR in the stack and is based on [PR #96](https://github.com/miuuyy/codex-chatgpt-web/pull/96).

PR #96 fixes the Electron startup/bootstrap problem. This PR addresses the next issue in the flow: finding the available default browser and verifying ChatGPT login correctly. Session-state transfer remains in the following PR.


miuuyy:
Thank you for investigating this. I’m closing the PR because the configured Chrome/Chromium executable is intentionally deterministic and authoritative; the operating-system default may be Safari, Firefox, or another browser that the Chromium login adapter cannot own. Replacing the configured executable with a default-browser lookup would make previously valid installations fail merely because their default browser differs, and it would weaken the launcher’s no-fallback ownership contract.

This branch is also stacked on #96 and still contains the older persistent-profile login path that was removed in v2.1.5/v2.1.6. The current launcher keeps one explicitly configured Chromium process and verifies the captured state without switching browsers.

PR #98, opened 2026-08-09
fix(launcher): handle partitioned cookies during login transfer

jooyal:
## Why

The system-browser login flow transfers browser storage into Electron. Some Chromium browsers return partitioned cookies with a `partitionKey`, but Electron’s cookie API cannot import those cookies. One unsupported cookie could therefore cause the entire session transfer to fail and leave the app stuck checking the saved session.

## What changed

- Skip only partitioned ChatGPT/OpenAI cookies that Electron cannot import.
- Continue importing supported allowlisted cookies and ChatGPT local storage.
- Continue excluding third-party browser state.
- Preserve cleanup and partial-state recovery when session import fails.
- Re-check authentication inside Electron after importing the session.
- Add regression coverage for partitioned cookies and successful session transfer.
- Update the login progress message to refer to the system default browser.

If the imported state is insufficient for authentication, the login flow still fails explicitly rather than claiming success.

## Relationship to PR #97

This is the third PR in the stack and is based on [[PR #97](https://github.com/miuuyy/codex-chatgpt-web/pull/97)](https://github.com/miuuyy/codex-chatgpt-web/pull/97).

PR #97 adds system-default-browser discovery and ChatGPT login detection. This PR fixes the handoff of the browser session into Electron. It does not change model selection or browser discovery.

## Validation

- `bun run verify`
- Root tests: 305 passed
- Launcher tests: 182 passed
- Root and launcher typechecks passed
- Renderer build passed
- Relocatable runtime smoke test passed


miuuyy:
Thank you — the partitioned-cookie diagnosis and the core `partitionKey` handling direction were correct. The narrow fix has now been implemented independently in v2.1.7, with regression coverage for both cases: supported unpartitioned cookies are imported alongside discarded CHIPS entries, while a transfer containing only unsupported partitioned cookies still fails closed.

I’m closing this PR rather than merging it because it is stacked on #96 and #97 and therefore also carries the rejected Electron self-repair, system-default-browser discovery, and obsolete login-flow changes. The v2.1.7 change touches only the cookie-import boundary and its tests.

Issue #108, opened 2026-08-11
Windows launcher: ChatGPT login is captured, but embedded browser fails with ERR_FAILED (-2) and setup never completes


senikv:
Windows launcher: ChatGPT login is captured, but embedded browser fails with ERR_FAILED (-2) and setup never completes

Environment

* OS: Windows 11 x64
* Launcher: latest release installed via the official Windows PowerShell installer
* Browser used by launcher: Google Chrome
* ChatGPT login works correctly in the dedicated Chrome window
* No system proxy configured
* No HTTP_PROXY / HTTPS_PROXY / ALL_PROXY environment variables
* WinHTTP proxy: direct access
* `curl https://chatgpt.com/?temporary-chat=true` reaches Cloudflare successfully and returns a normal 403 challenge response

Problem

The Windows launcher cannot complete initial setup.

When I click Sign in:

1. The launcher opens a dedicated Chrome/Chromium window.
2. I sign in to ChatGPT successfully.
3. The launcher detects the authenticated session.
4. The Chrome window closes automatically.
5. The embedded browser inside Codex Web GPT remains blank.
6. Setup never completes.
7. `~/.codex-chatgpt-web/config.json` is never created.

The Activity log shows that authentication is successfully captured:

browser.system_login_started

A dedicated system Chrome/Chromium window is open. Sign in to ChatGPT and leave it open; transfer continues automatically when the Temporary Chat composer is visible.

Launcher-controlled ChatGPT login captured for private-profile verification.

Immediately afterwards the embedded browser fails:

browser.system_login_failed

ERR_FAILED (-2) loading 'https://chatgpt.com/?temporary-chat=true'

This is reproducible on every login attempt.

Recent full sequence:

2026-08-11T07:36:22.321Z browser.system_login_started

2026-08-11T07:36:22.322Z runtime.operation_started
browser-login
login --launcher-control --chrome
C:\Program Files\Google\Chrome\Application\chrome.exe

2026-08-11T07:36:22.749Z
A dedicated system Chrome/Chromium window is open. Sign in to ChatGPT and leave it open; transfer continues automatically when the Temporary Chat composer is visible.

2026-08-11T07:36:30.423Z
Launcher-controlled ChatGPT login captured for private-profile verification.

2026-08-11T07:37:00.223Z
runtime.operation_completed
browser-login

2026-08-11T07:37:00.489Z
browser.system_login_failed
[...]

PR #110, opened 2026-08-12
Fix macOS login reuse and embedded authentication

eastonplace:
## Summary

Fixes the macOS login loop and app exit/error path when the user is already signed in to ChatGPT in Chrome.

Related to #108.

## Root causes

- ChatGPT authentication rejected the Electron/product-marked user agent during imported-session redirects.
- The login check treated the signed-out guest composer as authenticated.
- The private login profile ignored the current Chrome profile's existing ChatGPT session.
- A verified auth redirect could recursively start another login operation and supersede the in-flight navigation.

## Changes

- Use a Chrome-compatible user agent for managed ChatGPT pages.
- Require `/api/auth/session` user evidence plus a visible composer.
- On macOS, detect Chrome's current profile and seed a disposable login profile with allowlisted ChatGPT/OpenAI cookies.
- Exclude passwords, history, extensions, and Chrome preferences.
- Keep the verified import redirect inside the bounded handoff.
- Clear the Electron-owned ChatGPT session before installing transferred state.
- Delete disposable browser state after capture.

## Verification

- Core tests: 312 passed, 0 failed.
- Launcher tests: 185 passed, 0 failed.
- Core typecheck passed.
- Launcher typecheck passed.
- Renderer production build passed.
- `git diff --check` passed.
- Packaged and installed macOS arm64 v2.1.8.
- Live proof on macOS returned `/api/auth/session` status 200 with `authenticated: true`.
- Live user agent contained Chrome only. It excluded Electron and Codex Web GPT markers.
- The installed app remained open and reused the existing Easton Chrome session without another login.

## Security

The temporary Chrome profile contains only allowlisted `chatgpt.com` and `openai.com` cookie rows. It uses owner-only permissions and is deleted after capture. No password database, browsing history, extension data, or Chrome preferences are copied.

### DIFF
--- a/launcher/electron/browser-host.cjs
+++ b/launcher/electron/browser-host.cjs
@@ -1330,12 +1330,29 @@ class BrowserHost {
       this.setState({ status: "signed-out", message: "Sign in to ChatGPT", authenticated: false, url });
       return this.snapshot();
     }
-    const probe = (contents) => contents.executeJavaScript(`(() => {
+    const probe = (contents) => contents.executeJavaScript(`(async () => {
       const composer = ${visibleElementScript(COMPOSER_SELECTOR)};
-      return { composer: Boolean(composer), readyState: document.readyState };
-    })()`, true).catch(() => ({ composer: false, readyState: "unknown" }));
+      let sessionAuthenticated = false;
+      try {
+        const response = await fetch("/api/auth/session", {
+          cache: "no-store",
+          credentials: "include",
+        });
+        if (response.ok) {
+          const session = await response.json();
+          sessionAuthenticated = Boolean(
+            session
+            && typeof session === "object"
+            && !Array.isArray(session)
+            && session.user
+            && typeof session.user === "object"
+          );
+        }
+      } catch {}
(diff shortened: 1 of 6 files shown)


miuuyy:
Thank you for the careful macOS reproduction, live validation, and for trying to keep copied state allowlisted.

I am closing this because [v2.1.9](https://github.com/miuuyy/codex-chatgpt-web/releases/tag/v2.1.9) removed the dedicated external-Chrome/session-transfer architecture that this PR extends. Sign-in, provider popups, session state, and model turns now stay in one launcher-owned Electron partition again, with server-backed session proof before setup succeeds.

As a result, merging this PR would reintroduce dependencies and risks that no longer exist in the current design: reading Chrome's SQLite cookie database, handling locked/schema/encryption variations, maintaining a disposable copied profile, and overriding the user agent. The valuable part of the diagnosis—the guest composer is not authentication evidence—is preserved in the new embedded flow without copying a user's normal Chrome profile.

Please test the v2.1.9 sign-in flow if you have time, especially on the macOS environment used for this PR. I appreciate the substantial effort and the concrete live evidence even though the final architecture took a different direction.

Issue #86, opened 2026-08-08
[Windows] Bun 1.3.14 repeatedly segfaults and disconnects Codex streams in v2.1.1


cXSHoz:
Summary

On Windows 11 x64, the launcher-owned Responses proxy repeatedly crashes with Bun 1.3.14 while codex-chatgpt-web is running in full mode. When the daemon crashes, the active Codex stream is disconnected.

I see errors in Codex such as:

Reconnecting... /5
stream disconnected before completion: error sending request for url (http://127.0.0.1:17841/v1/responses)

and sometimes:

stream disconnected before completion: Transport error: network error: error decoding response body

The launcher log shows the local daemon crashing at the same time:

Bun v1.3.14 (...) Windows x64
panic(main thread): Segmentation fault at address 0xFFFFFFFFFFFFFFFF
oh no: Bun has crashed. This indicates a bug in Bun, not your code.
runtime.daemon_exited (code 3)

The supervisor then starts a new daemon process. In some cases the immediate recovery also fails because port 17841 is still in use:

codex-chatgpt-web: Failed to start server. Is port 17841 in use?
runtime.daemon_recovery_failed: Responses proxy did not become healthy on 127.0.0.1:17841 within 20000ms

Frequency

In the attached launcher log I counted 46 Bun crash occurrences:

v2.1.0: 27

v2.1.1: 19

So the issue is still reproducible on v2.1.1.

Environment

OS: Windows 11 x64

codex-chatgpt-web: v2.1.1 (also reproduced on v2.1.0)

Bun: 1.3.14

Mode: full

Local Responses endpoint: http://127.0.0.1:17841/v1

Reproduction / observed behavior

Start Codex Web GPT with the bridge route active.

Start a normal Codex task and let it run for a while, including tool calls.

The active stream eventually disconnects and Codex starts reconnecting.

At the same timestamp, the launcher log records a Bun segmentation fault and runtime.daemon_exited.

The supervisor restarts the daemon, sometimes requiring multiple attempts because 17841 remains temporarily occupied.

I have observed the stream-disconnect symptom with both ChatGPT Web models and native Codex models while the bridge route is active.

Expected behavior

The Responses proxy should remain alive for the lifetime of the active stream. If the daemon does fail, recovery should not break the active Codex response or get stuck on the old 17841 listener.

Additional notes

This looks related to #51, but I am filing/reporting it because the same crash pattern persists on v2.1.1.

[...]

PR #111, opened 2026-08-12
fix(browser): tolerate transient answer DOM re-renders

cXSHoz:


miuuyy:
Thank you for isolating this small and fail-closed candidate. It does correspond to the exact `removed a completed text block` error reported in #86, and I verified that the PR's focused tests and typecheck pass on its head.

I am not merging the current draft because the available evidence proves only that an already streamed DOM block disappeared; it does not show that the same segment keys and text returned unchanged within two seconds. The proposed 2,000 ms window is therefore unmeasured, and the new test models the hoped-for recovery sequence rather than a captured browser trace. For an append-only Responses stream, I do not want to relax the committed-prefix invariant based on timing alone.

v2.1.9 does not claim this specific failure is fixed. I am closing this draft so it is not mistaken for a ready patch, but I would be happy to reconsider a focused replacement if a fresh diagnostic records the segment key/text/timestamp sequence showing a transient shrink followed by exact restoration. That evidence would let us choose a real bound rather than guess one. Thank you for keeping the proposed change small.

cXSHoz:
Thank you for reviewing this proposal.

I agree that the current evidence did not prove that the same segment keys and text were restored within the proposed two-second window, so the recovery bound was not sufficiently justified.

If this error recurs, I will provide a fresh diagnostic containing the segment keys, text hashes or lengths, and timestamps around the shrink and recovery sequence, without including prompt contents or private browser data.

Issue #73, opened 2026-08-07
Full Harness loses exec in code_mode_only; Direct mode exposes functions namespace lookup bug


lyndonwest:
## Summary

On Codex CLI 0.146.x with Full Harness enabled, `chatgpt-web/high` inherits `tool_mode: "code_mode_only"`. In this setup the bound harness exposed only:

- `wait`
- `request_user_input`

`exec` was absent, so Full Harness could not access the local filesystem/shell.

This appears related to #72, although in my case `outer_tool_gateway` was `null` because `exec` itself was missing.

## Workaround / additional finding

I temporarily changed `src/model-catalog.ts` from:

```ts
tool_mode: config.mode === "full" ? template.tool_mode : null,
```

to:

```ts
tool_mode: null,
```

After rebuilding, the same Full Harness turn exposed 16 tools, including:

```text
functions__exec_command
functions__write_stdin
functions__view_image
...
```

However, `codex_bind_turn` still reported:

```text
command_tool: null
outer_tool_gateway: null
```

The cause appears to be `exactTool()`:

```ts
return environment.tools.find(tool => !tool.namespace && tool.name === name);
```

Current Codex exposes these tools in the `functions` namespace, so `exec_command` is not recognized.

Changing it to:

```ts
return environment.tools.find(
  tool => (!tool.namespace || tool.namespace === "functions") && tool.name === name
);
```

made the bind correctly report:

```text
tool_count: 16
command_tool: functions__exec_command
outer_tool_gateway: null
```

`codex_exec` then successfully ran `pwd`, `git status`, `git rev-parse`, and read repository files through Full Harness.

## Environment

- codex-chatgpt-web 2.0.0
- Codex CLI 0.146.x
- Linux x64
- Full Harness
- ChatGPT Web High
- PR #61 applied for the current effort-slider UI

There seem to be two related issues:

1. Full Harness inheriting `code_mode_only` results in `exec` disappearing entirely in this setup.
2. In Direct mode, `exactTool()` does not recognize ordinary Codex tools in the `functions` namespace.

lyndonwest:
My setup is a bit out of the ordinary as this is all installed on a remote server that I then access through windows/android chatgpt apps, so potentially this could be my own doing with my custom install. I thought I would raise it anyways just in case it was helpful and not specific to me.

miuuyy:
[...]

PR #115, opened 2026-08-13
fix: preserve full harness tool surface

steveh-works:
## Summary

Restores the regular Codex Responses tool surface for routed ChatGPT Web models in Full Harness mode.

Related to #73.

## Root cause

Full-mode routed rows inherited the native template's `tool_mode: "code_mode_only"`. For a Desktop turn, Codex applied that model metadata before creating the `/v1/responses` request, collapsing the outer registry before the bridge could parse or broker it.

The bridge therefore received only the code-mode helper surface rather than the active outer registry. This is earlier than parser, turn broker, or inventory search logic, so those layers cannot reconstruct the missing tools safely.

## Change

Publish routed ChatGPT Web rows with `tool_mode: null`.

This keeps the native template requirement for selecting a tool-capable source row, while letting Full Harness use the normal Responses surface for MCP namespaces, deferred `tool_search`, and custom outer tools. It does not hard-code MCP servers or rebuild a registry.

The existing V1 subagent-registry regression coverage remains unchanged.

## Evidence

A Desktop Full Harness turn previously exposed an inventory of 2 helper tools. With this catalog behavior, a later bounded acceptance exposed 16 entries, loaded Context7 through `tool_search`, and successfully invoked its library lookup tool.

## Testing

- `bun test tests/model-catalog.test.ts` — 11 passed
- root test suite — 311 passed
- root typecheck — passed
- launcher test suite — 160 passed
- launcher typecheck and production build — passed
- `git diff --check` — passed

`bun audit` did not complete in this environment, so `bun run verify` could not finish; this is recorded rather than treated as passing.

## Limit

An arbitrary `pwd` command remained blocked by the connector's existing open-world/destructive safety policy. That is not an inventory regression and is intentionally outside this change.

### DIFF
--- a/src/model-catalog.ts
+++ b/src/model-catalog.ts
@@ -93,10 +93,9 @@ export function buildChatGptWebModel(
     // provider cannot decrypt that cross-backend payload, so every routed Web model must stay on
     // the native V1 surface where `message` and `fork_context` remain ordinary Codex context.
     multi_agent_version: "v1",
-    // Keep every routed Web model inside Codex's native code-mode and subagent model registry.
-    // Pro's lack of local computer tools is enforced by the adapter runtime; `requiresPro` is only
-    // an account-entitlement gate and must not make the model disappear from native orchestration.
-    tool_mode: config.mode === "full" ? template.tool_mode : null,
+    // Code mode collapses the outer registry into an exec gateway; routed models need the regular
+    // Responses tool surface so MCP namespaces, deferred tool_search, and custom tools reach us.
+    tool_mode: null,
     upgrade: null,
     default_reasoning_level: route.codexEffort,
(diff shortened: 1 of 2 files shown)


miuuyy:
Reviewed against current `main` and the matching Codex runtime contract. In Codex 0.147, catalog `tool_mode` takes precedence over feature flags, and `code_mode_only` removes the ordinary outer tool registry before the bridge receives the request. Publishing the routed Web rows without that forced selector returns control to Codex's authoritative per-turn registry; it does not synthesize tools or add a fallback.

I also verified the exact PR head locally: 311 root tests passed, the focused model/server suites passed 14/14, TypeScript passed, and `git diff --check` is clean. This is a narrow, justified Full Harness fix. Thank you.

PR #116, opened 2026-08-13
fix: support ChatGPT tool approval cards

steveh-works:
## Summary

ChatGPT can render a connector confirmation as `div[data-testid="tool-approval-card"]` rather than an element with `role="dialog"`.

Recognize that exact approval-card surface in addition to the legacy dialog while retaining the existing confirmation text filter.

## Observed DOM evidence

A stalled connector confirmation exposed a `div` with:

- `role: null`
- `data-testid: "tool-approval-card"`

The same confirmation flow did not provide a dialog role, so the previous `[role="dialog"]` locator could not recognize the prompt.

## Behavior

- Legacy dialog confirmations remain supported.
- Manual confirmation still waits for user action and denies on timeout.
- Explicit auto-approval still selects only **Allow once**.
- Unmatched UI still fails closed; this does not broaden model or connector selection.

## Tests

- Added a card-only DOM-shaped fixture. It is invisible to the legacy dialog selector, reproducing the regression.
- `bun test tests/browser-worker-contract.test.ts --test-name-pattern 'approval'` — 4 passed
- `bun test tests/browser-worker-contract.test.ts` — 57 passed
- `bun run verify` — passed
- `git diff --check` — passed

### DIFF
--- a/src/adapters/chatgpt-web/browser-worker.ts
+++ b/src/adapters/chatgpt-web/browser-worker.ts
@@ -168,7 +168,7 @@ export async function resolveChatGptToolConfirmation(
   timeoutMs = CHATGPT_TOOL_CONFIRMATION_TIMEOUT_MS,
   onVisible?: () => Promise<void>,
 ): Promise<boolean> {
-  const dialog = page.locator('[role="dialog"]')
+  const dialog = page.locator('[role="dialog"], [data-testid="tool-approval-card"]')
     .filter({ hasText: `Allow ChatGPT to use ${appName}?` })
     .last();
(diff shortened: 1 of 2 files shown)


miuuyy:
Reviewed against current `main`. The change recognizes the exact observed `data-testid="tool-approval-card"` surface while preserving the existing connector-specific confirmation text, manual-confirmation timeout, fail-closed unmatched behavior, and the exact **Allow once** action for explicit auto-approval.

The exact head passed the full root suite (312/312), the focused browser-worker suite (57/57), TypeScript, and `git diff --check`. Thank you for keeping this surgical.

Issue #99, opened 2026-08-09
Second turn is always a stream disconnected before completion


jesussmile:
Hi, so I am in the latest version macos 2.14, Everything works great on the first run. I removed the old codex native pluging and used the new one only not sure if that is what is causing the issue but can't get the second turn working at all . I have added the log and the  screenshot 

[Archive.zip](https://github.com/user-attachments/files/30876630/Archive.zip)

<img width="966" height="1050" alt="Image" src="https://github.com/user-attachments/assets/5b83e932-4116-48c2-814c-e4e938c8209c" />

miuuyy:
Fixed in [v2.1.5](https://github.com/miuuyy/codex-chatgpt-web/releases/tag/v2.1.5).

The long-prompt writer no longer sends a document-end keyboard shortcut between its verified 100,000-character chunks. It now preserves the native caret left by each text insertion; intermediate attachment checks and the final exact prompt comparison remain enabled. This targets the exact evidence in your archive: expected and actual lengths were equal, while the common prefix ended at character 99,984—the first chunk boundary after the editor markup offset.

Please upgrade to v2.1.5 and retry the same project's second turn. Let us know whether it completes normally. If it still fails, attach fresh v2.1.5 diagnostics from that turn so we can compare the new boundary evidence.

jesussmile:
@miuuyy  its still the same,  V 2.1.7

[launcher.zip](https://github.com/user-attachments/files/30905637/launcher.zip)

<img width="3027" height="1723" alt="Image" src="https://github.com/user-attachments/assets/9a0e3afd-8715-43a9-96f5-8e507fd4362b" />

miuuyy:
A focused candidate fix is available in [v2.1.8](https://github.com/miuuyy/codex-chatgpt-web/releases/tag/v2.1.8).

After each exact intermediate prompt prefix, the browser writer now restores and verifies the live Lexical caret at the document end before the next native edit. Final text equality and cancellation checks remain fail-closed, and chunk boundaries no longer split UTF-16 surrogate pairs.

Please retry the same second-turn/automatic-compaction scenario using the official v2.1.8 package and report whether `prompt_attachment` completes. If it still fails, attach the fresh v2.1.8 prompt diagnostic containing `expected`, `actual`, and `commonPrefix`. I am keeping this issue open pending your confirmation.

jesussmile:
[...]

Issue #91, opened 2026-08-09
Login redirects to external Chrome browser and does not sync back to app


thc81347-hue:
On Windows, when I try to log in to GPT, it redirects me to the Chrome browser; however, in previous versions, the login took place within the app's built-in browser. Currently, even after successfully logging in via Chrome, the app remains unresponsive. I'm not sure if it's an issue with my local data.
<img width="1246" height="748" alt="Image" src="https://github.com/user-attachments/assets/62c3b368-c80b-45ac-b269-10f4ff543319" />

<img width="1660" height="1069" alt="Image" src="https://github.com/user-attachments/assets/5a32ad01-a99a-4f50-b2ae-e0f7fdb89382" />

thc81347-hue:
[v2.1.4] I tried logging in again at https://chatgpt.com but there was a problem

miuuyy:
Opening a dedicated system Chrome/Chromium window is intentional in v2.1.4. It is required so passkeys, security keys, and provider sign-in work in the browser that owns those credentials.

In your screenshot, the dedicated Chrome window is still open, so the launcher is still waiting for the login operation to finish. After the authenticated Temporary Chat composer is visible, fully close that dedicated Chrome window and wait for the launcher to import and verify the session. Do not navigate the embedded launcher browser while login is running; `Browser navigation is locked during ChatGPT login` is the expected safety lock for that operation.

If the launcher remains stuck after the dedicated Chrome process has completely exited, please attach the launcher log lines covering that attempt and confirm whether Chrome actually disappeared from Task Manager. Do not include cookies or authentication tokens.

anime-shed:
#90 

```
{"at":"2026-08-09T12:27:41.698Z","level":"info","event":"launcher.window_created","detail":{"platform":"win32","cdpPort":3714}}
{"at":"2026-08-09T12:27:41.702Z","level":"info","event":"browser.control_started","detail":{"port":3715}}
{"at":"2026-08-09T12:27:42.464Z","level":"info","event":"browser.initialized","detail":{"url":"about:blank#codex-web-gpt-browser-host"}}
{"at":"2026-08-09T12:27:43.014Z","level":"error","event":"browser.system_login_failed","detail":{"message":"ChatGPT browser is already busy with session refresh"}}
[...]

Issue #76, opened 2026-08-08
Unable to increase context beyond ChatGPT's composer transport ceiling


miuuyy:
## Problem

Codex Web GPT compiles each Codex turn into a fresh Temporary Chat so that the current compiled Codex task context, model choice, tool contract, and turn ownership are explicit and isolated. That architecture currently has to place the compiled turn into one ChatGPT composer message.

ChatGPT now applies account- and mode-specific limits to that composer independently of the underlying model context window:

- At or above the visible field limit, the Send button is disabled.
- Immediately below that boundary, the UI can allow submission but the page rejects the payload as too large before any model response is produced.
- Increasing the context number advertised to Codex cannot bypass this browser transport boundary.

The limits below were found empirically and may change when ChatGPT changes its product UI or transport.

## Measured boundaries in v2.1.1

| Account / mode | Practical model context | One-message browser boundary |
| --- | ---: | ---: |
| Free & Go — Luna | 28,000-token conservative browser request budget | Measured inputs were accepted at 25,400 and 28,547 estimated tokens and rejected at 32,283 |
| Plus — Instant | 41,000 tokens | 211,256 composer characters |
| Plus — Medium / High | 90,000 tokens | 1,048,572 composer characters |
| Pro — Instant | 137,000 tokens | 103,000 message tokens / 545,000 composer characters |
| Pro — Medium / High / Extra High | 256,000 tokens | 103,000 message tokens / 1,045,000 composer characters |
| Pro — Pro | 272,000 tokens | 104,000 message tokens / 1,635,000 composer characters |

Luna's underlying model advertises a 1,050,000-token window, but one Free/Go browser request uses a conservative 28,000-token budget derived from the measured accepted and rejected inputs. v2.1.1 therefore uses a private rolling checkpoint and adaptive summarization to carry completed task state between Luna turns. This keeps Luna useful and very fast, but it is not equivalent to sending the full canonical history in one request.

### Attaching the context as a file

This is not reliable enough to be a correctness path:

- ChatGPT does not consistently read an attached file automatically.
- Reading and indexing the file adds latency.
- The model may ignore the attachment, read only part of it, or answer before incorporating it.
- The runtime cannot honestly prove that the full context was consumed.

The project must not silently fall back to this behavior.

### Raising the advertised context window

This only delays local compaction. It cannot enable the disabled Send button or bypass the page-level payload rejection, so it creates false capacity rather than additional usable context.

## Goal

[...]

PR #117, opened 2026-08-13
Harden ChatGPT Web context transport, turn continuity, retry safety, and login

mr-gorjan:
## Summary

This PR hardens the ChatGPT Web bridge in four areas that showed up during real long-running Codex work:

1. **large-context transport and composer integrity**
2. **retry / rate-limit / accepted-turn safety**
3. **compaction and long-task continuity**
4. **system-browser login transfer into the embedded Electron session**

The implementation was developed from the exact v2.1.8 base:

`bda266b45c0e9d73c7a6e932a7c556954f9cea9c`

and the exact tested commit is:

`304e3a08a7faa8a786e90ff417f1ca3f42cee7cf`

It passed automated verification, forced file-transport smoke testing, and a real repeatedly-compacted long-session regression.

Current upstream `main` contains two newer commits after this base, including v2.1.9 login/runtime-recovery work, so several launcher/browser files overlap. I intentionally kept this PR branch at the exact tested implementation rather than resolving those conflicts into code that has not gone through the same live regression.

---

### Reproduction

A real compiled Codex prompt around 112k characters did not survive insertion into the ChatGPT Lexical composer exactly.

Observed during live reduction of insertion sizes:

- expected 111,999 chars, actual 103,999
- with smaller transactions: expected 111,999, actual 111,997
- corruption moved toward the tail as transaction size decreased
- reconnecting did not remove the failure

This was not simply a fixed 104k character ceiling. The browser composer could normalize/rewrite a large prompt while still leaving apparently usable UI state.

### Changes

Prompt insertion now:

- uses bounded native edit transactions
- verifies the exact accumulated prefix after each transaction
- re-anchors the Lexical caret before continuing
- never sends a six-figure native edit
- retries only while the prompt is still unsent

For tool-capable prompts, verified insertion can progressively retry:

`16k -> 8k -> 4k -> 2k`

A retry clears only the unsent composer and rebuilds connector/prompt state.

Once Send is accepted, this retry path is no longer available.

---

## 2. Exact context-file transport for large compiled prompts

For contexts where inline composer transport is not trustworthy, this adds an explicit file transport:

`CODEX_CHATGPT_WEB_CONTEXT_FILE_TRANSPORT=off|large|always`

Modes:

- `off` — current inline behavior
- `large` — use a context document when compiled text is >= 80,000 characters
- `always` — force the transport for testing

Unknown values fail closed.

### Payload

The canonical compiled Codex prompt is converted directly to an in-memory UTF-8 Buffer.

The generated attachment is named:

`codex-context-<sha256-prefix>.txt`

The small inline bootstrap contains:

- exact attachment filename
- full SHA-256
- character count
- UTF-8 byte count
- instruction that the attachment is the complete authoritative compiled Codex prompt
- instruction to read the entire file before reasoning, tools, or answering
- instruction to preserve chronology, roles, ordering, priorities, tool calls/results, compaction state, images, and live capability information

The bootstrap intentionally does **not** duplicate the large context in the composer.

The original complete compiled text still drives token/context accounting.

No temporary source file is created on disk; the attachment originates from memory.

### Important correctness boundary

This PR can verify:

- the exact canonical bytes used to construct the attachment
- its SHA-256
- the exact attachment name
- that ChatGPT accepted the document attachment
- that the intended bootstrap was submitted with it

It does **not** claim that browser automation can cryptographically prove that the model semantically consumed every byte of the attachment.

For that reason this is an explicit transport mode rather than an invisible compatibility fallback.

---

## 3. Document and image attachment safety

The generated context document is uploaded through ChatGPT's document-capable file input directly.

The implementation does not click the native OS upload action, avoiding the GTK/native file picker during automated upload.

Existing image attachment support is retained.

The bridge also enforces ChatGPT's per-message attachment budget:

- inline mode: up to 10 images
- context-file mode: context document + up to 9 images
- exceeding the limit returns an explicit error
- attachments are never silently dropped to make a turn fit

Attachment completion is checked against expected attachment names before Send becomes eligible.

---

## 4. No duplicate Send after accepted submission

One of the most important invariants in this patch is the distinction between:

**unsent prompt failure**

and

**failure after ChatGPT has accepted Send**

The first may be safely reconstructed and retried.

The second must never blindly resend the prompt.

Submission acceptance now uses multiple pieces of browser evidence rather than relying on one fragile DOM transition.

A visible 429 is evaluated only after checking for conclusive late submission evidence.

If ChatGPT already accepted the turn, the bridge continues that turn rather than classifying it as a safe resend.

---

## 5. Shared rate-limit backoff

Adds process-wide ChatGPT Web cooldown state.

Repeated rate-limit hits back off approximately:

`60s -> 120s -> 240s -> 300s`

with small bounded jitter.

While waiting:

- the bridge emits a heartbeat every 10 seconds
- cancellation remains active
- another turn does not immediately hammer the same browser/account again

A successfully accepted submission resets the escalation.

This is retry stabilization, not an attempt to bypass ChatGPT limits.

---

## 6. Accepted-response stall handling

A prompt can be accepted successfully and then stop making visible progress.

Previously this could leave a long-running Codex turn hanging indefinitely.

After accepted Send:

- visible reasoning/output/tool progress keeps the turn healthy
- a 10-minute interval with no visible progress captures diagnostics
- the turn fails explicitly with `browser_response_stalled`
- the error is non-retryable at the browser-send level
- most importantly, the already accepted prompt is **not resent**

---

## 7. Compaction / replay continuity

Long tasks can contain multiple Codex compaction checkpoints.

This patch treats the newest readable compaction checkpoint as authoritative for the task state before that point.

That means:

- old pre-checkpoint instructions remain history/evidence
- they are not resurrected as fresh pending work
- post-checkpoint instructions remain authoritative
- later user overrides still win normally
- tool calls/results retained by the canonical Codex context stay available to the browser turn

This avoids rolling a task backwards when a long session resumes after compaction.

---

## 8. System-browser login -> Electron session repair

During the reproduced login path, a captured authenticated system-browser session was imported into Electron.

The first Temporary Chat navigation could redirect to ChatGPT `/auth` before captured localStorage was restored.

The launcher's normal auth-routing guard intercepted that same-origin redirect, causing Electron `loadURL()` to fail with `ERR_FAILED (-2)` before session restoration completed.

This patch adds a bounded storage-bootstrap state:

- only same-origin `https://chatgpt.com` auth redirects are temporarily allowed
- external identity-provider authentication still routes through the system browser
- the bootstrap state stays active through storage restoration and authenticated Electron verification
- normal auth routing is restored afterward

### Anonymous-composer false positive

A signed-out Temporary Chat page can expose something that looks like a composer.

Composer visibility alone is therefore not sufficient authentication evidence.

System-browser login completion now also requires a ChatGPT authentication session cookie before treating the page as authenticated.

This prevents the login browser from being closed immediately because of an anonymous composer false positive.

---

## Fail-closed properties

The changes preserve the bridge's fail-closed behavior:

- no silent model fallback
- no silent reasoning-level fallback
- no silent context truncation
- no silent attachment dropping
- no blind resend after accepted Send
- no fabricated attachment success
- malformed context-file mode fails explicitly
- excessive attachments fail explicitly
- stalled accepted turns fail explicitly rather than being resent
- no browser/session credentials are intentionally logged
- generated context transport does not write a temporary source file

---

## Tests added / strengthened

Regression coverage includes:

- bounded composer transactions
- simulated large-edit corruption
- caret drift at transaction boundaries
- progressively smaller unsent retries
- submission evidence ordering
- late evidence before rate-limit classification
- process-wide cooldown behavior
- context-file mode parsing
- 80k threshold behavior
- exact context payload bytes
- SHA-addressed context filenames
- bootstrap/context separation
- preservation of tool calls/results in the attached canonical context
- 10-attachment enforcement
- context document + image accounting
- launcher same-origin auth-bootstrap routing
- bootstrap state lifetime through authenticated import

---

### Full repository verification

Final verification on the tested implementation:

- `bun run verify` — **PASS**
- launcher tests — **180 pass / 0 fail**
- launcher build/typecheck — **PASS**
- Vite build — **PASS**
- relocatable runtime smoke — **`RELOCATABLE_RUNTIME_SMOKE_OK`**
- `git diff --check` — **PASS**

Focused browser-worker suite after the final document-upload path:

- **66 pass**
- **0 fail**
- **379 expectations**

Earlier complete root suite during development:

- **321 pass**
- **0 fail**
- **1463 expectations**

A final combined test run after login/session changes completed with:

- **501 pass**
- **0 fail**
- **1463 expectations**

---

### Forced context-file smoke

Launched with:

`CODEX_CHATGPT_WEB_CONTEXT_FILE_TRANSPORT=always`

Observed:

1. `codex-context-<digest>.txt` appeared as a ChatGPT Document attachment
2. no native GTK/Open Files chooser appeared
3. attachment was accepted
4. the small bootstrap was submitted
5. ChatGPT returned the expected smoke response:
   `CODEX WEB GPT READY`

### Real long-session continuity regression

The final live test used an existing long-lived Codex session that had accumulated a large context over many hours and had already undergone repeated automatic compactions.

Before the fix, that same session repeatedly required manual `Continue from the current compacted state.` prompts.

With:

`CODEX_CHATGPT_WEB_CONTEXT_FILE_TRANSPORT=large`

the existing session resumed from its latest compacted state and produced a coherent state-aware answer after approximately 1m38s.

The same session was then given a new, tool-heavy task. That work continued for approximately 22 minutes, crossed several additional automatic compactions, performed multiple tool calls, retained the latest post-compaction objective, and completed coherently.

This exercised more than a single successful response after compaction: the same long-lived session continued evolving through new instructions, tool calls/results, and further compactions.

Observed across the live regression:

- repeated automatic compactions in the same long-lived session
- retained prior task and tool state remained available
- newer post-compaction instructions remained authoritative
- stale historical instructions were not restarted as pending work
- no duplicate accepted prompt was observed
- no task-state rollback was observed
- tool-heavy work continued through additional compactions
- large-context transport completed successfully
- the session remained usable for subsequent long-running work

### Addresses #76 — Unable to increase context beyond ChatGPT's composer transport ceiling

This PR provides an implementation and end-to-end evidence for a larger-context transport.

It deliberately does **not** claim that attachment acceptance is equivalent to cryptographic proof of complete semantic consumption.

The exact bytes and attachment acceptance can be verified; semantic consumption ultimately remains a model/browser-product property.

### Related to #99 — Second turn is always a stream disconnected before completion

This PR strengthens:

- turn lifecycle handling
- heartbeats
- accepted-submission detection
- post-acceptance stall handling
- retry boundaries
- rate-limit behavior

I am not claiming that every #99 reproduction necessarily has the same root cause.

### Related to #91 — Login redirects to external Chrome browser and does not sync back to app

The reproduced system-browser -> Electron session transfer failure is directly in this area.

The patch keeps the external authentication flow but fixes bounded same-origin session-bootstrap redirects and verifies the resulting embedded session.

### Related to #85 — System Chrome login succeeds but session import fails

The patch addresses two reproduced failure modes in this flow:

- same-origin `/auth` interception during Electron storage restoration
- false authentication detection from an anonymous visible composer

Current upstream contains additional v2.1.9 login/runtime recovery work, so this part of the patch will need reconciliation with the newer implementation during integration.

---

## Integration note

This branch intentionally contains the exact implementation that passed the live regression.

Tested base:

`v2.1.8`
`bda266b45c0e9d73c7a6e932a7c556954f9cea9c`

Tested commit:

`304e3a08a7faa8a786e90ff417f1ca3f42cee7cf`

Upstream has since added two commits, including v2.1.9 and #113, and five of the seven changed files overlap.

A trial cherry-pick onto current `main` showed conflicts in those overlapping files.

Rather than silently choosing one side and presenting that untested merge as equivalent, this PR preserves the exact tested implementation and its evidence.

I am happy to reconcile the small current-main delta with the maintainer once the intended behavior of the overlapping v2.1.9 paths is confirmed.

### DIFF
--- a/launcher/electron/browser-host.cjs
+++ b/launcher/electron/browser-host.cjs
@@ -294,6 +294,10 @@ class BrowserHost {
     this.selectedTabId = "home";
     this.manualOperation = null;
     this.loginOperation = null;
+    // During storage-state import, the first navigation may legitimately redirect to ChatGPT's
+    // same-origin /auth surface before captured localStorage has been restored. Allow only that
+    // bootstrap redirect; external identity-provider navigation remains routed to system Chrome.
+    this.systemBrowserLoginStorageBootstrap = false;
     this.cloudflareChallengeRecovery = null;
     this.cloudflareChallengeRecoveryArmed = true;
     this.cloudflareChallengeRecoveryDelayMs = CLOUDFLARE_CHALLENGE_RECOVERY_DELAY_MS;
@@ -590,6 +594,16 @@ class BrowserHost {
 
   routeAuthenticationToSystemBrowser(url) {
     if (!allowedAuthUrl(url)) return false;
+    if (this.systemBrowserLoginStorageBootstrap) {
+      let parsed;
+      try { parsed = new URL(url); } catch {}
+      if (parsed?.origin === CHATGPT_ORIGIN) {
+        this.logger.info("browser.system_login_storage_bootstrap_redirect", {
+          pathname: parsed.pathname,
+        });
+        return false;
+      }
+    }
     void this.openLogin({ force: true }).catch((error) => {
       this.logger.error("browser.system_login_failed", {
         message: error instanceof Error ? error.message : String(error),
@@ -1190,18 +1204,31 @@ class BrowserHost {
         for (const cookie of state.cookies) await contents.session.cookies.set(cookie);
         contents.session.flushStorageData();
         await contents.session.cookies.flushStore();
-        await contents.loadURL(TEMPORARY_CHAT_URL);
         if (state.localStorage.length > 0) {
-          const entries = javaScriptLiteral(state.localStorage);
-          await contents.executeJavaScript(`(() => {
-            if (location.origin !== ${JSON.stringify(CHATGPT_ORIGIN)}) {
-              throw new Error("ChatGPT storage import reached an unexpected origin");
-            }
-            for (const entry of ${entries}) localStorage.setItem(entry.name, entry.value);
-          })()`, true);
+          // A newly captured account can redirect the first Electron navigation to ChatGPT /auth
+          // until its captured localStorage is restored. The normal navigation guard used to
+          // prevent that same-origin redirect, which cancelled loadURL with ERR_FAILED (-2) before
+          // the storage import could run. Permit only the bootstrap redirect while localStorage is
+          // restored, then immediately restore normal auth routing before the authoritative reload.
+          this.systemBrowserLoginStorageBootstrap = true;
+          try {
+            await contents.loadURL(TEMPORARY_CHAT_URL);
+            const entries = javaScriptLiteral(state.localStorage);
+            await contents.executeJavaScript(`(() => {
+              if (location.origin !== ${JSON.stringify(CHATGPT_ORIGIN)}) {
+                throw new Error("ChatGPT storage import reached an unexpected origin");
+              }
+              for (const entry of ${entries}) localStorage.setItem(entry.name, entry.value);
+            })()`, true);
+            await contents.loadURL(TEMPORARY_CHAT_URL);
+            browser = await this.waitForAuthenticated(60_000);
+          } finally {
+            this.systemBrowserLoginStorageBootstrap = false;
+          }
+        } else {
           await contents.loadURL(TEMPORARY_CHAT_URL);
+          browser = await this.waitForAuthenticated(60_000);
         }
-        browser = await this.waitForAuthenticated(60_000);
         if (browser?.authenticated !== true) {
           throw new Error("Imported ChatGPT session did not produce an authenticated Electron composer");
         }
--- a/launcher/tests/browser-host.test.cjs
+++ b/launcher/tests/browser-host.test.cjs
@@ -510,6 +510,115 @@ test("system-browser storage transfer fails closed when only partitioned cookies
   );
 });
 
+test("system-browser login storage bootstrap permits only same-origin ChatGPT auth redirects", async () => {
+  const calls = [];
+  const fixture = Object.assign(Object.create(BrowserHost.prototype), {
+    systemBrowserLoginStorageBootstrap: true,
+    openLogin: async () => calls.push("open-login"),
+    logger: {
+      info: (event, detail) => calls.push(["info", event, detail]),
+      error: (event, detail) => calls.push(["error", event, detail]),
+    },
+  });
+
+  assert.equal(
+    BrowserHost.prototype.routeAuthenticationToSystemBrowser.call(
+      fixture,
+      "https://chatgpt.com/auth/login?next=%2F%3Ftemporary-chat%3Dtrue",
+    ),
+    false,
+  );
+  assert.equal(calls.some((call) => call === "open-login"), false);
+  assert.ok(calls.some((call) => Array.isArray(call)
+    && call[0] === "info"
+    && call[1] === "browser.system_login_storage_bootstrap_redirect"
+    && call[2]?.pathname === "/auth/login"));
+
+  assert.equal(
+    BrowserHost.prototype.routeAuthenticationToSystemBrowser.call(
+      fixture,
+      "https://accounts.google.com/o/oauth2/v2/auth",
+    ),
+    true,
+  );
+  await new Promise((resolve) => setImmediate(resolve));
+  assert.equal(calls.filter((call) => call === "open-login").length, 1);
+
+  fixture.systemBrowserLoginStorageBootstrap = false;
+  assert.equal(
+    BrowserHost.prototype.routeAuthenticationToSystemBrowser.call(
+      fixture,
+      "https://chatgpt.com/login",
+    ),
+    true,
+  );
+  await new Promise((resolve) => setImmediate(resolve));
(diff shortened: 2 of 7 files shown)


miuuyy:
Thank you for the extensive investigation and live evidence. I am not merging this branch as a single patch.

The PR is based on v2.1.8 and conflicts with current `main`. More importantly, its launcher/login portion restores and modifies the dedicated external-Chrome/session-transfer architecture that v2.1.9 intentionally removed after the Windows/macOS regression cluster; login and turns now share the launcher-owned Electron partition. Reintroducing that path would regress the current architecture.

The remaining changes combine several independent contracts—progressively smaller prompt retries, context-file transport, global rate-limit cooldown, post-acceptance stall handling, and compaction interpretation—into one 967-line patch. They need separate review and failure boundaries. In particular, attachment acceptance and a successful live answer do not prove that the model consumed the complete canonical context file, so that transport cannot silently replace the current correctness path.

Please rebase any still-relevant work onto current `main` and split it into narrow PRs with one proven fault domain each. The most useful candidates to isolate are: (1) accepted-turn stall handling without resend, and (2) any reproducible prompt-integrity failure that remains after the current 16k exact-prefix/caret writer. Do not include the removed external-login architecture. I am closing this combined/conflicting PR rather than asking you to resolve it wholesale.

Issue #125, opened 2026-08-17
still plagued with - Error running remote compact task:


jesussmile:
Its actually gotten a lot better since I last used it but this error still remains, it is definitely a compaction issue. Perhaps we can look into opencode and see how they compact long running tasks and take some que from it ?  I am also using a clean pro account x20 to check if I get banned or not.  It is looking pretty good but this issue is extremely persistent on long running tasks 

<img width="563" height="401" alt="Image" src="https://github.com/user-attachments/assets/e74a9ca8-cb22-49c3-890d-31dab3502e7c" />

lyndonwest:
Adding a related data point from our fork. With one specific roughly 54k-character prompt, the first two 16,000-character `Input.insertText` edits succeeded, but the third repeatedly made no observable change. Verification consistently ended at:

```text
expectedChars=47999, actualChars=31999, commonPrefixChars=31999
```

Manual typing into the stalled composer still worked, and `insertText` generally worked for our other prompts, including a previous prompt of roughly 52.5k characters. We could not make this particular prompt reliable through the chunked insertion path.

For our fork, we switched back to one atomic Playwright `fill()` followed by exact full-prompt readback, while retaining bounded native insertion as a fallback. The original failing prompt now works consistently for us, and controlled 80k, 100k, and 150k fills also preserved the exact text. We recognize that older ChatGPT/Electron behavior was the reason `fill()` had previously been avoided; this is only a report of what is working on our current surface.

Our working fork change is documented in [lyndonwest/codex-chatgpt-web#45](https://github.com/lyndonwest/codex-chatgpt-web/pull/45). Related insertion/ownership history is also tracked in #122 and #105.

jesussmile:
@lyndonwest  I tried your fix , no i get the stream disconnect repeatedly
<img width="666" height="521" alt="Image" src="https://github.com/user-attachments/assets/164da6b1-19aa-4a52-b5fc-efbe8e75fc3c" />

lyndonwest:
That's too bad. @miuuyy had changed it from this method because he had found insertText() in chunks to be more reliable, so no suprise that changing it back won't work for everybody, I just wanted to include that the fill() method is working for me at the moment. Hopefully @miuuyy can figure out another solution.

jesussmile:
@lyndonwest so what's working for you? I have a space plus account , I could try it with it. Do you have a fork of it or something? perhaps an executible ?64 bit windows?

[...]

Issue #122, opened 2026-08-15
[macOS] Compaction prompt can fail before submit; manual submit is orphaned and continuation handoff fails


eastonplace:
## Summary

Automatic remote compaction can still fail in the ChatGPT composer before the scripted submission completes. In the latest reproduction, the compact prompt appeared to reach the web composer, but the scripted path failed before it registered/submitted the turn. Manually pressing **Submit** in ChatGPT let the web UI continue, but that manual submission was outside the bridge's tracked turn ownership, so Codex did not register it as the compact result and could not carry it into the continuation thread.

The Codex task then terminated with `systemError` instead of resuming.

This should be tracked separately from open #102. #102 was explicitly classified by the maintainer as a **post-submission** failure where submission was accepted and no assistant response container appeared. This reproduction is a **pre-submit / prompt-attachment ownership** failure, followed by an orphaned manual submission.

It is closely related to #105, which contains the exact `expectedChars=16000, actualChars=0, commonPrefixChars=0` failure class and a macOS v2.1.9 regression report, but #105 is closed.

## Current reproduction: Codex Desktop thread

Thread title:

`Repair and push PR37 secure session`

Thread ID:

`01a005ea-3690-7092-a5bb-ec89c9ce0b3d`

Observed on **2026-08-15** during a long-running local iOS repository task.

The task had already performed substantial source inspection and planning. It reached automatic context compaction multiple times. The terminal failure was not a Git, Swift, BLE, build, test, or repository-command failure.

### Exact error

Both the main attempt and the automatic retry ended with the same compaction error:

```text
Error running remote compact task: stream disconnected before completion: ChatGPT composer did not commit a complete prompt insertion chunk (expectedChars=16000, actualChars=0, commonPrefixChars=0)
```

Recorded turn outcomes:

- Main turn: ran for about 10m 57s, then failed during remote compaction.
- Automatic retry: ran for about 69s, then failed with the same error.
- Thread status after both attempts: `systemError`.

There was no recorded product commit, push, test completion, or final SHA from the task. The failure happened while Codex was trying to compact/continue, not while executing repository work.

## Observed manual-submit behavior

This is the additional behavior that is not captured by the old #105 report:

1. Codex reaches automatic remote compaction.
2. The browser automation prepares/inserts the compact prompt.
3. The script does not successfully complete/register the submission.
4. The prompt is visible in the ChatGPT web app, so I manually press **Submit**.
[...]

Issue #105, opened 2026-08-10
[Windows] Automatic compaction can fail when long prompts cross ChatGPT composer chunk boundaries


cXSHoz:
## Summary

Automatic remote compaction can fail during `prompt_attachment` when a large inline prompt is inserted into the ChatGPT composer in multiple chunks.

The composer readback may report the same total number of UTF-16 code units as expected while differing from the expected prompt near a chunk boundary. The adapter correctly aborts before sending the prompt, but the surrounding operation reports:

```text
stream disconnected before completion
```

The original failure was first observed with launcher v2.1.6. The v2.1.7 source at commit `2804e24a2d05eacc45ac1fc1e9c8333531a8f4cf` still contained the same long-prompt insertion path.

## Environment

- Windows 11 x64
- Codex Web GPT launcher
- v2.1.6 runtime during the original report
- v2.1.7 source baseline
- Electron/Playwright ChatGPT Web Temporary Chat
- Full mode with the `Codex Native2` connector
- Large inline-compiled context
- No prompt files or images involved

## Steps to reproduce

1. Start a full-mode ChatGPT Web turn.
2. Enable the custom connector.
3. Trigger automatic compaction with a prompt large enough to require multiple `Input.insertText` calls.
4. Let the adapter insert the prompt into the composer.
5. Compare the composer readback with the expected prompt after each chunk and before sending.

The issue was reproduced with prompts around 100,000 to 200,000 UTF-16 code units.

## Representative errors

```text
ChatGPT composer did not commit a complete prompt insertion chunk
(expectedChars=100000, actualChars=100000, commonPrefixChars=99998)
```

Other observed variants included:

```text
expectedChars=199989, actualChars=199989, commonPrefixChars=99985
expectedChars=170108, actualChars=170108, commonPrefixChars=99984
expectedChars=31991,  actualChars=31991,  commonPrefixChars=15982
```

These values are JavaScript UTF-16 code-unit counts, not byte counts. Equal expected and actual lengths show that this is not simple truncation. The shorter `commonPrefixChars` indicates that the adapter read back different content near a chunk boundary.

A later trace still reproduced the same-length mismatch with 16,000-unit edits, showing that reducing the chunk size alone was insufficient when the caret was not explicitly recovered between edits.

## Confirmed behavior

- The failure occurs during `prompt_attachment`, before the prompt is sent.
- Failing traces showed no new user or assistant turn at the time of failure.
[...]

PR #130, opened 2026-08-20
fix(browser): tolerate Lexical NBSP prompt readback

BelKed:
## Summary

This fixes one reproducible prompt verification failure during long prompt insertion and automatic compaction.

ChatGPT's Lexical editor can expose some ASCII spaces (`U+0020`) as non-breaking spaces (`U+00A0`) in `textContent` when they are part of a repeated-space run. The inserted prompt is still the same length and the visible text is unchanged, but the adapter currently compares the DOM readback with the expected prompt using exact UTF-16 equality.

That makes a DOM representation detail look like prompt corruption, so the adapter aborts before submission.

This patch accepts that one observed Lexical transformation while keeping the rest of prompt verification strict.

## What I observed

A real automatic compaction failure on macOS produced:

```text
ChatGPT composer did not commit a complete prompt insertion chunk
(expectedChars=16000, actualChars=16000, commonPrefixChars=15818)
```

I added temporary local diagnostics around the first mismatch. The expected prompt contained a run of ASCII spaces:

```text
... U+0043 U+005C U+006E
U+0020 U+0020 U+0020 U+0020 U+0020 ...
```

The DOM readback contained the same-length run with alternating NBSP and ASCII spaces:

```text
... U+0043 U+005C U+006E
U+00A0 U+0020 U+00A0 U+0020 U+00A0 ...
```

The surrounding text was unchanged.

NFC and NFD normalization did not make the strings equal. That is expected because `U+00A0` is not canonically equivalent to `U+0020`.

The temporary diagnostics are not part of this PR.

## Fix

Prompt verification now treats observed `U+00A0` as equivalent to expected `U+0020` only when the expected space belongs to a run of at least two ASCII spaces.

The same comparison is used for:

- intermediate chunk verification
- final prompt verification
- common-prefix calculation used in failure diagnostics

The rule is directional and narrow.

Accepted:

```text
expected: "a  b"
observed: "a<NBSP> b"
```

Rejected:

```text
expected: "a b"
observed: "a<NBSP>b"
```

Also rejected:

```text
expected: "a<NBSP>b"
observed: "a b"
```

Tabs, newlines, missing characters, reordered text, length changes, and normal text mutations still fail verification.

The patch does not modify the prompt before insertion and does not normalize whitespace globally.

## Why only repeated spaces

The failure I captured happened inside a long indentation run. That matches how `contenteditable` editors may preserve consecutive spaces in DOM text.

Treating every ASCII space and NBSP as interchangeable would hide changes that the verifier should catch. Limiting the exception to repeated expected ASCII spaces covers the observed Lexical behavior without weakening normal single-space checks or intentional NBSP content.

## Tests

The regression test reproduces the captured failure with equal-length strings and alternating `U+00A0` / `U+0020` inside a repeated-space run.

It also checks that:

- intermediate chunk verification accepts the Lexical representation
- final prompt verification accepts the same representation
- a single-space `U+0020 -> U+00A0` change is rejected
- `U+00A0 -> U+0020` is rejected
- tab and newline changes are rejected
- text mutations are rejected
- length mismatches are rejected

Validation:

```text
bun install --frozen-lockfile
(cd launcher && bun install --frozen-lockfile)
bun run typecheck
bun test tests/browser-worker-contract.test.ts
bun run verify
```

## Related issues

This is related to:

- #105
- #122
- #125

It fixes the same-length `U+0020 -> U+00A0` prompt readback failure described above.

It does not fix separate failures where the composer is cleared or replaced, such as `expectedChars=16000, actualChars=0`. It also does not address manual-submit continuation reconciliation from #122.

### DIFF
--- a/src/adapters/chatgpt-web/browser-worker.ts
+++ b/src/adapters/chatgpt-web/browser-worker.ts
@@ -748,6 +748,58 @@ export class ChatGptBrowserWorker {
 
   private constructor(private readonly config: ResolvedBrowserConfig) {}
 
+  /**
+   * Lexical/contenteditable may preserve runs of ASCII spaces by exposing some of them as NBSP
+   * through DOM textContent. Treat that DOM-only representation as equivalent only when the
+   * expected U+0020 belongs to a multi-space run. Single spaces, tabs, newlines, intentional
+   * expected NBSP characters, and every other mutation remain exact and fail closed.
+   */
+  private promptCodeUnitEquivalent(
+    expected: string,
+    observed: string,
+    index: number,
+  ): boolean {
+    const expectedUnit = expected[index];
+    const observedUnit = observed[index];
+
+    if (expectedUnit === observedUnit) return true;
+    if (expectedUnit !== " " || observedUnit !== "\u00A0") return false;
+
+    return expected[index - 1] === " " || expected[index + 1] === " ";
+  }
+
+  private promptTextEquivalent(
+    expected: string,
+    observed: string,
+  ): boolean {
+    if (expected.length !== observed.length) return false;
+
+    for (let index = 0; index < expected.length; index += 1) {
+      if (!this.promptCodeUnitEquivalent(expected, observed, index)) {
+        return false;
+      }
+    }
+
+    return true;
+  }
+
+  private promptEquivalentPrefixLength(
+    expected: string,
+    observed: string,
+  ): number {
+    const length = Math.min(expected.length, observed.length);
+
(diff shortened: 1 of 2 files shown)


BelKed:
Manually tested, before it failed after trying to do the first compaction on

```
ChatGPT composer did not commit a complete prompt insertion chunk
(expectedChars=16000, actualChars=16000, commonPrefixChars=15818)
```

Now it's running quite good for about 40 minutes and managed to successfully go through 8 compactions :)

PR #133, opened 2026-08-21
fix(browser): recover effort slider after popover rerender

heathermhuang:
## Summary

- capture the visible effort slider's ARIA range atomically in the browser document
- reopen the effort menu for a bounded retry when ChatGPT replaces or closes the popover
- reuse the same recovery path after keyboard slider movement
- cover the visible-then-disappearing popover lifecycle with a regression test

## Why

ChatGPT can replace or close the effort popover immediately after Playwright observes the slider as visible. The existing locator is filtered to visible elements, so three later `getAttribute` calls can wait 30 seconds on a locator that no longer resolves. This makes launcher smoke tests and normal model selection intermittently fail even though the selector itself is still correct.

The new browser-document snapshot reads all three ARIA attributes in one evaluation. If the popover disappeared first, model selection reopens it for up to five seconds and then fails closed with a structured upstream error.

## Scope boundary

This PR fixes effort-popover selection only. Later Sub2API and BYOK Chat canaries exposed a separate intermittent failure after submission: ChatGPT can remove the active assistant response DOM before the bridge observes terminal completion, producing `ChatGPT response DOM disappeared while the browser turn was active`. That response-lifecycle problem needs a separate regression and follow-up change; it is not evidence that this effort-selector fix failed.

## Verification

- `bun run verify`
  - 315 runtime tests passed
  - 160 launcher tests passed
  - audit, typechecks, launcher build, runtime bundle, notices, and release smoke passed
- live authenticated launcher smoke with the menu initially closed returned `CODEX WEB GPT READY` at High effort

### DIFF
--- a/src/adapters/chatgpt-web/browser-worker.ts
+++ b/src/adapters/chatgpt-web/browser-worker.ts
@@ -43,6 +43,7 @@ import {
   CHATGPT_USER_TURN_SELECTOR,
   detectChatGptAccountCapabilities,
   parseChatGptEffortSliderState,
+  visibleChatGptEffortSliderSnapshot,
 } from "../../chatgpt-session";
 import { loginVerificationMarkerPath } from "../../browser-login";
 import {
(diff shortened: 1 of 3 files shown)


miuuyy:
Thanks for the work. I found a remaining race: the effort popover can disappear after the slider state is captured but before the keypress, while the test only covers disappearance before the snapshot. Since this is a small targeted patch and its core race is still open, I am not going to merge it as-is.

miuuyy:
To be clear, I am not asking you to iterate further on this PR; I am closing it here. Thank you again for taking the time to contribute.

PR #134, opened 2026-08-21
fix(browser): stabilize replaced and mutable response turns

heathermhuang:
## Summary

- re-resolve the active ChatGPT response from conversation-turn identity when the provisional assistant node is replaced
- reject pre-send, user, and generic wrappers until they expose response-specific DOM evidence
- defer append-only final-answer deltas until response-scoped completion and a stable final DOM snapshot are proven
- preserve live reasoning, commentary, heartbeats, tool rounds, cancellation, and five-tab isolation

### Replaced response turns

Failed turns showed one short provisional assistant response followed by loss of the assistant-semantic match while the browser turn remained active. Playwright locators re-evaluate correctly when a replacement stays inside the selector; the unstable boundary was selector membership and index ownership during ChatGPT's conversation-turn replacement.

The worker now snapshots pre-send turn IDs and dynamically selects only a new, non-user turn after the current user boundary. A replacement must expose assistant, status, alert, busy, Markdown, or response-scoped completion evidence before it is considered present.

### Mutable streamed blocks

ChatGPT can revise apparently completed text blocks long after the previous 750 ms commit window, especially after browser-native web search or Python artifact work. Responses text deltas cannot retract content, so committing those blocks early caused the bridge to disconnect when ChatGPT changed them later.

The worker now buffers final-answer Markdown through generation and emits it only after the existing response-scoped copy action and stable completed DOM are proven. Reasoning, commentary, heartbeats, and tool events continue streaming during the turn.

This remains separate from #133, which was scoped to effort-popover rerenders.

## Verification

- focused browser lifecycle suites: 107 passed
- full `bun run verify`: 316 runtime tests and 160 launcher tests passed
- dependency audit, TypeScript checks, production builds, relocatable runtime smoke, and real system-Chrome launch passed
- complex Extra High artifact canary completed with 23,809 Markdown characters and no response-DOM or mutation failure
- streamed High artifact canary: one `response.created`, one `response.completed`, zero `response.failed`, four final text deltas, and one `[DONE]`

## Test plan

- [x] provisional response is selected
- [x] removal yields no stale response
- [x] replacement response is rebound
- [x] pre-send, user, missing-ID, and response-evidence-free turns are rejected
- [x] mutable completed blocks remain replaceable until browser completion
- [x] final Markdown still reproduces the completed answer exactly
- [x] cancellation and five-tab isolation contracts remain green

### DIFF
--- a/src/adapters/chatgpt-web/browser-worker.ts
+++ b/src/adapters/chatgpt-web/browser-worker.ts
@@ -84,6 +84,7 @@ export const CHATGPT_EMPTY_RESPONSE_GRACE_MS = 10_000;
 export const CHATGPT_COMPLETION_ACTION_GRACE_MS = 60_000;
 export const CHATGPT_COMPLETION_SETTLE_MS = 2_000;
 export const CHATGPT_TOOL_CONFIRMATION_TIMEOUT_MS = 60_000;
+const CHATGPT_CONVERSATION_TURN_SELECTOR = '[data-testid^="conversation-turn-"]';
 const CHATGPT_SMOKE_TEXT = "Reply with exactly: CODEX WEB GPT READY";
 const CHATGPT_SMOKE_EXPECTED = "CODEX WEB GPT READY";
 /**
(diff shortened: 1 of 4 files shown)


miuuyy:
Thanks for the work, but buffering the entire final answer until completion removes live response streaming, so this is not a direction we want to take.

PR #132, opened 2026-08-21
feat: add Enhanced Native2 lifecycles, structured compaction, and Claude Code support

Evanlau1798:
## Summary

This change set expands ChatGPT Web from single-turn forwarding into a long-lived native workflow for both Codex and Claude Code. It adds an opt-in Enhanced Web session mode with Native2 tool bridging, structured compaction, retained conversations, multi-agent support, restart recovery, and stronger Windows launcher lifecycle management.

The branch is based on the latest upstream `main`. Although the implementation is broad, the existing Original/legacy Web path remains available and Enhanced mode must be enabled explicitly.

## Major changes

- **Enhanced Web sessions and Native2:** Keep a single `Codex Native2` connector and the existing seven-tool public ABI while adding lazy tool discovery, request-scoped capabilities, ordered tool-result delivery, context archives, and Multi-Agent V2 support.
- **Structured compaction:** Carry one-shot handoffs through the reserved `codex_tool_call` control operation. Active, retained, read-only, and preemptive checkpoint flows use structured submissions while preserving latest-user correctness, complete Web-response proof, and legacy compact compatibility.
- **Codex and Claude Code lifecycles:** Support exact-once steering, long-running continuations, retained root/subagent reuse, agent handoff across compaction, canonical-history rotation, TTL continuation, and full-transcript recovery after a server restart.
- **Browser resilience:** Add stable final Markdown projection evidence, public assistant-turn identity binding, DOM virtualization protection, Lexical caret/NBSP compatibility, same-surface recovery, structured rate-limit and terminal-error classification, and submitted prompt/tool replay prevention.
- **Launcher and Windows runtime:** Add atomic idle-drain restarts, cross-process ownership protection, explicit retained-surface release events, fail-closed child cleanup, persistent browser-helper ownership, and restart/resume behavior that does not depend on a short TTL.
- **Security and diagnostics:** Scope local request and token capabilities, require evidence-backed tool failure claims, redact sensitive diagnostics, and extend browser tab, turn, retained-session, and compaction lifecycle events.
- **Runtime pin:** Move local builds, CI, release workflows, and the Windows baseline runtime to Bun `1.4.0+34cbb9a40` stable.

## Compatibility decisions

- No Native3 connector is introduced; the connector identity remains `Codex Native2`.
- `tools/list` remains at seven tools with the existing public schemas and ABI hash.
- The Original/legacy Web compact path is preserved.
- Enhanced compaction never treats a Markdown marker as the authoritative handoff.
[...]

PR #136, opened 2026-08-21
fix(runtime): harden launcher ownership and Bun baseline

Evanlau1798:
## Summary

This establishes the runtime and Launcher lifecycle foundation needed before the larger browser-session changes from #132 can be reviewed independently.

The patch pins the stable Bun 1.4.0 runtime and its exact Windows baseline revision, makes configuration and Codex route updates byte-preserving and reversible, and hardens Launcher-owned daemon/tunnel restart behavior around drain admission, foreign ownership, stale versions, locked Windows bundles, and state-persistence failures.

It intentionally keeps the current Codex route session-neutral. Retained sessions, context archives, steering, structured compaction, Multi-Agent changes, and Claude Code integration are not included.

## What I observed

The monolithic #132 branch combined several runtime prerequisites with later session behavior. The lifecycle paths also had failure windows where:

- setup shutdown could erase an active marker owned by another Launcher;
- a failed ownership-state write could leave a later runtime generation alive without durable ownership evidence;
- tunnel monitoring could drop its process reference before degraded state was persisted;
- stale release recovery could respawn an incompatible runtime repeatedly;
- Windows replacement could fail after a successful install because the previous bundle remained transiently locked; and
- a restart could race new turn admission between an idleness check and drain acquisition.

The runtime and release metadata also needed one immutable Bun revision across package metadata, CI, release packaging, license generation, and the Windows baseline bundle.

## Fix

Runtime ownership and restart handling now:

- acquires an authenticated atomic `drain-if-idle` transition before shutdown;
- refuses to replace live foreign ownership evidence;
- terminates every still-live managed child whenever ownership persistence fails, including later runtime generations;
- retains tunnel identity until degraded ownership state is safely persisted;
- stops stale-version recovery before spawning and returns to setup;
- tolerates only bounded transient Windows cleanup locks while preserving structural failures; and
- uses the native runtime manager for long-lived tunnel ownership with a 24-hour MCP transport lifetime.

Configuration and distribution handling now:

- preserves UTF-8 BOM, line endings, trailing newline state, and the user's native Codex feature settings;
- migrates older route journals while managing only `openai_base_url`;
- retries Windows atomic renames within a bounded transient-error budget;
- pins `bun@1.4.0+34cbb9a40` consistently across package, workflow, release, bundle, and third-party notice contracts.

## Scope

Included:

- Bun 1.4 stable runtime/release metadata
- runtime and Launcher ownership, drain, restart, rollback, and recovery primitives
- reversible Codex route/configuration persistence
- lifecycle authentication and HTTP turn accounting
- focused process, configuration, packaging, catalog, and cross-platform contract tests

Not included:

- retained ChatGPT Web sessions
- context archive or steering behavior
- structured compaction
- Codex Multi-Agent V2 behavior changes
- Claude Code Messages support

Those contracts remain split into later dependent PRs.

## Tests

Validation completed on Windows with Bun `1.4.0 (34cbb9a40)`:

```text
Root test suite:      319 passed, 0 failed
Launcher test suite: 172 passed, 0 failed, 1 existing Linux-only skip
Root typecheck:       passed
Launcher typecheck:   passed
Root runtime build:   passed
Launcher build:       passed
Branch publish gate:  20/20 passed after fetching origin and upstream
Atomic source restart: passed with verified Launcher/server ownership (5.3 s downtime)
Live Codex catalog:    five routed Web models discovered through Codex 0.147.0
```

Publish provenance:

```text
Base (upstream/main): e2c69d54877c47aa6a34bce72e7536e1c31c9846
Head:                 8a4104ec71e80169e66607eb6d239566c1d620a7
```

Migration evidence follows the canonical commit index: `cfc16923`, `038e58d8`, `25c7c706`, `10e246a6`, and `17bc2aac` were migrated into this PR; only the runtime portion of `1e00784f` was included; and the obsolete canary-pin commits `41077198` and `51fd4c4e` were superseded by the stable Bun baseline. Review-time hardening fixes were kept as focused follow-up commits on the same branch.

The ownership persistence regressions were also repeated five consecutive times without failure. A relocatable runtime bundle smoke passed, the packaged Launcher extracted from the Windows installer reported `packaged=true` and `runtimeVerified=true`, and third-party notices regenerated without changing the working tree.

The Windows package produced during validation had SHA-256:

```text
D010A39B6CFF9BFAA9014DB683A213AEBD332CD985D78B0569776AD79D781489
```

## Limits

The full retained-session / Native2 / Claude lifecycle smoke from the original monolithic branch is not applicable to this foundation PR because those modules and behaviors are intentionally deferred. This PR instead runs the complete repository and Launcher suites plus the runtime, ownership, catalog, packaging, and atomic-restart contracts that are in scope here.


miuuyy:
Thanks for splitting this work out. I found a blocker in the Windows HTTP turn counter: its eagerly drained tee branch can buffer the entire SSE response and report zero active turns before the client has read it; the referenced Bun bug is already fixed in the pinned Bun 1.4.0. I cannot merge a runtime-ownership change with that failure mode.

PR #137, opened 2026-08-21
fix(runtime): harden launcher ownership and Bun baseline

Evanlau1798:
## Summary

This rebuilds the runtime and Launcher lifecycle foundation from the released 3.0.0 baseline before the larger browser-session changes from #132 are reviewed independently.

The patch pins Bun 1.4.0 and its exact Windows baseline revision, makes Codex routing changes byte-preserving and reversible across the upstream v7 journal and the new v8 ownership schema, and hardens Launcher-owned daemon/tunnel restart behavior around drain admission, foreign ownership, stale versions, locked Windows bundles, updater staging, and state-persistence failures.

It also keeps HTTP turn ownership attached to client-side SSE consumption on Windows. A completed upstream source can no longer make the active-turn count reach zero while the client still has unread buffered output.

The current Codex route remains session-neutral. Retained sessions, context archives, steering, structured compaction, Multi-Agent changes, and Claude Code integration are intentionally deferred to later dependent PRs.

## Supersedes

This refresh replaces the previous PR #137 branch history, which was based on pre-3.0.0 `e2c69d5`. The implementation was reconstructed and revalidated on 3.0.0 `e076a29` instead of replaying the old patches across upstream changes.

## What I observed

The monolithic #132 branch combined several runtime prerequisites with later session behavior. The lifecycle and update paths also had failure windows where:

- setup shutdown could adopt or stop a tunnel before rejecting a live marker owned by another Launcher;
- a failed ownership-state write could leave a later runtime generation alive without durable ownership evidence;
- tunnel monitoring could drop its process reference before degraded state was persisted;
- stale release recovery could respawn an incompatible runtime repeatedly;
- Windows replacement could fail after a successful install because the previous bundle remained transiently locked;
- the detached updater worker could be staged without its adjacent BOM-safe JSON dependency;
- a restart could race new turn admission between an idleness check and drain acquisition;
- an eagerly drained Windows `ReadableStream.tee()` observer could release lifecycle ownership before the client consumed the buffered SSE body; and
- upstream 3.0.0 v7 route journals owned only `openai_base_url`, so treating their provider/catalog values as newly managed fields could overwrite legitimate user changes made while disconnected.

The tee workaround targeted an older Bun async-pull teardown defect. The pinned Bun 1.4.0 baseline already contains that fix, so retaining the workaround created a separate ownership bug.

## Fix

Runtime ownership and restart handling now:

- acquires an authenticated atomic `drain-if-idle` transition before shutdown;
- rejects live foreign ownership evidence before any tunnel adoption or stop attempt;
- terminates every still-live managed child whenever ownership persistence fails, including later runtime generations;
- retains tunnel identity until degraded ownership state is safely persisted;
- stops stale-version recovery before spawning and returns to setup;
- tolerates only bounded transient Windows cleanup locks while preserving structural failures;
- uses the native runtime manager for long-lived tunnel ownership with a 24-hour MCP transport lifetime; and
- uses the Bun 1.4 direct reader lifecycle on every platform, releasing HTTP turn ownership only on client-observed completion, read failure, cancellation, or request abort.

Configuration, update, and distribution handling now:

- preserves UTF-8 BOM, line endings, trailing newline state, and the user's native Codex feature settings;
- distinguishes upstream URL-only v7 journals from the new v8 provider/catalog ownership contract;
- migrates active v7 routing only when native routing fields still exactly match their recorded baseline;
- preserves provider/catalog edits made while a legacy v7 route is disconnected;
- stages the detached updater worker together with its JSON helper;
- retries Windows atomic renames within a bounded transient-error budget; and
- pins `bun@1.4.0+34cbb9a40` consistently across package, workflow, release, bundle, and third-party notice contracts.

## Scope

Included:

- Bun 1.4 stable runtime/release metadata
- runtime and Launcher ownership, drain, restart, rollback, and recovery primitives
- reversible Codex route/configuration persistence, including upstream v7 to v8 migration
- detached updater dependency staging
- lifecycle authentication and client-consumption-bound HTTP turn accounting
- focused process, configuration, packaging, catalog, and cross-platform contract tests

Not included:

- retained ChatGPT Web sessions
- context archive or steering behavior
- structured compaction
- Codex Multi-Agent V2 behavior changes
- Claude Code Messages support

Those contracts remain split into later dependent PRs.

## Tests

Validation completed on Windows with Bun `1.4.0 (34cbb9a40)`:

```text
Root test suite:                     353 passed, 0 failed
Launcher test suite:                 194 passed, 0 failed, 1 intentional platform-conditional skip
Codex route integration:              37 passed, 0 failed
Ownership/shutdown race lane:         41 passed, 0 failed, repeated 5 times
Updater/packaging focused lane:        13 passed, 0 failed, 1 intentional platform-conditional skip
Atomic restart script Pester tests:     3 passed, 0 failed
Root typecheck:                       passed
Launcher typecheck:                   passed
Runtime bundle build:                 passed
Relocatable runtime smoke:            passed
Launcher production build:            passed (434 modules)
Native Codex catalog smoke:           passed
Version / Shell / PowerShell gates:   passed
actionlint:                           passed
Bun audit:                            0 vulnerabilities across 106 packages
Third-party notice regeneration:      clean
Branch publish policy:                18 checks passed
PR-01 scope and 500-line gates:       passed (63 changed source/script files checked)
Independent targeted review:         Ready, no findings
```

The route regressions were first exercised against the incorrect v7/v8 ownership behavior and failed before the fixes. The detached updater regression likewise reproduced `MODULE_NOT_FOUND`, and the foreign-owner regression proved that tunnel adoption occurred before rejection. All pass after the bounded fixes.

The full migration audit remains complete:

```text
Original commits:       227 / 227
Changed files:          220 / 220
Production manifest:    131 files
Test manifest:           89 files
Unassigned / missing / low-confidence: 0
```

Publish provenance:

```text
Base (upstream/main): e076a29049a1290a783521a151070a46c6dd4311 (3.0.0)
Head:                 57582c5b941adf62352a736afcb8a3cddd811cea
```

## Limits

The full retained-session / Native2 / Claude lifecycle smoke from the original monolithic branch is not applicable to this foundation PR because those modules and behaviors are intentionally deferred. This PR instead runs the complete repository and Launcher suites plus the runtime, ownership, catalog, packaging, and migration contracts that are in scope here.

The actual packaged Windows installer was not installed during this refresh, avoiding registry/application mutation and UAC. Packaging contracts and the production renderer/runtime builds were validated instead. The atomic restart script's Pester suite passed, but the currently active self-hosted forwarding session was not restarted because doing so from its own active turn would wait on itself; no live-restart claim is made here.


Evanlau1798:
Thank you for reviewing this first foundation PR.

I have completed a migration plan for decomposing the original #132 change set into approximately ten feature-scoped PRs. PR #137 is the first of that sequence. I plan to submit them incrementally, with only one primary PR marked ready for review at a time, and rebase each subsequent branch onto the latest upstream `main` after the preceding dependency is resolved.

Each PR will include:

- an explicit scope and non-goals section;
- the relevant production code and tests migrated from #132;
- focused regression coverage for its failure, timeout, abort, race, and rollback paths;
- the applicable full-suite, typecheck, build, packaging, and runtime validation gates; and
- migration notes identifying which original commits were migrated, split, superseded, or intentionally excluded.

I will use Codex to assist with implementation, review, and regression verification, but I will review the resulting changes myself and remain responsible for the final code and validation evidence.

Some later areas—such as retained sessions, context archives, structured compaction, Multi-Agent V2, and Claude Code support—are still substantial features, but they will be isolated rather than combined into another monolithic PR. I will also adjust the sequencing or scope boundaries if maintainer feedback indicates a more reviewable split.

Thank you for the additional review effort this series will require.

miuuyy:
Thank you for splitting the original work and documenting it carefully. This is still too broad for me to review and merge safely: 67 files and nearly 10,000 changed lines combine runtime ownership, route-journal migration, updater packaging, tunnel lifecycle, HTTP streaming, and a large launcher refactor. v3.0.0 now includes an isolated DEV mode specifically for platform-scale work like this. Please move only the smallest independent changes onto a fresh branch from current main and validate them end to end through DEV mode with reproducible scenarios and evidence. That still does not guarantee acceptance because the scope may remain too large, but it would make me much more willing to review a future focused PR.

PR #142, opened 2026-08-22
Allow custom ChatGPT connector names

### LINKED ISSUE #138: Custom connector name
Having everyone use `Codex Native2` as a name doesn’t seem very wise in terms of account flagging

devnoname120:
## Summary

- let users choose a unique ChatGPT connector name during MCP setup
- preserve configured connector names when reconnecting while retaining the legacy migration safeguards
- use the configured name for runtime setup and browser verification
- update launcher tests and user-facing documentation

## Testing

- `bun run verify`

Fixes #138


miuuyy:
Thank you for implementing this. I am closing it because arbitrary connector identities remove the deterministic MCP schema migration path: ChatGPT caches the schema by connector name, and a future harness contract change must move everyone to one known new identity while retiring the old connector. The implementation is thoughtful, but this product constraint means the feature itself cannot be accepted.

devnoname120:
Thanks, I understand the schema-cache concern. My motivation here is not customization for its own sake. Requiring every installation to use the same uncommon connector name creates a shared account-level fingerprint, which could make bulk identification or flagging easier. I cannot prove that the connector name is currently used for enforcement, but I do not think the project should require that common marker unless it is technically necessary.

I also do not think deterministic migration requires one globally identical identity. The launcher could own a stable per-installation identity together with a schema generation. When the MCP contract changes, it can deterministically create the next identity and retain or recompute the previous one, so the user can be told exactly which connector to retire.

There also seems to be an existing inconsistency: before this PR, the CLI already exposed `--app-name`, core setup persisted arbitrary non-legacy names, and the launcher preserved those configured names. This PR mainly exposed that existing behavior in the launcher. If one global connector identity is a hard invariant, it would need to be enforced in those paths as well.

Would you be open to a launcher-generated per-installation identity instead of unrestricted free-text naming?

devnoname120:
@miuuyy Are you sure that this button doesn’t work? As far as I know it nukes the cache and refreshes all the available tools

image

PR #153, opened 2026-08-23
fix(chatgpt-web): recover trusted environment on resumed threads (missing cwd)

alexiss31:
## Problem

Tool-capable ChatGPT-web turns require a trusted Codex environment. But Codex emits the `<environment_context>` envelope **only when a task starts or its environment changes** — not on every follow-up turn. Codex CLI **0.147+** also stopped stamping `internal_chat_message_metadata_passthrough.turn_id` on replayed history, and assistant output sits between the initial envelope and the active prompt.

Every locator in `rawEnvironmentText` is keyed on either per-item `itemTurnId` **or** on the envelope being structurally adjacent to the active prompt. On a **resumed / long conversation** none of them match, so `extractChatGptTurnEnvironment` throws `MissingTrustedCodexEnvironmentError("cwd")`. Because that first turn never succeeds, `ChatGptThreadEnvironmentStore` can never bootstrap its stored authority — so **every** tool-capable turn fails with:

```
```

### Repro / evidence (v3.0.1, Codex 0.149.0, Windows)

It appears right after enabling the local-tool harness (MCP), because that flips every turn onto the tool-capable path. From `logs/launcher.jsonl`:

```
[chatgpt-web] trusted environment unavailable (thread_id=present, turn_id=present, previous_response_id=none, replay_prefix_items=0, context_messages=96)
```

`thread_id` and `turn_id` are present; the real `<environment_context>` (with `<cwd>` and a `permission_profile` → workspace-write) is in the transcript, just far behind the active prompt under an earlier turn with assistant output in between.

## Fix

Add a final fallback in `rawEnvironmentText`: recover a historical **server-owned** `<environment_context>` when genuine assistant output separates it from the active prompt (evidence of a real resumed transcript), binding it to the authoritative canonical turn metadata via `environmentMatchesCanonicalMetadata`.

Security is preserved:
- Requires a server-owned native item id (a user-typed envelope in the active prompt cannot qualify).
- Requires `hasAssistantOutputBetween(...)` — a genuine resumed transcript, not a fresh injection.
- The primary cwd stays constrained to the trusted metadata `workspaces`, so a user-authored envelope cannot widen filesystem authority. Auxiliary Codex roots are tolerated exactly as on the first-turn path.

## Tests

`bun test tests/environment.test.ts` → **29 pass** (26 existing + 3 new):
- recovers the trusted envelope from replayed history bound to canonical metadata
- cannot trust a replayed cwd outside the canonical workspace metadata
- rejects a replayed envelope that is not a server-owned native item

`bunx tsc --noEmit` clean. No new failures in the full suite (pre-existing env-dependent failures unchanged).

### DIFF
--- a/src/adapters/chatgpt-web/environment.ts
+++ b/src/adapters/chatgpt-web/environment.ts
@@ -316,6 +316,48 @@ function hasAssistantOutputBetween(input: unknown[], startIndex: number, endInde
   return false;
 }
 
+/**
+ * Codex only emits the <environment_context> envelope when a task starts or its environment
+ * changes, not on every follow-up turn, and current builds (0.147+) no longer stamp per-item
+ * turn IDs on replayed history. When a thread is resumed — or the local-tool harness is enabled
+ * after the thread already has turns — the trusted envelope sits far behind the active prompt with
+ * neither structural adjacency nor a per-item turn_id, so `environmentBeforeUser` and the replay
+ * loop above (both keyed on `itemTurnId`) miss it and the thread can never bootstrap its stored
+ * authority. Accept a historical server-owned <environment_context> only when genuine assistant
+ * output separates it from the active prompt (evidence of a real resumed transcript) and its
+ * filesystem authority still agrees with the authoritative canonical turn metadata, so a
+ * user-authored envelope typed into a chat message cannot qualify.
+ */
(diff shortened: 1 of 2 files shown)


alexiss31:
Closing: reverting the local codex-chatgpt-web integration; no longer pursuing this change.

PR #155, opened 2026-08-24
feat: add authenticated external Responses API

guberm:
## Summary
- add an opt-in, separately authenticated OpenAI Responses API listener
- expose only `chatgpt-web/*` models through `/v1/models` and `/v1/responses`
- normalize external requests into server-owned thread/turn metadata and a synthetic read-only environment
- support streaming, `previous_response_id`, and standard function-tool round trips in Full harness mode
- add setup flags, lifecycle integration, security documentation, and focused network/boundary tests

## Security
- disabled by default with a generated 256-bit bearer token
- no native model passthrough, admin, health, search, compaction, or native input items
- bearer credentials are stripped before internal dispatch
- requests without tools are forced to browser-only mode
- caller metadata and workspace authority are discarded

## Verification
- `bun test tests/external-api.test.ts tests/server-lifecycle.test.ts tests/setup-lifecycle.test.ts tests/cli.test.ts` (39 pass)
- `bun run typecheck`
- `bun run build`
- `bun audit` (no vulnerabilities)
- full Windows suite reached 357 pass / 1 unchanged baseline failure in `tests/dev-profile.test.ts` (Windows launcher candidate resolves the real profile instead of the injected test home); the file matches `upstream/main`

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,6 +132,37 @@ rows, but changing them cannot silently change the selected browser model. In Fu
 available effort receives the same turn-bound MCP capability. Pro has no separate restriction or
 reduced tool contract.
 
+## External Responses API
+
+An optional second listener exposes the available `chatgpt-web/*` routes to OpenAI Responses API
+clients. It is disabled by default, uses a separate bearer token, and never exposes native model
+passthrough, health, lifecycle, search, or compaction routes.
+
+Enable it during setup. Omitting `--external-api-token` generates and prints a 256-bit token:
+
+```bash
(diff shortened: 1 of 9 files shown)


miuuyy:
Thanks for the thoughtful and well-tested contribution. I’m closing this because it turns ChatGPT Web into a general-purpose API for other applications, while this launcher is intentionally scoped only to Codex integration.

That boundary keeps the project focused and limits the account and terms risk already inherent in unofficial browser automation; a reusable external API would expand that risk and maintenance surface much further. I appreciate the work, and focused improvements to the Codex integration itself would be much more useful here.

PR #156, opened 2026-08-24
feat(launcher): install external Responses routes safely

heathermhuang:
## Summary
- add a credential-free `codexwebgpt://install/responses` handoff for HTTPS `/v1` gateways
- collect the downstream API key only in a masked local launcher field, with a second explicit confirmation before changing Codex
- journal and restore the prior `config.toml` and `auth.json` with owner-only permissions, cache invalidation, crash reconciliation, and fail-closed drift checks
- reconnect a previously active local Codex Web GPT route when the external provider is disconnected
- register the protocol across packaged macOS/Windows builds and the Linux desktop installer
- document that this installs routing/model discovery only; Full MCP remains separate

## Security boundary
The deep link accepts only a credential-free HTTPS endpoint ending in `/v1` plus a display name. API keys, browser sessions, cookies, and MCP credentials are never accepted in the link or process arguments. A request change after key entry is rejected so a later deep link cannot reuse the secret for another endpoint.

ChatGPT Web allowance remains unofficial browser automation; it is not OpenAI API credit.

## Verification
- `bun run verify`
  - 320 runtime tests
  - 166 launcher tests
  - dependency audit clean
  - TypeScript and renderer builds
  - relocatable runtime smoke
  - real Chrome launch smoke


miuuyy:
Thanks, but installing external Responses routes would turn this launcher into a general API-routing tool for other applications. This project intentionally stays scoped to Codex, so I don’t want to expand it in that direction.

PR #157, opened 2026-08-24
fix(browser): stabilize replaced turns and tool markdown

heathermhuang:
## Summary
- dynamically rebind the active browser turn when ChatGPT removes a provisional assistant turn and mounts its replacement
- exclude Markdown owned by native ChatGPT tool cards from the final-answer stream while preserving their status/commentary trace
- keep the existing incremental `ChatGptMarkdownBuffer` unchanged, including stable completed-block deltas
- preserve cancellation, submission evidence, completion-action requirements, and five-tab isolation

## Why this replaces #134
#134 correctly identified the two browser failures, but its follow-up buffered the entire final answer until completion. This version addresses the maintainer feedback: it does not add a non-incremental mode and does not disable live answer streaming.

The two root causes are separate:
1. ChatGPT can remove a provisional `conversation-turn-*` node while generation remains active. A locator bound once before send never sees the replacement.
2. Native Python/search tool cards contain `.markdown` nodes. When the analysis UI changes shape, those mutable tool nodes can otherwise be classified as final-answer blocks and violate Responses' append-only delta contract.

The tool-card filter is structural and label-independent: it checks current `cot-v5` tool markers only within the Markdown root's local screenshot-content container. Ordinary answer roots still use the existing stable-block buffer.

## Regression and live evidence
- current main reproduced the provisional-turn disappearance on an Extra High HTML-artifact task
- rebinding alone progressed to the Python tool card, then reproduced `ChatGPT changed a completed text block that was already streamed to Codex`
- the complete fix passed the same tool-heavy Extra High task with four live commentary items and one concise 183-character final-answer item; mutable Python/HTML code did not enter the final-answer channel
- one Light multi-section canary completed (585 final-answer characters)
- one High multi-section canary completed (5,426 final-answer characters)

These are tier-specific canaries, not a claim that every tier is universally stable.

## Verification
- `bun run verify` with Bun 1.4.0
  - 351 runtime tests
  - 186 launcher tests
  - dependency audit clean
  - TypeScript and renderer builds
  - relocatable runtime smoke
  - real Chrome launch smoke

### DIFF
--- a/src/adapters/chatgpt-web/browser-worker.ts
+++ b/src/adapters/chatgpt-web/browser-worker.ts
@@ -88,6 +88,8 @@ export const CHATGPT_COMPLETION_SETTLE_MS = 2_000;
 export const CHATGPT_TOOL_CONFIRMATION_TIMEOUT_MS = 60_000;
 export const MAX_CHATGPT_CONNECTOR_TRIGGER_ATTEMPTS = 3;
 const CHATGPT_CONVERSATION_TURN_SELECTOR = '[data-testid^="conversation-turn-"]';
+const CHATGPT_TOOL_MARKER_SELECTOR = '[data-testid^="cot-v5-tool-"], [data-testid^="cot-v5-native-tool-"]';
+const CHATGPT_SCREENSHOT_CONTENT_SELECTOR = "[data-conversation-screenshot-content]";
 const CHATGPT_SMOKE_TEXT = "Reply with exactly: CODEX WEB GPT READY";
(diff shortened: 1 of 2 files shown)


miuuyy:
Thanks. v3.0.2 already shipped the streaming/finalization part of this through the new append-only DOM handling, so this branch now conflicts with the current browser lifecycle.

I’m closing this version. If provisional-turn replacement is still reproducible on v3.0.2, please open a small fresh PR containing only that remaining behavior and a real DOM regression test.

PR #160, opened 2026-08-24
feat: add multi-account and workspace rotation for ChatGPT Web

jamie950315:
## Summary

This PR introduces complete multi-account and multi-workspace rotation for the ChatGPT Web backend.

Each native Codex turn is assigned to one eligible ChatGPT workspace using persistent round-robin selection. Workspaces that are signed out or temporarily cooling down are skipped automatically.

## Features

- manage multiple ChatGPT accounts and workspaces
- share tunnel credentials between workspaces belonging to the same account
- keep separate credentials for distinct ChatGPT accounts
- select workspaces using persistent round-robin rotation
- preserve one account assignment throughout the same native Codex turn
- add, rename, sign in to, and remove accounts and workspaces
- manage accounts through the desktop Settings UI and CLI
- detect expired ChatGPT sessions and mark affected workspaces as signed out
- apply cooldowns after rate limits and session failures
- refresh account configuration without restarting the server
- preserve cookies for every workspace partition during shutdown

### Account and workspace management

Manage multiple ChatGPT accounts and their associated workspaces from the desktop Settings page.

Account and workspace management

### Add another ChatGPT account

Each account can use its own tunnel ID and API key.

Add another ChatGPT account

### Add another workspace to an existing account

Additional workspaces reuse the same ChatGPT account and tunnel credentials.

Add another workspace

## Reliability and safety

- serialize concurrent account, cooldown, and rotation-state updates
- recover safely from abandoned file locks
- prevent simultaneous processes from losing account updates
- prevent retries from switching accounts within the same turn
- avoid reopening a browser session for an account that was just rate-limited or signed out
- preserve account state during setup and Codex configuration updates
- synchronize primary rotation credentials after tunnel changes
- require verified workspace login state instead of relying on historical email data
- keep Settings validation non-blocking
- report incomplete runtime shutdown instead of silently leaving services running

## Compatibility

- includes the latest changes from upstream `main`
- preserves upstream Bigger Context, streaming, MCP continuation, and diagnostics behavior
- legacy single-account configurations continue to use the existing primary ChatGPT session

## Verification

- full `bun run verify`
- dependency audit: no vulnerabilities found
- launcher tests: 219 passed
- repeated concurrent-writer regression tests: 20 passed
- TypeScript checks passed
- renderer and runtime builds passed
- macOS package build passed
- packaged launcher smoke test passed on macOS arm64
- `git diff --check` passed


miuuyy:
Thanks for the work. This PR is too large to accept as-is, but it does contain useful pieces — for example, cooldown detection. Please feel free to extract focused improvements like that into a separate, smaller PR.

The broader multi-account rotation and limit-avoidance functionality makes the tool much more directly aimed at abuse, which this repository does not encourage. For that reason, I’m closing this PR.

jamie950315:
Understood. I’ll keep multi-account rotation in my fork and extract cooldown detection into a smaller, focused PR.

PR #162, opened 2026-08-24
fix: report privacy-safe HTTP stream failures

### LINKED ISSUE #161: Native passthrough stream resets lack lifecycle diagnostics
## Summary

Native Codex passthrough streams can end with `error decoding response body` when the upstream SSE socket is reset after the HTTP response has already started. The Windows lifecycle observer currently catches and discards the corresponding stream error, so the launcher log only receives Bun's raw stack trace and cannot distinguish this case from a daemon crash, client cancellation, or a pre-response fetch failure.

This is separate from #86:

- Bun is stable `1.4.0` (`1.4.0+34cbb9a40`).
- The launcher-owned daemon remains alive and retains the same PID.
- The failing URL is the native passthrough endpoint, not a ChatGPT Web browser turn.
- The underlying error is an upstream socket reset after response headers were accepted.

## Observed behavior

Sanitized launcher excerpts repeatedly show the same failure shape:

```text
TypeError: The socket connection was closed unexpectedly.
  path: "https://chatgpt.com/backend-api/codex/responses",
  errno: 0,
  code: "ECONNRESET"
      at async <anonymous> (.../src/server.ts:153:33)
```

The corresponding Codex client error is:

```text
stream disconnected before completion: Transport error: network error: error decoding response body
```

Four occurrences were observed in one launcher log on 2026-08-24. None was accompanied by `runtime.daemon_exited`, a Bun panic, or a process restart.

The referenced source location is the eagerly consumed lifecycle branch created by `response.body.tee()` in `HttpTurnCounter.track()`. Its `catch` block intentionally leaves the client branch responsible for delivering the stream failure, but it also discards the only local opportunity to record whether any chunks had already crossed the proxy.

## Expected behavior

The proxy should preserve the existing end-to-end failure semantics, while emitting one privacy-safe diagnostic that distinguishes:

- a stream failure after response headers;
- whether any chunks/bytes were observed;
- client or lifecycle cancellation;
- the normalized error name/code.

The diagnostic must not include request, response, or authorization content.

## Safety boundary

This issue does **not** propose automatically replaying the Responses request. Once any part of an SSE response may have been delivered, replay could duplicate model work or tool side effects. The minimal correction is observability only; retry policy can be considered separately for failures proven to occur before any response bytes are delivered.

## Proposed minimal change

- Count chunks and bytes in the existing tracked response stream.
- Emit one sanitized diagnostic when either the direct stream reader or the Windows lifecycle observer fails.
- Keep the original stream error visible to the client.
- Add regression coverage proving that diagnostics do not buffer, replace, or convert the SSE stream and that cancellation remains distinct from an upstream failure.

## Environment

- Windows 11 x64
- codex-chatgpt-web behavior confirmed unchanged on upstream `v3.0.2`
- Bun `1.4.0` / revision `1.4.0+34cbb9a40`
- Native Codex Responses passthrough

Evanlau1798:
## Summary

- record privacy-safe chunk and byte counts when a tracked response stream fails
- distinguish direct readers from the Windows lifecycle tee observer
- suppress failure diagnostics for locally aborted/cancelled turns
- preserve the original stream error, backpressure, and cancellation behavior

Fixes #161.

## Why

When an upstream native Codex SSE socket resets after response headers, the client correctly receives a stream error, but the Windows lifecycle observer previously discarded its local failure evidence. The launcher therefore had only Bun's raw stack trace and could not tell whether the failure happened before or after response bytes crossed the proxy.

The new diagnostic contains only:

- reader stage and platform
- chunk and byte counts
- sanitized error name and code

It never records the error message, request, response body, URL, or authorization data. It also does not retry a partial Responses stream.

## Verification

- `bun test tests/server-lifecycle.test.ts`: 20 passed
- root suite: 360 passed; 1 pre-existing `installed launcher discovery has explicit platform candidates` failure, independently reproduced on an unmodified `v3.0.2` worktree
- launcher suite: 188 passed, 1 platform skip
- root TypeScript check: passed
- launcher TypeScript check: passed
- runtime bundle build: passed

### DIFF
--- a/src/http-stream-diagnostics.ts
+++ b/src/http-stream-diagnostics.ts
@@ -1,0 +1,38 @@
+export interface HttpStreamFailureDiagnostic {
+  stage: "direct" | "lifecycle";
+  platform: NodeJS.Platform;
+  chunks: number;
+  bytes: number;
+  errorName: string;
+  errorCode: string;
+}
+
+export type HttpStreamFailureReporter = (failure: HttpStreamFailureDiagnostic) => void;
+
+function safeErrorField(value: unknown, fallback: string): string {
+  return typeof value === "string" && /^[A-Za-z0-9_.-]{1,64}$/.test(value) ? value : fallback;
+}
+
+export function httpStreamFailureDiagnostic(
+  error: unknown,
+  stage: HttpStreamFailureDiagnostic["stage"],
+  platform: NodeJS.Platform,
+  chunks: number,
+  bytes: number,
+): HttpStreamFailureDiagnostic {
+  const candidate = error && typeof error === "object"
+    ? error as { name?: unknown; code?: unknown }
+    : {};
+  return {
+    stage,
+    platform,
+    chunks,
+    bytes,
+    errorName: safeErrorField(candidate.name, "Error"),
+    errorCode: safeErrorField(candidate.code, "unknown"),
+  };
+}
+
+export const reportHttpStreamFailure: HttpStreamFailureReporter = failure => {
+  console.warn(`[http-stream] tracked response failed ${JSON.stringify(failure)}`);
+};
--- a/src/server.ts
+++ b/src/server.ts
@@ -9,6 +9,11 @@ import type { AppConfig } from "./config";
 import { providerConfig } from "./config";
 import { AsyncEventQueue } from "./event-queue";
 import { readJsonRequestBody } from "./http-body";
+import {
+  httpStreamFailureDiagnostic,
+  reportHttpStreamFailure,
+  type HttpStreamFailureReporter,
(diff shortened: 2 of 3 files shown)


miuuyy:
Thank you — the report, tests, and privacy constraints were genuinely useful.

We already had the privacy-safe launcher export and overlapping uncommitted lifecycle work, so I implemented the same diagnostic goal in the current `HttpTurnCounter` path rather than merging a second diagnostics layer. The tailored version keeps the existing stream and cancellation semantics and records only endpoint/stage/counts plus sanitized error identifiers.

Closing this PR to avoid two competing implementations, not because the underlying observation was wrong.

PR #165, opened 2026-08-24
fix: reject partial launcher shutdown

jamie950315:
## Summary

- validate the runtime supervisor result before committing launcher exit
- keep the app open when forced cleanup is only partially successful
- preserve all cleanup failures in the user-visible error
- keep this change independent from accounts and rotation

Extracted as a focused follow-up from #160.

## Verification

- `npx -y bun@1.4.0 run verify`
- `npx -y bun@1.4.0 run app:package`
- `npx -y bun@1.4.0 run app:smoke`


miuuyy:
Thanks, but Command-Q is intentionally an emergency exit. A partial runtime cleanup must be recorded, but it must not make the launcher impossible to quit, so this would reintroduce behavior we deliberately removed.

PR #173, opened 2026-08-26
fix: detect ChatGPT unusual activity blocks

naakto14:
## Problem

ChatGPT sometimes terminates a browser turn with a visible message like:

> Unusual activity has been detected from your device. Try again later. (...)

In v3.0.3 this terminal state is not classified by the browser adapter. Because it is neither the generic `Something went wrong` alert nor a normal completed response, the worker can continue waiting until a generic browser timeout instead of returning the real cause promptly.

## What this changes

- Adds a dedicated detector for the `Unusual activity has been detected from your device` terminal alert.
- Accepts the optional parenthesized request/incident identifier that ChatGPT may append to the message.
- Checks for this state in both places where the turn can otherwise remain pending:
  - while waiting for ChatGPT to accept the submitted prompt;
  - while polling the active assistant response.
- Converts the alert into a structured retryable adapter error:
  - HTTP status `429`
  - `errorType: "rate_limit_error"`
  - `code: "unusual_activity_detected"`
  - `retryable: true`

## Why 429 / retryable

The UI explicitly asks the user to try again later and represents a temporary request block rather than an authentication failure, malformed request, or permanent connector error. Classifying it separately also preserves the original cause instead of collapsing it into a generic timeout or upstream 502.

## Scope and safety

The matcher is intentionally narrow: it requires the known `Unusual activity...` sentence followed by `Try again later.`. Existing handling for session failures, subscription failures, generic terminal errors, and normal responses is unchanged.

No automatic retry loop is added here; this PR only makes the terminal condition observable to the existing outer retry/error handling.

## Tests

Adds a contract test using the real alert shape, including a parenthesized identifier, and verifies the exact structured error classification.

The orchestration changes are also covered by source-level contract checks to ensure the detector is called during both submission acceptance and response polling.

## Verification

- `bun test tests/browser-worker-contract.test.ts` — 71 passed, 0 failed
- `git diff --check`

### DIFF
--- a/src/adapters/chatgpt-web/browser-worker.ts
+++ b/src/adapters/chatgpt-web/browser-worker.ts
@@ -234,6 +234,18 @@ export async function throwIfChatGptTerminalErrorAlert(scope: ChatGptTextScope):
   );
 }
 
+const chatGptUnusualActivityAlert = (scope: ChatGptTextScope): Locator => scope
+  .getByText(/Unusual activity has been detected from your device\.[\s\S]*Try again later\.(?:\s*\([^)]+\))?/i)
+  .last();
+
+export async function throwIfChatGptUnusualActivityAlert(scope: ChatGptTextScope): Promise<void> {
+  if (!await chatGptUnusualActivityAlert(scope).isVisible().catch(() => false)) return;
+  throw new ChatGptWebAdapterError(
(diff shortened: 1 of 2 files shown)


miuuyy (changes requested):
Thanks for the focused detector. I’m declining this PR because marking the block as retryable allows the entire turn to be resent up to three times after ChatGPT reports unusual activity, which is unsafe for the account.

PR #174, opened 2026-08-26
fix: tolerate markerless ChatGPT completions

naakto14:
## Problem

v3.0.3 treats ChatGPT's response-scoped copy action as a required completion marker. In practice, ChatGPT can finish rendering a valid non-empty answer while that action is temporarily absent or omitted by the current DOM shape.

That leaves the adapter in a bad state: generation has stopped and the final answer is visible, but the turn is never accepted as complete because the copy-action marker is missing. The worker then keeps waiting and may eventually report a DOM/timeout failure for a response that was already complete.

## What this changes

A stopped, non-empty assistant response can now become a completion candidate even when the response-scoped copy action is absent.

The copy action is still useful evidence, but it is no longer the single required completion signal:

- with the copy action present, the existing fast settle window remains `2s`;
- without the copy action, the response must remain stable for a longer `5s` window before completion is accepted.

## Stability safeguards

Markerless completion is deliberately conservative:

- the response DOM must exist;
- generation must be stopped;
- the response text must be non-empty;
- the candidate signature includes both rendered text and HTML;
- any text/HTML change resets the stability timer;
- generation becoming active again invalidates the candidate and resets the timer.

This means the fallback does not accept a still-streaming or mutating response simply because the copy button is missing.

The existing DOM-health checks for vanished responses and empty completions remain in place.

## Why keep two settle windows

The copy action is a strong ChatGPT-owned completion signal, so the existing `2s` settling behavior is preserved when it is available. A markerless response has weaker evidence, so it receives a longer `5s` stability requirement to avoid trading the timeout bug for premature completion.

## Tests

Adds focused coverage for all three important cases:

- a stable markerless final answer is accepted only after the longer window;
- markerless output is not accepted while generation is active or while content is changing;
- responses with the normal copy action still use the existing shorter settle window.

The harness expectation is also updated to reflect that the copy action is supporting evidence rather than an absolute requirement.

## Verification

- `bun test tests/browser-worker-contract.test.ts tests/chatgpt-web-harness.test.ts` — 121 passed, 0 failed
- `git diff --check`

### DIFF
--- a/src/adapters/chatgpt-web/browser-worker.ts
+++ b/src/adapters/chatgpt-web/browser-worker.ts
@@ -102,6 +102,7 @@ export const CHATGPT_RESPONSE_DOM_GRACE_MS = 60_000;
 export const CHATGPT_EMPTY_RESPONSE_GRACE_MS = 10_000;
 export const CHATGPT_COMPLETION_ACTION_GRACE_MS = 60_000;
 export const CHATGPT_COMPLETION_SETTLE_MS = 2_000;
+export const CHATGPT_MARKERLESS_COMPLETION_SETTLE_MS = 5_000;
 export const CHATGPT_TOOL_CONFIRMATION_TIMEOUT_MS = 60_000;
 export const MAX_CHATGPT_CONNECTOR_TRIGGER_ATTEMPTS = 3;
 const CHATGPT_SMOKE_TEXT = "Reply with exactly: CODEX WEB GPT READY";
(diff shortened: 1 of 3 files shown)


miuuyy:
Thanks, but this removes a deliberate completion invariant. Five seconds of stable DOM with `running=false` is not ChatGPT-owned proof of completion and can truncate a slow or temporarily paused stream.

I’m closing this approach. A replacement would need deterministic finalization evidence rather than a stability timer.

PR #175, opened 2026-08-26
fix: preserve turn provenance after compaction

naakto14:
## Problem

Native Codex compaction can start a fresh turn while retaining the human instruction that was summarized from the previous turn.

In that wire shape:

- canonical turn metadata and the freshly injected environment envelope use the new `turn_id`;
- the retained historical human instruction still carries its original `turn_id`;
- a native compaction item marks the boundary between the retained history and the fresh continuation.

v3.0.3 treats the retained user `turn_id` mismatch as untrusted and can therefore fail to recover the fresh Codex environment after compaction. That breaks continuation of the same task even though Codex supplied a new, authoritative environment envelope for the post-compaction turn.

## What this changes

Adds a narrow compaction-resume recovery path that accepts the historical user turn provenance only when a native compaction boundary proves this exact transition.

The recovery path requires all of the following:

- current canonical metadata contains a non-empty `thread_id` and fresh `turn_id`;
- canonical sandbox metadata is present and recognized;
- the retained real user message has a historical `turn_id` different from the fresh turn;
- a native compaction boundary follows that retained user message;
- the adjacent environment message is explicitly owned by the fresh current turn;
- any intervening developer messages satisfy the existing server-owned/current-turn provenance checks;
- the recovered environment still matches canonical workspace and sandbox metadata.

Only after those checks does the adapter recover the fresh post-compaction environment.

## Security / authority boundary

This exception is intentionally about provenance continuity, not filesystem authority.

The recovered environment is still passed through the strict metadata-bound root validation. A retained user message cannot inject an extra workspace root through compaction, and the sandbox policy must still agree with canonical metadata.

A regression test explicitly adds an out-of-scope root to the fresh environment and verifies that recovery fails closed rather than widening authority.

## Turn identity behavior

The fresh Codex `turn_id` remains authoritative for the resumed execution. The harness verifies that the post-compaction execution key differs from the pre-compaction one while the task's trusted cwd is still recovered successfully.

## Tests

Coverage includes:

- successful recovery when compaction retains the older user `turn_id`;
- rejection of a compaction resume that attempts to widen canonical workspace authority;
- harness coverage showing that the fresh turn identity is accepted and produces a new execution key.

## Verification

- `bun test tests/environment.test.ts tests/chatgpt-web-harness.test.ts` — 77 passed, 0 failed
- `git diff --check`

### DIFF
--- a/src/adapters/chatgpt-web/environment.ts
+++ b/src/adapters/chatgpt-web/environment.ts
@@ -93,6 +93,37 @@ function contextualUserMessage(value: Record<string, unknown>): boolean {
     || text === OPAQUE_COMPACTION_NOTE;
 }
 
+function compactionBoundaryItem(value: Record<string, unknown>): boolean {
+  return value.type === "compaction"
+    || value.type === "compaction_summary"
+    || value.type === "context_compaction";
+}
+
+function latestRealUserIndex(input: unknown[]): number {
+  for (let index = input.length - 1; index >= 0; index -= 1) {
+    const item = record(input[index]);
+    if (item?.type !== "message" || item.role !== "user" || contextualUserMessage(item)) continue;
+    const messageTurnId = itemTurnId(item);
+    const serverOwnedId = typeof item.id === "string" && item.id.length > 0;
+    if (messageTurnId === undefined && !serverOwnedId) continue;
+    return index;
+  }
+  return -1;
+}
+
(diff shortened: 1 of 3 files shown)


miuuyy (changes requested):
Thanks for the compaction work. I’m declining this PR because it relaxes `turn_id` provenance from a compaction boundary before the fresh current-turn environment is authenticated, leaving the trust proof incomplete.

PR #177, opened 2026-08-26
Harden ChatGPT web trusted cwd recovery and turn replay

naakto14:
## Summary

This PR hardens the ChatGPT Web adapter around native Codex environment recovery, compaction/resume, browser submission ambiguity, localized rate limits, and release packaging.

The immediate bug that prompted the latest fixes was a ChatGPT Web turn failing with:

> `stream disconnected before completion: ChatGPT web turn is missing cwd in trusted Codex environment context`

That could occur when Codex supplied an authenticated environment envelope whose `cwd` was omitted or reshaped during a trusted native/compaction flow. The adapter previously treated that as an unrecoverable missing `cwd`, even when canonical Codex metadata was sufficient to authenticate exactly one workspace root.

### Trusted environment and `cwd` recovery

- Distinguishes an actually missing `cwd` from an invalid/conflicting `cwd` declaration.
- Recovers `cwd` only from metadata-authenticated workspace roots when a trusted environment envelope omits the explicit `<cwd>` field.
- Keeps the trust boundary narrow: arbitrary user-authored environment XML cannot widen filesystem authority.
- Supports canonical multi-environment payloads while rejecting malformed, duplicate, ambiguous, or conflicting `cwd` declarations.
- Authenticates the task-scoped Codex visualization root using the current thread id and Codex home layout instead of trusting arbitrary extra workspace roots.
- Preserves the resolved trusted `cwd` across the ChatGPT Web turn lifecycle so the value is not lost after initial extraction.

### Native compaction / resume handling

- Accepts the native Codex case where a resumed turn receives a fresh `turn_id` while the retained pre-compaction user message still carries the historical turn id.
- Requires a native compaction boundary before relaxing that turn-id equality check.
- Recovers the fresh environment envelope only when it is owned by the new turn and still matches canonical sandbox/workspace metadata.
- Continues to reject unrelated cross-turn user messages and spoofed environment envelopes.

### Browser submission replay safety

- Tracks submission state as `prepared`, `send_activated`, and `accepted`.
- Introduces an acknowledgement handshake between the browser helper and the turn owner at the irreversible Send boundary.
- If a 502/transport failure happens after Send activation or after ChatGPT has accepted the prompt, the adapter records a terminal non-retryable submitted-turn error instead of resending the same prompt on a fresh browser surface.
- Replays the already-recorded terminal result for duplicate ownership/retry attempts, preventing duplicate ChatGPT submissions.
[...]

PR #178, opened 2026-08-26
fix(chatgpt-web): recover trusted cwd from canonical metadata

naakto14:
## Summary

Fixes ChatGPT web turns failing with `ChatGPT web turn is missing cwd in trusted Codex environment context` when the native Codex environment envelope omits an explicit `<cwd>` but still provides canonical workspace metadata for the active turn.

This PR is intentionally scoped to the trusted cwd/environment handling only. It is based directly on upstream `main` and changes only:

- `src/adapters/chatgpt-web/environment.ts`
- `tests/environment.test.ts`

## Problem

Recent Codex request shapes can contain a trusted native `<environment_context>` with workspace roots and sandbox information while omitting an explicit `<cwd>`. The ChatGPT web adapter previously required exactly one explicit cwd declaration, so otherwise-valid turns failed before execution.

There is also a replay/compaction edge case where contextual user-envelope messages may appear after the actual human prompt. Treating the last `role=user` item as the active user message can make trusted-environment lookup inspect the wrong boundary and report a missing cwd.

## Fix

- Recover cwd only from canonical turn metadata when the native environment envelope has no explicit cwd.
- Require the metadata workspace to be absolute, uniquely identifiable, consistent with the declared workspace roots, and bound to the same sandbox metadata.
- Keep malformed, ambiguous, conflicting, or unprovenanced cwd declarations fail-closed instead of silently falling back.
- Distinguish an absent cwd from an invalid cwd declaration so metadata recovery is only used for the former.
- Ignore contextual/system-generated user-envelope messages when locating the active human user message for trusted environment extraction.

## Security / trust properties

The fallback does not trust arbitrary prompt or developer text. It only accepts cwd recovered from canonical `x-codex-turn-metadata` when that metadata authenticates against the native environment envelope. Ambiguous roots, mismatched sandbox metadata, malformed `<cwd>` markup, duplicate primary environments, and conflicting candidates continue to fail closed.

## Tests

Added regression coverage for:

- cwd recovery from canonical workspace metadata
- rejection of unprovenanced developer fallback text
- workspace/root mismatch
- multiple matching canonical workspaces
- malformed or ambiguous explicit cwd declarations
- multiple primary environments and multiple legacy candidates
- invalid native envelopes remaining fail-closed

Validation performed:

- `bun test tests/environment.test.ts` — 34 passed, 0 failed
- `bun run typecheck` — passed
- `git diff --check` — passed

This supersedes the previously closed #177, whose head was the fork `main` branch and therefore unintentionally included unrelated fork commits.

### DIFF
--- a/src/adapters/chatgpt-web/environment.ts
+++ b/src/adapters/chatgpt-web/environment.ts
@@ -434,7 +434,17 @@ function rawEnvironmentText(parsed: CodexParsedRequest): string | undefined {
 function hasRawEnvironmentEnvelope(parsed: CodexParsedRequest): boolean {
   const body = record(parsed._rawBody);
   const input = Array.isArray(body?.input) ? body.input : [];
-  for (const value of input) {
+  let activeUserIndex = -1;
+  for (let index = input.length - 1; index >= 0; index -= 1) {
+    const item = record(input[index]);
+    if (item?.role === "user" && !contextualUserMessage(item)) {
+      activeUserIndex = index;
+      break;
+    }
+  }
+  if (activeUserIndex < 0) return false;
+
+  for (const value of input.slice(0, activeUserIndex)) {
     const item = record(value);
     if (item?.type !== "message" || item.role !== "user") continue;
     const content = Array.isArray(item.content) ? item.content : [];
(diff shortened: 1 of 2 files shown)


miuuyy (changes requested):
Thanks for narrowing the earlier branch. I’m declining this PR because it infers `cwd` without a demonstrated real request fixture, while the raw-envelope scan can treat historical context as a current failure and no longer composes cleanly with the environment validation on `main`.

Issue #194, opened 2026-08-29
Preserve one native Codex turn across ChatGPT Web long-run termination


David-T-Campos:
## Problem

The bridge intentionally has no absolute browser-turn deadline, but ChatGPT Web can still terminate a long reasoning run independently of the bridge. One visible terminal state is `Stopped thinking`; current `browser-worker.ts` treats that state as a non-retryable `client_cancelled` error after 5 seconds. A Web-side long-run termination can therefore end an otherwise healthy native Codex task even though the outer Codex harness, tool state, broker capability, and browser conversation are still available.

There is a second ambiguity: a Web-side execution ceiling can sometimes look like an ordinary completed assistant turn. In that case the bridge has completion DOM evidence, but it cannot know whether the original Codex task was actually finished or the Web execution simply returned the best partial result it had at the ceiling.

## Desired invariant

Keep one native Codex turn alive across a Web-side execution boundary without resending the original task and without fabricating completion.

A safe implementation should:

- never replay the original accepted Codex prompt;
- continue only inside the same owned Temporary Chat / turn-bound MCP capability;
- preserve already streamed Markdown and tool results exactly;
- on a durable `Stopped thinking` state, ask the retained Web conversation to continue the same Codex task rather than converting it into `client_cancelled`;
- for unusually long turns that otherwise look complete, run a private completion audit in the same conversation and suppress an exact completion sentinel from the Codex output;
- stream actual continuation content normally if the audit finds unfinished work;
- stay bounded and fail closed if the continuation protocol cannot prove a sane result;
- exclude compaction and Luna checkpoint flows unless their separate invariants are handled explicitly.

I am preparing this as a focused change with pure protocol tests plus the existing browser contract suite, rather than changing retry behavior. Making `Stopped thinking` retryable at the HTTP layer would be unsafe because the original prompt has already been submitted and must not be resent.

miuuyy:
Before treating this as a general Web execution boundary, could you clarify whether `Stopped thinking` was observed on a real account or is currently a protocol hypothesis?

If it was observed, please provide the account tier, approximate recent usage/number of long turns or retries, whether any cooldown or rate-limit UI appeared, and how long the turn ran before stopping. It would also help to know whether the same task reproduces on a fresh low-usage session.

David-T-Campos:
[...]

PR #195, opened 2026-08-29
fix: recover one durable stopped-thinking boundary

David-T-Campos:
Addresses #194.

This PR is now deliberately narrow. It handles only the durable ChatGPT Web `Stopped thinking` UI state that the bridge already recognizes as terminal; it no longer infers an execution boundary from elapsed time.

What changes:

- On a normal non-Luna, non-compaction turn, the first durable `Stopped thinking` observation may trigger exactly one private recovery message in the same owned Temporary Chat.
- The already accepted Codex request is not replayed. The recovery path does not call `turn.prepare` or `turn.onSendActivated`; it sends only the private recovery instruction into the existing conversation.
- Existing work, tool effects, decisions, and streamed text remain authoritative. The recovery instruction tells the model to keep using the same turn-bound Codex Native capability / `turn_token` if more local work is actually required.
- If the prior work was already complete, a private completion sentinel is suppressed from Codex output. Otherwise only the missing continuation is returned.
- Recovery is event-driven and single-use. A second durable `Stopped thinking` observation fails closed through the existing terminal error path.
- Compaction and Luna rolling-checkpoint flows remain excluded.
- The previous 20-minute completion audit, generic long-turn continuation behavior, and eight-round continuation budget have been removed. This adds no absolute browser-turn deadline; the existing 30-second stalled-response capture remains diagnostic only.

Validation on the exact proposed tree (`eb0988ea8f992e4648c0638b530df38de3cbe8c6`):

- focused continuation + browser-worker contract tests: passed
- `bunx tsc --noEmit`: passed
- one clean commit on current upstream `main` (`34336fb2deb5d17bcf12be74c6573c2fb26ee0ed`)
- changed files are only `browser-worker.ts`, `continuation.ts`, and the focused regression test; no temporary Gauntlet workflow is in the proposed tree

Evidence boundary: I do **not** have a real-account reproduction or redacted browser trace for a spontaneous `Stopped thinking` incident. The case was a protocol hypothesis based on the bridge's existing durable terminal-state handling. This PR therefore should be read as narrowly tested recovery hardening for that observed DOM state, not proof that current ChatGPT accounts hit such a boundary under a particular tier, usage level, or duration.

### DIFF
--- a/src/adapters/chatgpt-web/browser-worker.ts
+++ b/src/adapters/chatgpt-web/browser-worker.ts
@@ -87,6 +87,10 @@ import {
   ChatGptExternalTurnProgress,
   chatGptExternalProgressIsLive,
 } from "./turn-progress";
+import {
+  ChatGptCompletionAuditGate,
+  chatGptStoppedThinkingRecoveryPrompt,
+} from "./continuation";
 
 export { MAX_CHATGPT_BROWSER_TABS } from "./concurrency";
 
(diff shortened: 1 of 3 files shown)


miuuyy:
Thanks for keeping the original prompt single-submit and the continuation in the owned conversation. I am not going to merge this as-is yet.

Important clarification: the current bridge has no absolute browser-turn deadline. This PR does **not** cut off a Pro generation at 20 or 25 minutes; its 20-minute threshold runs only after ChatGPT already exposes a completed turn. However, the change combines two different behaviors:

1. recovery from a durable `Stopped thinking` terminal state;
2. a speculative completion audit after any apparently completed response lasting at least 20 minutes.

The first may be a real recovery path. The second adds an extra model turn after a potentially correct long answer and can trigger more work or tool use based only on elapsed time. The eight-round continuation budget also needs real browser evidence, not only pure protocol tests.

Please split the durable `Stopped thinking` recovery from the elapsed-time completion audit and attach a redacted real trace showing the Web-side termination this is intended to recover. Keep compaction and Luna excluded. I am marking this **P2** while the runtime evidence and boundary are unresolved.

David-T-Campos:
Good catch on the boundary. I narrowed the branch rather than trying to defend the elapsed-time heuristic: the 20-minute completion audit, generic long-turn continuation, and 8-round loop are gone. Current head `eb0988e` performs at most one same-conversation recovery only after the durable `Stopped thinking` DOM state is actually observed; a second occurrence fails closed, and Luna/compaction stay excluded. It also remains single-submit for the original Codex request.

On the evidence request: I cannot substantiate a real-account spontaneous `Stopped thinking` reproduction. This was a protocol hypothesis based on the bridge's existing durable terminal-state handling, so I do not have honest tier/usage/cooldown/duration data or a redacted live trace to attach. I have made that limitation explicit in the PR instead of treating it as runtime-verified. Focused tests, typecheck, and full verify are green on the narrowed one-commit branch.

miuuyy:
Thank you for narrowing the change and for being explicit about the lack of a real reproduction. I am closing this PR because it would still add another model turn for a protocol hypothesis rather than a demonstrated failure.

I do not want to ask for further speculative revisions. A future, independently reproducible `Stopped thinking` trace can justify reopening the design discussion.

PR #197, opened 2026-08-29
fix: preserve native subagent delegation for Pro

### LINKED ISSUE #196: Remove Pro-only subagent suppression from the native Codex harness contract
## Native-parity mismatch

Full mode advertises the same turn-bound Codex Native harness to every Web effort, including Pro, and the project invariant says not to introduce effort-specific MCP exclusions. The model catalog was also explicitly changed to keep Pro inside Codex's native code-mode/subagent registry (`d3049cb`, "Enable Pro subagent spawning").

However, the current prompt compiler adds a Pro-only instruction:

> Complete this task directly in the current parent response. Do not create, spawn, delegate to, or wait on sub-agents...

That means Pro can receive the same collaboration tools but is told not to use them, producing behavior that intentionally differs from the same native Codex task running on another effort.

## Proposed narrow fix

Remove only the Pro-specific anti-delegation transport instruction. Do not add new agent logic to the bridge. Let the outer Codex tool registry, permissions, subagent protocol selection, and declared collaboration schemas determine whether delegation is available, just as they already do for other Web efforts.

Regression coverage should prove:

- Pro and Extra High receive the same generic transport contract with no effort-specific anti-agent instruction;
- the stable turn token and existing local-tool contract remain unchanged;
- compaction remains tool-isolated;
- existing subagent/MCP protocol tests remain green.

This is deliberately smaller than changing MultiAgent V2 routing or encrypted provider-private payload handling, which should remain a separate parity problem.

David-T-Campos:
Closes #196.

Full mode already gives every Web effort the same turn-bound Codex Native capability, and the outer Codex harness remains authoritative for collaboration tools and subagent protocol selection. The prompt compiler nevertheless added a Pro-only instruction telling the model not to spawn, delegate to, or wait on subagents.

This removes only that effort-specific behavioral suppression. It does not add a second agent runtime and it does not change the MCP broker, permissions, sandboxing, or MultiAgent protocol handling. Pro now receives the same generic native Codex tool contract as Extra High and can use collaboration tools when the outer Codex turn exposes them.

Validation on the proposed source change:

- targeted prompt/subagent/harness tests: passed
- `bunx tsc --noEmit`: passed
- no temporary CI workflow files are included in this clean branch

The branch is one commit and two files on top of upstream `97fe185f19bd10358d532f5c4966617d9cd48388`.

### DIFF
--- a/src/adapters/chatgpt-web/prompt.ts
+++ b/src/adapters/chatgpt-web/prompt.ts
@@ -413,11 +413,6 @@ export function compileChatGptWebPrompt(
       "Do not claim a new local inspection, command, edit, or verification unless it actually appears in the task history. If the latest request requires fresh local-computer access or a local mutation, state only that exact limitation instead of inventing success.",
       "Otherwise perform the full requested research, analysis, or synthesis with every capability actually available to you; do not stop at a plan or progress report.",
     ];
-  const proDelegationContract = !parsed._compactionRequest && mode.effort === "max"
-    ? [
-      "Complete this task directly in the current parent response. Do not create, spawn, delegate to, or wait on sub-agents, parallel agents, background agents, or delegated workers, even if such tools are available. Use non-agent tools directly instead.",
-    ]
-    : [];
   const checkpointContract = captureLunaCheckpoint
     ? [
       "After the complete user-facing answer, append one private rolling task checkpoint for the next Luna turn.",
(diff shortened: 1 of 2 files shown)


miuuyy:
This is a valid native-parity question, but the Pro-only suppression was added intentionally when Pro received the Full MCP loop. Removing seven lines is mechanically safe; proving the behavioral boundary is the actual requirement.

Before accepting this, please provide one live Pro trace showing that `spawn_agent` / `wait_agent` is executed through the outer Codex collaboration registry, that no ChatGPT-internal delegated worker is started, and that the parent/child result returns through the same turn-bound MCP lifecycle. Also cover cancellation and compaction while a child is active.

I am keeping this open as **P2** rather than treating the current difference as an accidental bug.

PR #200, opened 2026-08-29
fix: preserve native history metadata in Web prompts

### LINKED ISSUE #199: Preserve native assistant phase and MCP namespace metadata in Web context replay
## Native-parity gap

The Responses parser already preserves native Codex message metadata that affects conversation semantics:

- assistant message `phase` (`commentary` vs `final_answer`);
- tool-call `namespace` for namespaced/MCP tools;
- tool-result `toolNamespace`.

But `src/adapters/chatgpt-web/prompt.ts` drops all three when compiling accumulated Codex history for ChatGPT Web. `messageEnvelope()` emits an assistant only as `{role, content}`, `assistantContent()` emits a tool call without its namespace, and tool results omit their namespace.

That makes the Web model see a lossy history compared with the native Responses conversation. A previous commentary message can become indistinguishable from a final answer, and two tools with the same short name in different MCP namespaces can lose the identity Codex originally recorded.

## Proposed focused fix

Preserve this metadata verbatim in the JSON transport envelope:

- assistant envelope: include `phase` when present;
- prior tool-call item: include `namespace` when present;
- tool-result envelope: include `tool_namespace` when present.

Do not expose provider-private signatures or encrypted reasoning; this issue is only about metadata already modeled as normal Codex conversation semantics.

Add regression coverage for inline and multipart prompt compilation and prove no fields are invented when absent.

David-T-Campos:
Closes #199.

The Responses parser already preserves native Codex assistant phase and MCP namespace metadata, but the ChatGPT Web prompt compiler dropped it when replaying accumulated history. This makes routed Web history semantically lossier than the native Responses conversation.

This preserves, without inventing new values:

- assistant `phase` (`commentary` / `final_answer`);
- prior tool-call `namespace`;
- tool-result `tool_namespace`.

Both inline and experimental multipart prompt paths use the same envelope builder, so both now retain the metadata.

Validation on the exact proposed tree:

- targeted history + prompt + harness tests: passed
- `bunx tsc --noEmit`: passed
- no temporary CI workflow files are included in the clean branch

The branch is one commit and two files on top of upstream `97fe185f19bd10358d532f5c4966617d9cd48388`.

### DIFF
--- a/src/adapters/chatgpt-web/prompt.ts
+++ b/src/adapters/chatgpt-web/prompt.ts
@@ -210,7 +210,13 @@ function assistantContent(content: CodexAssistantContentPart[]): unknown[] {
   return content.map(part => {
     if (part.type === "text") return { type: "text", text: part.text };
     if (part.type === "thinking") return { type: "thinking_summary", text: part.thinking };
-    return { type: "tool_call", id: part.id, name: part.name, arguments: part.arguments };
+    return {
+      type: "tool_call",
+      id: part.id,
+      name: part.name,
+      ...(part.namespace ? { namespace: part.namespace } : {}),
+      arguments: part.arguments,
+    };
   });
 }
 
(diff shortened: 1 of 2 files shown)


miuuyy:
The underlying parity gap is valid and the production diff is appropriately narrow: existing `phase`, tool `namespace`, and result `toolNamespace` values should not be discarded when history is compiled for Web.

Before merge, please add one parser-to-prompt regression fixture using an actual Responses request. The current tests construct `CodexParsedRequest` directly, so they prove serialization but not the complete native request path that is claimed here. Please also rebase on current `main` before the next review.

Keeping this as **P2**.

PR #201, opened 2026-08-29
feat: preserve native Responses text controls

### LINKED ISSUE #198: Preserve native Codex Responses text controls in the Web transport
## Native-parity gap

Current Codex sends Responses `text` controls when the selected native model supports them. In current upstream Codex, `TextControls` carries response `verbosity` plus an optional strict JSON-schema `format`, and `build_responses_request` sends that field with the model request.

`codex-chatgpt-web` currently accepts `text` as `unknown` in the request schema but `parseRequest()` never consumes it, so routed Web models silently lose both controls. That changes observable behavior for configured verbosity and breaks the model-side contract for `codex exec --output-schema` / SDK structured-output flows.

## Proposed focused fix

- Parse native `text.verbosity` (`low|medium|high`) and `text.format` when it is a Responses `json_schema` format.
- Preserve those controls in `CodexRequestOptions`.
- Add a transport-only output contract to the ChatGPT Web prompt:
  - match the requested native verbosity;
  - when a JSON schema is present, emit the final user-facing answer as one JSON value matching that exact schema, with no Markdown fence or extra prose.
- Preserve the schema as data; do not reinterpret task priority or replace Codex's own prompt assembly.
- Do not claim provider-enforced schema validation: consumer ChatGPT Web has no Responses `text.format` protocol field. A separate follow-up can add local validation/repair if needed.

Regression tests should cover parsing, multipart and inline prompt paths, no-op behavior when `text` is absent, and strict schema transport.

David-T-Campos:
Closes #198.

Current Codex sends Responses `text` controls for native models: response verbosity when supported, plus a JSON-schema `text.format` for structured-output flows. The bridge schema accepted `text`, but the routed Web parser discarded it.

This preserves those controls through the Web transport and enforces the strict-output boundary locally:

- parses native `text.verbosity` (`low`, `medium`, `high`) and preserves it as a best-effort Web instruction;
- parses Responses `json_schema` output formats without rewriting the schema;
- carries both controls in `CodexRequestOptions` and the inline/multipart Web transport contracts;
- treats the schema as output-format data rather than higher-priority instructions;
- excludes final-answer output controls from compaction turns;
- for `strict: true`, compiles the supplied schema locally with Ajv before browser execution;
- buffers final-answer text for strict structured-output turns until the complete browser answer is available;
- parses the complete answer with `JSON.parse` exactly as returned and validates it against the requested schema;
- fails closed on malformed JSON or schema-invalid JSON and never emits that invalid final answer as apparent success;
- does not invoke a repair model and does not strip Markdown fences, prefixes, suffixes, or surrounding prose.

`strict: false` remains best-effort.

Validation on the exact proposed tree (`eefa4a434fc38e3d0ba6491e2942ec4fb5df3e78`):

- adapter-level strict-output regression verifies invalid browser JSON produces no final-answer delta and ends with `structured_output_validation_failed`;
- adapter-level success regression verifies valid strict JSON is emitted only after complete-answer validation;
- targeted Responses text-control + prompt + harness tests: passed;
- `bunx tsc --noEmit`: passed;
- full `bun run verify`: passed;
- no temporary Gauntlet workflow files are included.

The branch is one clean commit on current upstream `main` (`34336fb2deb5d17bcf12be74c6573c2fb26ee0ed`).

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -6,6 +6,8 @@
       "name": "codex-chatgpt-web",
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.30.0",
+        "ajv": "8.20.0",
+        "ajv-formats": "3.0.1",
         "chromium-bidi": "12.1.0",
         "fflate": "^0.8.2",
         "playwright-core": "^1.62.0",
--- a/package.json
+++ b/package.json
@@ -53,6 +53,8 @@
   },
   "dependencies": {
     "@modelcontextprotocol/sdk": "^1.30.0",
+    "ajv": "8.20.0",
+    "ajv-formats": "3.0.1",
     "chromium-bidi": "12.1.0",
     "fflate": "^0.8.2",
     "playwright-core": "^1.62.0",
--- a/src/adapters/chatgpt-web/index.ts
+++ b/src/adapters/chatgpt-web/index.ts
@@ -10,6 +10,7 @@ import { ChatGptBrowserWorker } from "./browser-worker";
 import { extractChatGptTurnEnvironment, extractChatGptTurnIdentity } from "./environment";
 import { CHATGPT_WEB_LUNA_MODEL_ID, resolveChatGptWebModelMode, type ChatGptWebCapabilities } from "./model";
 import { chatGptReadOnlyContextWarning, compileChatGptWebPrompt } from "./prompt";
+import { createChatGptStructuredOutputValidator } from "./output-validation";
 import { chatGptWebTurnRetryPolicy } from "./retry-policy";
 import { TurnBroker, type BrokerToolRequest, type BrokerToolResult, type TurnBrokerOwner } from "./turn-broker";
 import { ChatGptTextFeed, ChatGptTraceFeed, chatGptCompactionSourceExecutionKey, chatGptThreadOwnershipKey, chatGptTurnExecutionKey, chatGptTurnRetryKey, chatGptTurnRoundKey, chatGptTurnSessions, type ChatGptBrowserOutcome, type ChatGptTraceEvent, type ChatGptTurnRuntime, type ChatGptTurnSession } from "./turn-execution";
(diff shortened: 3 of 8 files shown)


miuuyy:
Preserving `text.verbosity` as a best-effort Web instruction is reasonable. The strict JSON-schema portion is not complete enough to merge in its current form.

A prompt instruction does not preserve native `strict: true` semantics: ChatGPT Web can still return malformed JSON or a schema-invalid value, while Codex/SDK callers may reasonably treat this route as structured output. That would turn a strict native contract into unverified apparent success.

Please either split this PR so it handles verbosity only, or add deterministic local JSON parsing plus schema validation and fail closed when the final answer does not satisfy the requested schema. Do not add a repair model or silently strip surrounding prose. Marked **P2** pending that boundary.

David-T-Campos:
Addressed the P2 boundary on the current head (`6a7aae5`). `strict: true` is no longer prompt-only: the bridge now buffers the complete Web final answer, runs exact `JSON.parse` plus local Ajv schema validation, and only emits the final-answer delta after validation succeeds. Malformed JSON, surrounding prose/fences, or schema-invalid JSON fail closed; there is no repair model and no stripping/cleanup. Invalid schemas also fail before browser execution. Targeted tests, `tsc --noEmit`, and full `bun run verify` are green on the one-commit branch rebased to current `main`.

David-T-Campos:
I tightened the review boundary one more step on head `eefa4a4`: there are now adapter-level regressions through `createChatGptWebAdapter(...).runTurn`, not just validator unit tests. An invalid strict browser answer is streamed into the adapter in chunks and is verified to produce **zero** `final_answer` deltas before ending with `structured_output_validation_failed`; a valid strict answer is verified to emit exactly one complete final-answer delta only after validation, then `done`. Targeted tests, typecheck, and full `bun run verify` all pass. The branch is back to one clean commit on current `main`, with all temporary Gauntlet workflows removed.

PR #214, opened 2026-08-30
fix(composer): tolerate strict length-preserving prompt mutations by …

buitheanh874:
…Lexical


miuuyy:
Thanks for the contribution. I am closing this because the change weakens a core prompt-integrity boundary: equal character length is not proof that ChatGPT preserved the submitted prompt. There is also no isolated real-world reproduction or diagnostic evidence establishing that this broader acceptance is safe.

PR #215, opened 2026-08-30
Fix headless/Xvfb viewport collapse and stale turn-tab acquisition

jaivial:
## Summary
Running the launcher on a headless Linux server (Xvfb, systemd) surfaced three failure modes that all end in `ChatGPT browser surface did not expose an operational viewport` or `multipart_stage_2_send` timeouts. This PR fixes the underlying viewport lifecycle:

- **Default surface bounds `1x1` → `1280x800` + floor in `setBounds`** — until the renderer sends its first measured bounds, the embedded ChatGPT surface sits at 1x1 and the operational-viewport probe can never pass.
- **Home surface stays composited with a real viewport** — `setVisible(false)` makes Electron collapse a WebContentsView renderer to `0x0` even with valid bounds; hidden-turn parking now keeps views alive.
- **On-screen parking for hidden turns** — offscreen `x = contentWidth + 1` parking still collapses to `0x0` under Xvfb; parked views now keep a real (`>= 800x600`) viewport at `0,0`.
- **Turn-tab reaper** — finished tabs retained their surface ids and dead `0x0` viewports, so later acquisitions could bind to a dead surface (`multipart_stage_2_send` after several turns). New turns now reap finished tabs first.
- **Stall budget 300s → 900s** — long reasoning + large tool writes are legitimate upstream silence; the old budget cut healthy turns mid-stream.

## Testing
- Headless Xvfb `:99` + systemd, single account, `browser-only` and `full` modes.
- `codex exec` turns repeatedly before (viewport timeout loops, `multipart_stage_2_send` after ~3 turns) and after (stable, multi-turn, large docker-build tool outputs).
- Interactive launcher sign-in, smoke test, and model install unaffected on a desktop session.

No API or protocol changes; Electron view lifecycle only.

### DIFF
--- a/launcher/electron/browser-host.cjs
+++ b/launcher/electron/browser-host.cjs
@@ -243,7 +243,7 @@ class BrowserHost {
     this.turnLeaseSweep = setInterval(() => this.reapExpiredTurnTabs(), TURN_HEARTBEAT_SWEEP_MS);
     this.turnLeaseSweep.unref?.();
     this.boundsReady = false;
-    this.bounds = { x: 0, y: 0, width: 1, height: 1 };
+    this.bounds = { x: 0, y: 0, width: 1280, height: 800 };
     this.state = {
       status: "idle",
       message: "No active task",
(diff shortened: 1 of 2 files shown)


miuuyy:
Thanks for documenting the Xvfb environment and the failures you observed. I am closing this PR because it changes core launcher behavior globally for a narrow headless scenario. The proposed viewport, visibility, retained-tab, and timeout changes alter established desktop and conversation-lifecycle invariants, so this cannot be accepted as a general fix on the current evidence.

Headless server support needs an agreed product boundary before more implementation is submitted. Please do not assume that another revision will be merged: repeated iterations on an unapproved direction increase review cost and do not by themselves make the contribution more valuable to the project.

PR #217, opened 2026-08-30
Headless CLI mode: CLI-only setup, headless login, systemd user services

jaivial:
## Summary
Makes the full lifecycle drivable from a terminal on a displayless Linux server — no launcher, no Electron, no persistent Xvfb. Turns already supported `browserHost: "managed-chrome"` + `headless: !headed`; this closes the remaining gaps: interactive-only login, macOS-only service management, and zero headless guidance in `doctor`.

## Login (src/browser-login.ts, src/cli.ts)
- `login --login-headless auto|force|off` (default `auto`): Chromium headless attempt → bot-challenge detection → automatic fallback to ephemeral `Xvfb :N` + headful Chrome (auto-cleanup, free-display scan).
- `login --storage-state-file PATH`: import a Playwright `storageState` JSON, verify against the live ChatGPT surface, skip interactive login.
- Composes with the existing headed Chrome flow (`off` preserves today's behavior verbatim).

## Services (src/service.ts, src/tunnel-service.ts)
- Linux: systemd **user** units for daemon + tunnel client — `ExecStart` from the durable runtime command, `Restart=always`, `CODEX_CHATGPT_WEB_HEADLESS=1`, `enable --now`, best-effort `loginctl enable-linger` with a warning when it fails.
- Space-containing `ExecStart` args are quoted; existing args pass through untouched.
- `restart`/`stop` keep the atomic drain/idle contract; macOS launchd path is platform-guarded and unchanged.

## Setup + doctor
- `setup --browser-host managed-chrome --headless --login-headless auto` → config persists `managed-chrome` + `headed: false`; no launcher-browser-host requirement.
- `doctor` (Linux): xvfb availability, linger state, browser-host/headless mode; service messages de-MacOS'd.

## Tests
- New `tests/service-unit.test.ts`: unit rendering incl. HEADLESS env, Restart, WantedBy, and space-quoted ExecStart args.
- Full suite: 427 tests pass; `tsc --noEmit` clean.
- Verified on a headless Xvfb-less Linux server path: user systemd units match the hand-rolled units this setup was validated against in production (Restart/linger/headless env), and the managed-chrome headless turn path exercises the same adapter flags already shipped for `headed: false`.

## Out of scope
- Auto Xvfb fallback for turns (login only) — turns run pure headless; guard can reuse the login helper if ChatGPT ever blocks headless turn browsers.
- Windows headless.

### DIFF
--- a/src/browser-login.ts
+++ b/src/browser-login.ts
@@ -1,7 +1,8 @@
 import { spawn } from "node:child_process";
 import { existsSync, mkdirSync, readFileSync, rmSync } from "node:fs";
 import { dirname, join } from "node:path";
-import { chromium, type BrowserContextOptions } from "playwright-core";
+import { chromium, type BrowserContextOptions, type Page } from "playwright-core";
+import { runCommand } from "./process";
 import type { AppConfig } from "./config";
 import { atomicWriteFile } from "./config";
 import {
@@ -47,10 +48,11 @@ function writeVerificationMarker(
 async function inspectStoredState(
   config: AppConfig,
   storageState: NonNullable<BrowserContextOptions["storageState"]>,
+  headless = false,
 ): Promise<ChatGptWebAccountCapabilities & { url: string }> {
   const verifierBrowser = await chromium.launch({
     executablePath: config.chromeExecutablePath,
-    headless: false,
+    headless,
     ignoreDefaultArgs: ["--password-store=basic", "--use-mock-keychain"],
     args: ["--no-first-run", "--no-default-browser-check"],
   });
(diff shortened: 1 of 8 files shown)


miuuyy:
Thank you for the substantial work here. This is a technically strong and thoughtful contribution, and closing it is not a judgment about the contributor's intent or the implementation effort.

I have decided not to take the repository in this direction. Combining launcher-free operation, headless login, importable browser session state, persistent systemd services with automatic restart and linger, and a near one-command VPS workflow materially increases the risk of this project being used for unattended or abusive account automation. Some lower-level headless capabilities already exist, but productizing the entire lifecycle as a first-class server mode changes the project's risk profile significantly.

I am choosing the repository's long-term legitimacy, stability, and defensible intended use over maximizing unattended deployment capability. For that reason, this PR is being closed as a product-boundary decision rather than as a request for another implementation revision. The technical work is appreciated, but this feature direction will not be accepted.

PR #219, opened 2026-08-30
feat(context): hybrid split for prompts exceeding 15,000 characters

buitheanh874:


miuuyy:
Thanks for the idea, but I can’t accept this implementation.

The new split function is never connected to the turn path, so no prompt is actually split and no overflow buffer reaches `attachFiles`. At the same time, the PR changes the existing image-upload selector and alert contract without tests, which adds regression risk to a working path while leaving the proposed feature inactive.

The current Bigger Context transport already handles large structured prompts without turning the tail into an unverified generic attachment. I’m closing this PR rather than carrying dead feature surface into the browser worker.

PR #222, opened 2026-08-30
fix: carry MCP turn progress into the launcher browser helper process

1saifj:
Fixes the class of failures in #221: Full-harness turns cancelled with `stream disconnected before completion` while their MCP tool calls are still completing.

## Root cause

In launcher mode the ChatGPT browser worker runs in a **separate helper process** spawned by `launcher-helper-client.ts`, while the Codex MCP broker runs in the daemon. `BrowserTurn.externalProgress` was never forwarded across that boundary — `browser-helper-main.ts` built its `BrowserTurn` without the field, and the client sent no progress frames.

So in the helper `turn.externalProgress` was always `undefined`, and `chatGptExternalProgressIsLive(undefined, …)` returns `false` on its first line. **Every liveness guard was dead code in the mode most users run.** The guard at `browser-worker.ts:3511` exists and is correct; it simply never fired.

That explains the reported timings: one turn failed 29.8s after its last recorded tool progress, well inside the 60s `CHATGPT_RESPONSE_DOM_GRACE_MS` window, and another completed a tool call *one second after* the harness declared the response DOM never existed.

## What this changes

**Transport.** Mirror daemon progress into the helper via a `progress` frame and `ChatGptMirroredTurnProgress`. `BrowserTurn.externalProgress` is now a read-only `ChatGptTurnProgressReader`, since the worker only observes. Mirrors reject replayed, regressing, and malformed frames.

**Liveness vetoes terminal verdicts.** All three `ChatGptTurnDomHealthTracker` branches — missing response, empty completion, missing completed-turn action — assert that ChatGPT stopped producing this turn, which an in-flight tool call disproves. While live, every window is cleared; each restarts fresh once liveness lapses, so a live stretch is never charged against the grace. `sawResponse` still records, preserving "disappeared" vs "never created".

**"Stopped thinking" defers to progress**, and the window is *cleared* rather than merely ignored — suppressing only the verdict let it accrue during a tool call and fire the instant progress ended.

**Bounded by staleness, not duration.** An outstanding call reports liveness regardless of age, so a stuck call is bounded by silence since last activity (`CHATGPT_EXTERNAL_PROGRESS_STALL_CEILING_MS`). Nothing in this repo sets `turnTimeoutMs`, so turns have no deadline by default; bounding total duration instead would have cancelled legitimate multi-hour agent turns. The same rule now bounds the pre-response wait, which previously deferred unconditionally while a call was outstanding.

**Multipart staging** gets the progress reader in both its discovery and acknowledgement paths; it previously had no guard at all.

**Answer Markdown is no longer misclassified as commentary.** Commentary was any Markdown root with *some* status container after it, which holds only while a turn has one live status. Once a second tool call opened another status container below already-emitted answer text, that answer was reclassified. Verified in a real browser, old rule vs fix:

| layout | old | fixed |
|---|---|---|
| Pro commentary, live status, answer | `ANSWER` | `ANSWER` |
| commentary nested inside status, answer | `ANSWER` | `ANSWER` |
| status, answer | `ANSWER` | `ANSWER` |
| status, answer, **second tool status** | `` (empty) | `ANSWER` |
| status, part one, second status, part two | `PART TWO` | `PART ONE \| PART TWO` |

The empty row reproduces the reported `textChars=0`. The interleaved row is worse and had not been noticed: **answer content emitted between tool calls was silently dropped.** Commentary now means Markdown preceding the *first* status container, plus chain-of-thought containment as a position-independent signal, since position alone cannot separate "commentary between two statuses" from "answer between two tool calls".

**An accepted turn survives internal faults.** A `TypeError` while reading the page is a defect in this worker, not evidence about ChatGPT, and failing on one destroys an accepted turn that is never resent. A bounded consecutive run is retried; faults raised *after* observation succeeded belong to consumers and still fail immediately. Deliberate signals — adapter errors, aborts, closed tabs, DOM-health verdicts — are untouched.

**Protocol compatibility.** The launcher advertises the helper inside its signed application bundle while the daemon runs from a versioned runtime directory, so the two update independently. Three changes: the daemon prefers the helper beside its own entrypoint, keeping both halves on one build; the helper advertises optional frames on `ready` and the daemon withholds anything unadvertised; and unrecognised frames are answered with a protocol error instead of being routed to the run handler. That last path is how a newer daemon against an older helper produced `Cannot read properties of undefined (reading 'traceId')`, reported as "ChatGPT failed after accepting the Web prompt".

**Completion is vetoed too, not just DOM health.** `ChatGptCompletionTracker` concluded a turn had finished whenever the composer looked idle. During a tool call the composer *is* idle — ChatGPT is waiting on the harness — so a live progress mirror now vetoes that conclusion as well. Without this, the DOM-health veto merely moved the failure one tracker over.

**A turn-abort notice is context, not the next instruction.** Interrupting a turn makes Codex send `<turn_aborted>…</turn_aborted>` as the following user message. `contextualUserMessage` did not recognise it, so it was staged as a fresh prompt carrying the aborted turn's `turn_id`, and every later turn in the thread failed with `ChatGPT web current user message conflicts with native Codex turn_id metadata`. One interrupt poisoned the whole session. The notice is now classified as context like the other Codex-generated wrappers.

**A Bigger Context part that stalls is restaged.** Multipart staging sends the payload in parts and waits for each to be acknowledged. When ChatGPT settled into "Stopped thinking" on a part, the turn was cancelled outright even though nothing had been committed yet and the part was safe to resend. Attach, send, and acknowledge are now wrapped in a bounded retry (`MAX_CHATGPT_MULTIPART_STAGE_ATTEMPTS = 3`) that re-sends only on that cancellation, leaving every other error to fail immediately. Reproducing the stall itself was not possible — the same prompt succeeded at default, Pro, and 136k-char payloads across 21 samples — so this bounds the symptom rather than removing its cause.

**Staged sends get a budget sized for their payload.** Staged parts reused the ordinary 20s `send` budget, but a part is two orders of magnitude larger than a normal prompt and ChatGPT's composer takes proportionally longer to accept it. Observed stage-2 sends timed out at exactly 20001ms and 20002ms — the budget, not the site. Staged sends and the multipart commit now use `multipartStageSend: 180_000`; single-part turns keep the 20s budget.

## A second failure class: turns cancelled while healthy

The bridge cancels any turn whose adapter emits no event for `DEFAULT_STALL_TIMEOUT_SEC`, reporting `upstream_stall_timeout`. The chatgpt-web adapter's only periodic keep-alive was armed part-way through `runTurn`, after several unbounded awaits, so every path reaching those awaits was invisible to the watchdog. Two are reachable in ordinary use, both measured against the real adapter:

- A turn parked in `getOrCreateAfterOwnerRetirement`, waiting for the previous owner of its thread to physically settle, emitted **nothing at all for 25 seconds**.
- A structured compaction turn returns before the interval is ever armed, so a whole summarizing browser turn runs behind a single one-shot heartbeat.

Both exceed five minutes routinely on Pro. `runTurn`'s body moved to `runChatGptWebTurn`; `runTurn` now arms the heartbeat as its first act and clears it in a `finally`. The body cannot reach the timer, so no future path can quietly become a silent one. Read with `git diff -w` the change is 34 lines — the rest of that file's churn is dedent.

The cancellation also wrote nothing to any log, which is why the two windows had to be found by reading source rather than evidence. It now records the last adapter event, its age, the event count and the stream position, and warns at half the budget so a keep-alive gap surfaces before it costs a turn. `stallTimeoutSec` was declared on the bridge options and read when resolving the budget but never passed by any caller; it is now reachable from provider config.

## An upstream reset after the turn had already finished

ChatGPT's backend resets the native Codex connection rather than closing it. Forwarded verbatim that reached Codex as a truncated body and the opaque `Transport error: network error: error decoding response body`. On this machine it lands at a byte-identical offset across independent turns — 111876 bytes, on chunk counts from 10 to 53 — which is a deterministic upstream close, not a flaky network.

A reset arriving after the stream has already delivered `data: [DONE]` is an unclean TCP close on a turn that completed, so the client stream is now closed normally. A reset before that genuinely truncated the turn and is still raised: fabricating a terminal event there would tell Codex a turn ended when it did not, turning a retryable transport failure into a silently short answer.

## A composer insert that raced ChatGPT's focus

Attaching a staged part failed with `ChatGPT composer rejected the plain-text editing command` 84ms after effort selection closed its menu. Nothing had rejected anything — the guard required the caret to already sit inside the composer, and that soon after a menu closes it often does not. The caret is now placed explicitly; an existing collapsed caret inside the composer is left where it is, so sequential staged parts still append correctly, and only a missing or foreign one is replaced. The page-side logic moved to `insertPlainTextIntoComposer` so it can be tested against the focus and selection states the effort menu actually leaves behind.

## A staged part that was merely slow

A staged part was sent, ChatGPT accepted it, and 72 seconds later the turn died with `ChatGPT accepted the message but did not expose its assistant turn in the DOM`. Nothing had gone wrong — a staged part is two orders of magnitude larger than an ordinary prompt and ChatGPT reads all of it before answering. The same payload acknowledged at 19s on one attempt and 30s on the next, so the ordinary 60s DOM grace is not a budget staging can be held to.

The window matters more here than anywhere else: a part still being ingested has produced no MCP activity, so nothing else can vouch for the turn and this grace is the only thing between a slow ingest and a cancellation. Staged parts now get the same budget as the staged send, which bounds the other half of the same oversized exchange. Ordinary turns keep the 60s grace unchanged.

## System sleep no longer costs turns

pmset matched a whole failure cluster to the second: the Mac entered Idle Sleep 41 seconds after two turns started, and on each DarkWake both died at once — `effort_selection` "timed out" at 901s of a 120s budget, both tabs reaped as `orphan_turn_reaped`, heartbeats rejected with `was already released`, and the retries then failing on pages the reaper had closed. Sleep freezes both sides of the lease protocol together, so the launcher's first sweep after a wake saw fifteen-minute-stale heartbeats and read them as a dead helper.

Three layers, in order of preference. The launcher holds a `prevent-app-suspension` power blocker exactly while a browser turn is running, so an idle Mac no longer sleeps mid-turn at all (verified live: `pmset -g assertions` shows the Electron assertion up during a turn and the release logged the moment it ends). When sleep happens anyway, the wake side stops punishing it: the lease sweep knows its own cadence, treats a gap of many multiples of it as its own suspension, and re-baselines every running lease instead of reaping — a helper that is genuinely gone still dies on the ordinary sixty-second cadence. And stage budgets stop charging for slept time: a suspension clock watches for tick gaps, and a stage timer that fires with slept time on the books re-arms for what the stage is still owed awake.

## The suite is green

Seven tests failed on every macOS checkout, including `main`, which is enough noise to hide a regression. Six were the retained-compaction fixtures: they bind the turn broker to a Unix socket under `TMPDIR`, and macOS puts that at `/var/folders/<32 chars>/T`, so the socket path came to 106 bytes against a 104-byte `sun_path` limit and `listen()` failed with nothing but "Failed to listen". The fixtures root themselves somewhere short, and the broker now rejects an over-long path with a message naming the length and the limit — the same wall a real install with a long runtime directory would hit. The seventh is Linux-only and already declared `skip`, which Bun's `node:test` shim ignores; it returns early instead.

**678 tests, 0 failures.**

## Note on the cot-v5 locator hypothesis

#221 speculated that `cot-v5-*` rendering had broken the response *locator*. Instrumenting live turns shows that does not hold: during the `cot-v5` phase there are zero `.markdown` roots, so `visibleText` of `0` is correct — the chain-of-thought is reasoning, not the answer, and the answer has not been written yet.

```
cot:2  mdRoots:0  statusContainers:1  visibleTextChars:0    stopBtn:1  copyBtn:0   <- thinking / tool phase
cot:0  mdRoots:1  statusContainers:0  visibleTextChars:139  stopBtn:1  copyBtn:0   <- answer streaming
cot:0  mdRoots:1  statusContainers:0  visibleTextChars:262  stopBtn:0  copyBtn:1   <- complete
```

`ui.response.textChars` counts the whole response element including CoT status text, a different measure from the tracker's `visibleText`. The reported `textChars=0` therefore has two causes: this correct-by-design reading, and the genuine misclassification above once a second tool call starts.

## Testing

- `bun run typecheck` clean. `bun test`: 656 pass; the 7 failures are identical to `main` on this machine (Linux `/proc` AppImage test plus pre-existing compaction tests).
- Behavioural tests for mirror apply/staleness/validation/waiters, staleness-bounded suppression, the terminal-branch veto, `sawResponse` preservation, the Stopped-thinking window reset, the completion veto, future-dated timestamps, and the fault budget.
- Wiring-contract tests for both ends of the progress transport, helper pinning, capability negotiation, and unknown-frame handling.
- The commentary classifier is now covered: the test extracts the shipped source between two sentinels in `browser-worker.ts` and executes it against a real DOM built with `@mixmark-io/domino`, so the five layouts in the table above are asserted on the code that actually runs, not a copy of it.
- Every fix was mutation-tested: reverting it makes its test fail.

### Verified in production

Built and installed into a real launcher, then driven with live multi-tool Full-harness turns. The two most recent fixes both fired and recovered rather than being merely theoretical:

```
turn 8f30e4e38202 restaging multipart part 1/3 (attempt 1/3):
    ChatGPT remained in 'Stopped thinking' for 5 seconds, so the Codex turn was cancelled.
turn 8f30e4e38202 stage=multipart_stage_1_send completed durationMs=13316
turn 8f30e4e38202 multipart part 1/3 submission accepted evidence=user_turn
```

```
turn c9bd3e0bf01b stage=multipart_stage_2_send completed durationMs=41853
turn c9bd3e0bf01b stage=send             completed durationMs=41575
turn c9bd3e0bf01b completed (markdownChars=9686, domFullScans=90, domCacheHits=302)
```

Both sends exceed the old 20s budget and are exactly the `multipart_stage_2_send` timeout reported on #221. That turn ran five minutes across eight MCP tool calls, all `completed pending=0`. Across the session: zero `stream disconnected`, zero `traceId` protocol errors, zero unadvertised-frame rejections, zero turns lost to a missing progress mirror.

**Remaining caveats:** the CoT-containment signal assumes Pro commentary renders inside a `cot-v5` component — consistent with the testids in the reports on #221, but not confirmed against captured Pro markup. If Pro emits top-level commentary outside CoT between two status containers, it would be appended to the answer rather than dropped; that direction was chosen deliberately, since leaking commentary is visible and recoverable while losing answer text is not.

The "Stopped thinking" stall on a staged part is bounded, not explained — it could not be reproduced on demand, so the retry is a mitigation. And the 180s staged-send budget is sized from observed sends of ~42s with headroom; a pathologically slow upload would still time out, now after three minutes instead of twenty seconds.

### DIFF
--- a/launcher/electron/browser-host.cjs
+++ b/launcher/electron/browser-host.cjs
@@ -1,7 +1,7 @@
 const fs = require("node:fs");
 const path = require("node:path");
 const { randomBytes } = require("node:crypto");
-const { WebContentsView, shell } = require("electron");
+const { WebContentsView, powerMonitor, powerSaveBlocker, shell } = require("electron");
 const { writePrivateFileAtomic } = require("./atomic-file.cjs");
 const {
   runBrowserHelperOperation,
@@ -9,6 +9,11 @@ const {
 } = require("./browser-helper-verifier.cjs");
 const { validateConnectorName } = require("./connector-identity.cjs");
 const { processRunning } = require("./process-tree.cjs");
+const {
+  refreshTurnLeasesAfterSuspension,
+  shouldBlockSleepForTurns,
+  sweepGapIndicatesSuspension,
+} = require("./turn-suspension.cjs");
 const {
   browserViewVisible,
   constrainBrowserBounds,
@@ -240,8 +245,16 @@ class BrowserHost {
     this.authView = null;
     this.authNavigationError = null;
     this.homeNavigationTimeout = null;
+    this.lastTurnSweepAt = Date.now();
+    this.powerSaveBlockerId = null;
     this.turnLeaseSweep = setInterval(() => this.reapExpiredTurnTabs(), TURN_HEARTBEAT_SWEEP_MS);
     this.turnLeaseSweep.unref?.();
+    this.resumeListener = () => this.refreshTurnLeases("system_resume");
+    if (powerMonitor && typeof powerMonitor.on === "function") {
+      powerMonitor.on("resume", this.resumeListener);
+    } else {
+      this.resumeListener = null;
+    }
     this.boundsReady = false;
     this.bounds = { x: 0, y: 0, width: 1, height: 1 };
     this.state = {
@@ -362,6 +375,7 @@ class BrowserHost {
       lastHeartbeatAt: Date.now(),
     };
     this.turnTabs.set(id, tab);
+    this.syncPowerSaveBlocker();
     this.window.contentView.addChildView(view);
     this.presentTurnView(tab, false);
     view.webContents.setZoomFactor(this.state.zoomFactor);
@@ -479,6 +493,7 @@ class BrowserHost {
         (error) => {
           tab.status = "error";
           tab.message = `Browser ownership failed: ${error instanceof Error ? error.message : String(error)}`;
+          this.syncPowerSaveBlocker();
           this.publishState?.(this.snapshot());
         },
       );
@@ -837,7 +852,42 @@ class BrowserHost {
     return this.snapshot();
   }
 
+  refreshTurnLeases(reason, now = Date.now()) {
+    const refreshed = refreshTurnLeasesAfterSuspension(
+      [...this.turnTabs.values()],
+      now,
+      TURN_TAB_BOOTSTRAP_TIMEOUT_MS,
+    );
+    if (refreshed.length > 0) {
+      this.logger.warn("browser.turn_leases_refreshed_after_suspension", { reason, traceIds: refreshed });
+    }
+  }
+
+  syncPowerSaveBlocker() {
+    // Under plain Node (the launcher test harness) require("electron") exposes no APIs; the
+    // blocker is an Electron-only concern and its absence must not break lease bookkeeping.
+    if (!powerSaveBlocker || typeof powerSaveBlocker.start !== "function") return;
+    const wanted = shouldBlockSleepForTurns([...this.turnTabs.values()]);
+    const active = this.powerSaveBlockerId !== null && powerSaveBlocker.isStarted(this.powerSaveBlockerId);
+    if (wanted && !active) {
+      this.powerSaveBlockerId = powerSaveBlocker.start("prevent-app-suspension");
+      this.logger.info("browser.sleep_blocked_for_turns", { blockerId: this.powerSaveBlockerId });
+    } else if (!wanted && active) {
+      powerSaveBlocker.stop(this.powerSaveBlockerId);
+      this.logger.info("browser.sleep_block_released", { blockerId: this.powerSaveBlockerId });
+      this.powerSaveBlockerId = null;
+    }
+  }
+
   reapExpiredTurnTabs(now = Date.now()) {
+    const lastSweepAt = this.lastTurnSweepAt;
+    this.lastTurnSweepAt = now;
+    if (sweepGapIndicatesSuspension(lastSweepAt, now, TURN_HEARTBEAT_SWEEP_MS)) {
+      // The launcher itself was frozen, so missing heartbeats are evidence of the sleep, not of a
+      // dead helper. Reaping here is what turned every system sleep into a lost turn.
+      this.refreshTurnLeases("sweep_gap", now);
+      return;
+    }
     for (const tab of [...this.turnTabs.values()]) {
       if (tab.status === "ready") {
         if (now - (tab.lastHeartbeatAt ?? 0) < RETAINED_TURN_TAB_TTL_MS) continue;
@@ -968,6 +1018,7 @@ class BrowserHost {
   removeTurnTab(tab, abortRunning) {
     if (!this.turnTabs.has(tab.id)) return;
     this.turnTabs.delete(tab.id);
+    this.syncPowerSaveBlocker();
     if (abortRunning && tab.status === "running") {
       this.closedTurnOwners.set(tab.traceId, tab.helperPid);
       tab.status = "aborted";
@@ -1374,6 +1425,7 @@ class BrowserHost {
     }
     const cancelledByUser = this.userCancelledTurnOwners.get(traceId) === helperPid;
     tab.status = status === "completed" ? "ready" : status === "aborted" ? "aborted" : "error";
+    this.syncPowerSaveBlocker();
     tab.message = status === "completed" ? "Task completed" : message || `ChatGPT turn ${status}`;
     tab.loading = false;
     if (!tab.view.webContents.isDestroyed()) tab.view.webContents.setBackgroundThrottling(true);
@@ -1779,6 +1831,14 @@ class BrowserHost {
     this.closeAuthView(this.authView, true);
     this.clearHomeNavigationTimeout();
     if (this.turnLeaseSweep) clearInterval(this.turnLeaseSweep);
+    if (this.resumeListener && powerMonitor && typeof powerMonitor.removeListener === "function") {
+      powerMonitor.removeListener("resume", this.resumeListener);
+      this.resumeListener = null;
+    }
+    if (this.powerSaveBlockerId !== null && powerSaveBlocker && typeof powerSaveBlocker.stop === "function") {
+      powerSaveBlocker.stop(this.powerSaveBlockerId);
+      this.powerSaveBlockerId = null;
+    }
     for (const tab of this.turnTabs.values()) {
       try { this.window.contentView.removeChildView(tab.view); } catch {}
       if (!tab.view.webContents.isDestroyed()) tab.view.webContents.close();
--- a/launcher/electron/turn-suspension.cjs
+++ b/launcher/electron/turn-suspension.cjs
@@ -1,0 +1,52 @@
+"use strict";
+
+/**
+ * System sleep freezes both sides of the turn-lease protocol at once: the helper cannot send
+ * heartbeats and the launcher cannot observe them. On wake every stalled timer fires together, and
+ * without these guards the first sweep after a sleep reaped live turns as orphans — pmset logged
+ * 'Idle Sleep' 41 seconds into two running turns, and both died on the next DarkWake to the second.
+ */
+
+/**
+ * The turn-lease sweep runs on a short fixed interval for the whole life of the launcher. A gap
+ * between consecutive sweeps far larger than that interval means the process was suspended, not
+ * busy: nothing the launcher does blocks its event loop for multiples of the sweep interval.
+ */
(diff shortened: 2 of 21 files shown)


miuuyy:
The core finding is valid and important. I independently ran the current HEAD through the full Bun 1.4.0 verification gate sequentially; runtime tests, launcher tests, typechecks, audits, build, and relocatable smoke all passed.

I’m not merging this shape yet because it now combines several independently risky lifecycle surfaces in one patch: helper progress transport and DOM completion, answer classification, abort-history handling, Bigger Context timing/retry behavior, bridge stall policy, native stream reset handling, composer focus, macOS sleep/lease behavior, and test-platform cleanup.

Please split it into two reviewable PRs based on current `main`:

1. the #221 P1 path: progress transport, liveness/completion vetoes, answer-classifier correction, helper protocol pinning/negotiation, and the directly related regression tests;
2. the operational follow-ups: compaction timing, bridge watchdog diagnostics/config, post-DONE reset handling, composer focus, sleep/lease handling, and platform-test cleanup.

One piece should not be carried into either PR as written: automatically re-staging a part after `Stopped thinking`. The PR itself correctly says the cause was not reproduced and this is a mitigation. Until the cause is established, that condition should remain an explicit failure rather than an automatic resend.

This is a request to reduce the review and release blast radius, not a rejection of the core contribution.

1saifj:
Split as requested, both based on current `main`:

- #232 — the #221 P1 path: progress transport, liveness and completion vetoes, answer-classifier correction, helper protocol pinning/negotiation, and the directly related regression tests. 10 commits.
- #233 — the operational follow-ups: Bigger Context timing, bridge watchdog diagnostics/config, post-`[DONE]` reset handling, composer focus, sleep/lease handling, and platform-test cleanup. 10 commits.

The automatic re-staging after "Stopped thinking" is dropped from both. That condition is back to an explicit failure and stays that way until the cause is established.

Two notes on the split itself:

The abort-history change (`<turn_aborted>` classified as context) was named in the review but not assigned to either half. It is in #233 because it is not on the liveness path — it is an isolated four-line change in `environment.ts`. Happy to move it to #232 if you would rather it travel with the P1 work.

Platform-test cleanup went to #233 as listed, so #232 carries the same seven macOS failures that `main` does — six compaction fixtures over the `sun_path` limit plus the Linux-only AppImage test. #232 is no worse than `main` on that front, but if you would rather review it green, that commit moves to #232 in one cherry-pick.

Verification on each branch separately: #232 typecheck clean, 646 pass / 7 fail (the `main` seven). #233 typecheck clean, 659 pass / 0 fail.

This PR can be closed in favour of the two.

miuuyy:
Closing this original combined PR now that it has been split into #232 and #233 for separate review. Thanks for doing the split — we'll continue the review on those two PRs.

PR #226, opened 2026-08-30
fix: disconnect stalled browser before launcher rebind

kocourekwork-sudo:
## Summary

Disconnect a stalled launcher Playwright/CDP observation transport before opening the same-page recovery connection.

The existing post-submit DOM observation timeout is intentionally bounded, but it races the Playwright operation rather than cancelling the underlying `page.evaluate()` itself. The current recovery path leaves that old CDP connection alive while it tries to open and validate a replacement, then closes the old connection only after the rebind succeeds.

This changes the ordering to:

1. preserve the existing launcher surface identity
2. detach and close the stale observation transport
3. open a fresh CDP connection to the same launcher-owned surface
4. prove the operational viewport
5. continue observing the already-submitted response

No prompt is resent and the ChatGPT document is not reloaded.

## Real-world evidence

Observed on Codex Web GPT v4.0.5 on Windows during long-running Full Harness / retained-compaction work.

Safe diagnostics contained several `response_page_rebind_1` incidents:

- successful recovery in ~7.5 s
- failed recovery in ~37.5 s with `ChatGPT browser surface did not expose an operational viewport`
- two successful recoveries in one long-running trace in ~29.6 s and ~21.5 s
- a later recovery in that same trace timed out at the full 60 s stage limit

The final failure was:

`ChatGPT browser stage timed out: response_page_rebind_1`

The observation probe immediately before it had already reported:

`ChatGPT browser DOM observation did not respond within 5000ms`

## Why this is scoped

For launcher connections, Playwright `Browser.close()` on a `connectOverCDP` browser disconnects the local observation transport; it does not close or reload the launcher-owned Electron browser page.

The patch therefore does not:

- resend an accepted prompt
- reload the retained ChatGPT conversation
- allocate a replacement ChatGPT conversation
- change connector approval behavior
- weaken submission evidence
- force completion

## Regression coverage

The existing stalled-DOM recovery contract now also proves source ordering:

`previous connection -> detach -> close -> reconnect stage -> connectLauncherBrowserHost`

It continues to assert that recovery never calls `sendAttachedPrompt()` again.

## Verification

- browser-worker contract suite passed locally
- TypeScript `tsc --noEmit` passed locally
- `git diff --check` passed

### DIFF
--- a/src/adapters/chatgpt-web/browser-worker.ts
+++ b/src/adapters/chatgpt-web/browser-worker.ts
@@ -3195,6 +3195,18 @@ export class ChatGptBrowserWorker {
           + ` ${redactChatGptUiDiagnostic(cause.message)}`,
         );
         const previousConnection = turnConnection;
+        if (previousConnection) {
+          // The observation timeout races the Playwright operation but cannot cancel the
+          // underlying page.evaluate by itself. Disconnect the stale CDP transport before
+          // opening its replacement so the hung probe cannot contend with its own recovery.
+          turnConnection = undefined;
+          await previousConnection.close().catch(error => {
+            console.warn(
+              `[chatgpt-web] previous browser observation connection did not close before rebind:`
+              + ` ${error instanceof Error ? error.message : String(error)}`,
+            );
+          });
+        }
         const connection = await this.runStage(
           turn.traceId,
           `response_page_rebind_${attempt}`,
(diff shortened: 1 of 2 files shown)


kocourekwork-sudo:
Live verification after installing the source-built v4.0.5 hotfix: a real post-submit DOM observation stalled for the 5s probe window, `response_page_rebind_1` then completed in ~17.4s, the same already-submitted browser turn resumed, completed normally a few seconds later, and the retained tab was released as completed. No prompt resend or document reload was needed. This is the first end-to-end confirmation from the original long-running Full Harness workload that the close-before-reconnect ordering recovers an actual stalled renderer.

miuuyy:
The recovery order is correct: the timed-out Playwright operation can still be alive, so the stale CDP connection must be disconnected before opening the replacement.

One invariant still needs tightening before merge. `previousConnection.close()` currently catches, logs, and continues. If that close fails, the old probe can still contend with the new connection — the exact condition this PR is meant to eliminate. Please fail the rebind explicitly in that case instead of continuing.

A focused behavioral test should make the old connection’s `close()` reject and prove that no replacement `connectLauncherBrowserHost()` attempt starts afterward. With that change, this becomes an accept candidate.

miuuyy:
Merged, thank you. Maintainer follow-up 3e1a1eb makes a failed stale-connection close terminal before any replacement connection can start, and adds the requested behavioral regression test. The complete Bun 1.4.0 repository verification passes.

PR #231, opened 2026-08-30
fix: preserve native AgentMessage routing semantics

### LINKED ISSUE #203: Preserve native AgentMessage author/recipient semantics in Web context
## Native-parity gap

Current Codex has a distinct Responses input item:

`{ type: "agent_message", author, recipient, content }`

The Web parser currently detects `agent_message`, checks for opaque encrypted content, then converts every plaintext agent message into an ordinary `role: "user"` message and discards `author` and `recipient`.

That loses native MultiAgent V2 conversation semantics even when the payload is perfectly readable. The model can no longer tell which agent sent the message or which agent was addressed, and the accumulated Web conversation is not equivalent to the native Responses history.

## Proposed focused fix

For plaintext agent messages:

- preserve `author` and `recipient` in the internal user-message representation (or an equally lossless dedicated representation);
- compile them into a distinct `<codex_context_json>` message envelope such as `role: "agent_message"` with the original `author`, `recipient`, and content;
- keep ordinary user messages unchanged;
- preserve this identically in inline and multipart context transports;
- keep the existing fail-closed rejection for provider-private encrypted agent_message content, since ChatGPT Web cannot decrypt OpenAI backend ciphertext.

Regression tests should prove author/recipient/content round-trip and prove no metadata is invented for ordinary user input.

David-T-Campos:
Closes #203.

Plaintext Responses `agent_message` items are native inter-agent inputs, not human user messages. The current Web parser collapses them to `role: "user"`, drops `author` / `recipient`, and can invent fallback text for empty content.

This keeps that native meaning through the full parser-to-compiler path:

- adds a distinct internal AgentMessage representation;
- preserves `author`, `recipient`, and content verbatim when present;
- serializes it to the Web context as `role: "agent_message"` instead of a human `user` turn;
- keeps ordinary user messages unchanged and does not invent routing metadata;
- uses the same envelope in inline and multipart context transports;
- retains the existing fail-closed `_opaqueMultiAgentV2Payload` path for provider-private encrypted AgentMessage content.

Validation on the exact proposed tree:

- focused parser-to-prompt / collaboration / compaction tests: passed
- `bunx tsc --noEmit`: passed
- branch is one clean commit on current upstream `main` (`34336fb2deb5d17bcf12be74c6573c2fb26ee0ed`)
- no temporary Gauntlet workflow files are included in the proposed tree

### DIFF
--- a/src/adapters/chatgpt-web/prompt.ts
+++ b/src/adapters/chatgpt-web/prompt.ts
@@ -284,6 +284,14 @@ function messageEnvelope(
       content: assistantContent(message.content),
     };
   }
+  if (message.role === "agentMessage") {
+    return {
+      role: "agent_message",
+      ...(message.author !== undefined ? { author: message.author } : {}),
+      ...(message.recipient !== undefined ? { recipient: message.recipient } : {}),
+      content: inputContent(message.content, images, budget),
+    };
+  }
   return { role: message.role, content: inputContent(message.content, images, budget) };
 }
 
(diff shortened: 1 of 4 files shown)


miuuyy:
Thanks — this is a solid and well-scoped implementation. I compared it line by line with the pending local release work, and the core fix for #203 was already implemented in local commit `15392d5` at 2026-08-30 04:13 UTC, around 13 hours before this PR was authored. That implementation already preserves the distinct AgentMessage role, author/recipient, inline and multipart serialization, and regression coverage.

I have now completed the two remaining edges in the local patch as well: literal empty content is preserved, and the prompt contract explicitly distinguishes inter-agent input from human user text. The exact-version full verification gate passes.

Because the pending release implementation already covered substantially the same change, I am closing this rather than merging duplicate code. Sorry about the timing overlap — this is not a criticism of the quality of the contribution.

Issue #220, opened 2026-08-30
codex-chatgpt-web: Codex openai_base_url changed after setup


1saifj:
<img width="1098" height="700" alt="Image" src="https://github.com/user-attachments/assets/1ffba9e7-b0fe-4b38-83d1-2870b8854a46" />

miuuyy:
Thanks for the report. The screenshot shows two related conditions: the Launcher is v4.0.5 while the saved runtime configuration was created for v2.1.11, and Codex's `openai_base_url` has changed since the original setup. The Launcher refuses to overwrite that newer value intentionally.

Could you please confirm:

1. Did you upgrade directly from v2.1.11 to v4.0.5?
2. Are you using any other Codex extensions or wrappers that may change model routing or the bridge port (for example OpenCodex or Headroom), or did you edit `openai_base_url` manually?
3. Please attach the Launcher logs from the failed setup attempt.

If possible, also share the current `openai_base_url` line from your Codex config, with any private hostnames or tokens redacted. That should let us distinguish a stale upgrade state from an active routing conflict.

1saifj:
Confirmed from the local state:

- The saved launcher configuration is still `2.1.11`, while the launcher/runtime being started is `4.0.5`.
- The active integration journal is version 6 and expects `openai_base_url = "http://127.0.0.1:17841/v1"`.
- The current Codex config has no top-level `openai_base_url` assignment.
- The journal preserves an older route to `http://127.0.0.1:4102/_codex-router/[redacted]/v1`, so another local router or a manual/configuration change previously owned the route. I did not edit this value during the investigation, and I cannot identify the wrapper from the redacted route alone.
- `route status` reproduces the same refusal: `Codex openai_base_url changed after setup`.

I exported the launcher diagnostics through the privacy-safe exporter and uploaded the sanitized file here:

https://gist.github.com/1saifj/9a6fd54e49e15ff6f767e7df6ee4aaf6

miuuyy:
Thanks for providing the detailed diagnostics. I have not encountered this during my own upgrades between Launcher versions, and the logs show that the previously managed `openai_base_url` had been removed or changed while the old integration journal still considered it active. The later clean setup completed successfully.

[...]

PR #233, opened 2026-08-30
fix: operational follow-ups for chatgpt-web turn lifecycle

1saifj:
Split of #222, part 2 of 2, per review. Based on current `main`.

Operational follow-ups only: Bigger Context timing, bridge watchdog diagnostics and config, post-`[DONE]` reset handling, composer focus, macOS sleep/lease handling, and platform-test cleanup. The #221 liveness path is #232.

The automatic re-staging after "Stopped thinking" is **not** in this PR. As requested, that condition remains an explicit failure until the cause is established.

## Turns cancelled while healthy

The bridge cancels any turn whose adapter emits no event for `DEFAULT_STALL_TIMEOUT_SEC`, reporting `upstream_stall_timeout`. The chatgpt-web adapter's only periodic keep-alive was armed part-way through `runTurn`, after several unbounded awaits, so every path reaching those awaits was invisible to the watchdog. Two are reachable in ordinary use, both measured against the real adapter:

- A turn parked in `getOrCreateAfterOwnerRetirement`, waiting for the previous owner of its thread to physically settle, emitted nothing at all for 25 seconds.
- A structured compaction turn returns before the interval is ever armed, so a whole summarizing browser turn runs behind a single one-shot heartbeat.

Both exceed five minutes routinely on Pro. `runTurn`'s body moved to `runChatGptWebTurn`; `runTurn` now arms the heartbeat as its first act and clears it in a `finally`. The body cannot reach the timer, so no future path can quietly become a silent one. Read with `git diff -w` the change is 34 lines — the rest of that file's churn is dedent.

The cancellation also wrote nothing to any log, which is why the two windows had to be found by reading source rather than evidence. It now records the last adapter event, its age, the event count and the stream position, and warns at half the budget so a keep-alive gap surfaces before it costs a turn. `stallTimeoutSec` was declared on the bridge options and read when resolving the budget but never passed by any caller; it is now reachable from provider config.

## An upstream reset after the turn had already finished

ChatGPT's backend resets the native Codex connection rather than closing it. Forwarded verbatim that reached Codex as a truncated body and the opaque `Transport error: network error: error decoding response body`. On this machine it lands at a byte-identical offset across independent turns — 111876 bytes, on chunk counts from 10 to 53 — which is a deterministic upstream close, not a flaky network.

A reset arriving after the stream has already delivered `data: [DONE]` is an unclean TCP close on a turn that completed, so the client stream is now closed normally. A reset before that genuinely truncated the turn and is still raised: fabricating a terminal event there would tell Codex a turn ended when it did not, turning a retryable transport failure into a silently short answer.

## A composer insert that raced ChatGPT's focus

Attaching a staged part failed with `ChatGPT composer rejected the plain-text editing command` 84ms after effort selection closed its menu. Nothing had rejected anything — the guard required the caret to already sit inside the composer, and that soon after a menu closes it often does not. The caret is now placed explicitly; an existing collapsed caret inside the composer is left where it is, so sequential staged parts still append correctly, and only a missing or foreign one is replaced. The page-side logic moved to `insertPlainTextIntoComposer` so it can be tested against the focus and selection states the effort menu actually leaves behind.

## Staged sends sized for their payload

Staged parts reused the ordinary 20s send budget, but a part is two orders of magnitude larger than a normal prompt and ChatGPT's composer takes proportionally longer to accept it. Observed stage-2 sends timed out at exactly 20001ms and 20002ms — the budget, not the site. Staged sends and the multipart commit now use `multipartStageSend: 180_000`; single-part turns keep the 20s budget.

Separately, a staged part was sent, ChatGPT accepted it, and 72 seconds later the turn died with `ChatGPT accepted the message but did not expose its assistant turn in the DOM`. The same payload acknowledged at 19s on one attempt and 30s on the next, so the ordinary 60s DOM grace is not a budget staging can be held to. Staged parts now get the same budget as the staged send, which bounds the other half of the same oversized exchange. Ordinary turns keep the 60s grace unchanged.

## System sleep no longer costs turns

`pmset` matched a whole failure cluster to the second: the Mac entered Idle Sleep 41 seconds after two turns started, and on each DarkWake both died at once — `effort_selection` "timed out" at 901s of a 120s budget, both tabs reaped as `orphan_turn_reaped`, heartbeats rejected with `was already released`, and the retries then failing on pages the reaper had closed. Sleep freezes both sides of the lease protocol together, so the launcher's first sweep after a wake saw fifteen-minute-stale heartbeats and read them as a dead helper.

Three layers, in order of preference. The launcher holds a prevent-app-suspension power blocker exactly while a browser turn is running, so an idle Mac no longer sleeps mid-turn at all (verified live: `pmset -g assertions` shows the Electron assertion up during a turn and the release logged the moment it ends). When sleep happens anyway, the wake side stops punishing it: the lease sweep knows its own cadence, treats a gap of many multiples of it as its own suspension, and re-baselines every running lease instead of reaping — a helper that is genuinely gone still dies on the ordinary sixty-second cadence. And stage budgets stop charging for slept time: a suspension clock watches for tick gaps, and a stage timer that fires with slept time on the books re-arms for what the stage is still owed awake.

## An abort notice is context, not the next instruction

Interrupting a turn makes Codex send `<turn_aborted>…</turn_aborted>` as the following user message. `contextualUserMessage` did not recognise it, so it was staged as a fresh prompt carrying the aborted turn's `turn_id`, and every later turn in the thread failed with `ChatGPT web current user message conflicts with native Codex turn_id metadata`. One interrupt poisoned the whole session. The notice is now classified as context like the other Codex-generated wrappers.

This surface was named in review but not assigned to either half of the split. It is here because it is not on the #221 liveness path; say the word and it moves to #232.

## Platform-test cleanup

Seven tests failed on every macOS checkout, including `main`, which is enough noise to hide a regression. Six were the retained-compaction fixtures: they bind the turn broker to a Unix socket under `TMPDIR`, and macOS puts that at `/var/folders/<32 chars>/T`, so the socket path came to 106 bytes against a 104-byte `sun_path` limit and `listen()` failed with nothing but "Failed to listen". The fixtures root themselves somewhere short, and the broker now rejects an over-long path with a message naming the length and the limit — the same wall a real install with a long runtime directory would hit. The seventh is Linux-only and already declared `skip`, which Bun's `node:test` shim ignores; it returns early instead.

## Testing

`bun run typecheck` clean. `bun test`: **659 pass, 0 fail** across 55 files.

## Caveat

The 180s staged-send budget is sized from observed sends of ~42s with headroom; a pathologically slow upload would still time out, now after three minutes instead of twenty seconds.

### DIFF
--- a/launcher/electron/browser-host.cjs
+++ b/launcher/electron/browser-host.cjs
@@ -1,7 +1,7 @@
 const fs = require("node:fs");
 const path = require("node:path");
 const { randomBytes } = require("node:crypto");
-const { WebContentsView, shell } = require("electron");
+const { WebContentsView, powerMonitor, powerSaveBlocker, shell } = require("electron");
 const { writePrivateFileAtomic } = require("./atomic-file.cjs");
 const {
   runBrowserHelperOperation,
@@ -9,6 +9,11 @@ const {
 } = require("./browser-helper-verifier.cjs");
 const { validateConnectorName } = require("./connector-identity.cjs");
 const { processRunning } = require("./process-tree.cjs");
+const {
+  refreshTurnLeasesAfterSuspension,
+  shouldBlockSleepForTurns,
+  sweepGapIndicatesSuspension,
+} = require("./turn-suspension.cjs");
 const {
   browserViewVisible,
   constrainBrowserBounds,
@@ -240,8 +245,16 @@ class BrowserHost {
     this.authView = null;
     this.authNavigationError = null;
     this.homeNavigationTimeout = null;
+    this.lastTurnSweepAt = Date.now();
+    this.powerSaveBlockerId = null;
     this.turnLeaseSweep = setInterval(() => this.reapExpiredTurnTabs(), TURN_HEARTBEAT_SWEEP_MS);
     this.turnLeaseSweep.unref?.();
+    this.resumeListener = () => this.refreshTurnLeases("system_resume");
+    if (powerMonitor && typeof powerMonitor.on === "function") {
+      powerMonitor.on("resume", this.resumeListener);
+    } else {
+      this.resumeListener = null;
+    }
     this.boundsReady = false;
     this.bounds = { x: 0, y: 0, width: 1, height: 1 };
     this.state = {
@@ -362,6 +375,7 @@ class BrowserHost {
       lastHeartbeatAt: Date.now(),
     };
     this.turnTabs.set(id, tab);
+    this.syncPowerSaveBlocker();
     this.window.contentView.addChildView(view);
     this.presentTurnView(tab, false);
     view.webContents.setZoomFactor(this.state.zoomFactor);
@@ -479,6 +493,7 @@ class BrowserHost {
         (error) => {
           tab.status = "error";
           tab.message = `Browser ownership failed: ${error instanceof Error ? error.message : String(error)}`;
+          this.syncPowerSaveBlocker();
           this.publishState?.(this.snapshot());
         },
       );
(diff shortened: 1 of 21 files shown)


1saifj:
@miuuyy Verified both branches separately against current `main`: each merges clean on its own, #232 typecheck clean with the same seven macOS failures `main` has, #233 typecheck clean at 659 pass / 0 fail. Commit and test parity check out — 20 of the original 21 commits, only the restage one dropped.

One thing worth knowing for merge order: #232 and #233 touch the same regions of `browser-worker.ts`, so whichever lands second needs a rebase. Happy to rebase #233 onto `main` once #232 is in.

1saifj:
Repushed this branch after a scope check: one of my conflict resolutions had it passing `turn.externalProgress` at the multipart acknowledgement site, which is #232's liveness wiring, not this PR's concern. It now passes `undefined` there and changes only the grace, so this branch introduces no new progress wiring — same three `turn.externalProgress` uses `main` already has.

I also dry-ran the merge order end to end: merged #232 into `main` locally, then merged this branch on top. Five conflict hunks, all mechanical — the two source ones restore `ChatGptTurnProgressReader` alongside the new `graceMs` default, and put `turn.externalProgress` back next to `CHATGPT_MULTIPART_RESPONSE_DOM_GRACE_MS`. The other three are additive test blocks. Resolved state: typecheck clean, **677 pass / 0 fail**, and it matches the original #222 tree — same five `turn.externalProgress` uses, same two grace references, zero restage references. Happy to do that rebase myself once #232 lands.

miuuyy:
Please stop adding unrelated fixes to this pull request.

I was close to making a decision on this PR, with one remaining problem to resolve, and the branch then gained another production-upgrade change tied to a separate, already closed edge case (#220). That makes it impossible to finish a stable review and keeps increasing the risk of merging the package.

I cannot accept a continuously growing collection of fixes. Please freeze the scope of #233:

- remove the latest unrelated route-conflict commit from this branch;
- make further changes here only when they address an existing review finding or the failing CI check;
- submit any newly discovered independent bug separately, with a clear reproduction and a focused PR.

Some of the individual fixes here are valuable, but they need clear boundaries. Further unrelated scope expansion may result in this PR being closed rather than reviewed again from the beginning.

1saifj:
@miuuyy I removed the unrelated route-conflict change, rebased #233 onto current `main`, and kept the remaining updates within the existing review scope. Local and CI checks are green.

PR #238, opened 2026-08-31
perf: deduplicate retained system prompts safely

xtt22175-svg:
## Summary

- include the canonical system prompt in the retained conversation key
- omit that unchanged system prompt from retained follow-up payloads
- invalidate retained reuse after a main-frame or in-page navigation so the next turn falls back to a fresh full-context tab
- cover stable prompt reuse, prompt changes, and manually navigated retained tabs

## Safety

The full Codex request remains canonical. A system prompt change rotates the conversation key, while any navigation after retention prevents the launcher from returning `reused: true`. Fresh tabs therefore continue through the existing full `prepare()` path.

## Validation

- focused retained conversation tests: pass
- `bun run typecheck`: pass
- `bun run launcher:typecheck`: pass
- `bun run launcher:test`: 215 pass, 2 skipped, 0 fail
- launcher build, runtime bundle, third-party notices, and relocatable runtime smoke: pass
- `bun run verify`: reaches the root test suite with 460 pass / 1 unrelated existing Windows failure in `installed launcher discovery has explicit platform candidates` (`C:\Users\tester` expected, current user path received)

### DIFF
--- a/launcher/electron/browser-host.cjs
+++ b/launcher/electron/browser-host.cjs
@@ -345,6 +345,8 @@ class BrowserHost {
       conversationKey,
       connectorIdentity,
       connectorBound: false,
+      retainedUrl: null,
+      retainedNavigationInvalidated: false,
       helperPid,
       view,
       status: "running",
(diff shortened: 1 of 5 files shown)


miuuyy:
Thanks for the implementation, but the system prompt is only roughly 10k tokens, so the saving is not worth weakening this invariant.  
ChatGPT may silently compact retained chats, which can weaken old instructions and reduce model quality or adherence.  
For that reason, we deliberately resend the prompt every turn and are closing this PR.

PR #252, opened 2026-08-31
Relay native realtime voice WebSocket through the local bridge

nokiyliao:
## Summary

- add an authenticated `/v1/live` WebSocket relay to the native OpenAI realtime endpoint
- preserve the incoming bearer token, attestation headers, query string, and negotiated subprotocol while stripping hop-by-hop handshake headers
- include realtime relays in health, cancellation, drain, shutdown, and active-turn accounting
- leave the existing `/v1/responses` routing behavior unchanged

## Why

When Codex Desktop uses this bridge as its configured OpenAI base URL, realtime voice starts an app-server sideband WebSocket at `/v1/live`. The bridge currently returns `404 Not Found`, so the voice session fails after the WebRTC setup has already started.

## Validation

- `bun run typecheck`
- `bun test tests/native-realtime-passthrough.test.ts tests/server-lifecycle.test.ts` (26 passed)
- packaged local canary reached `realtime_session_started` and received a realtime session ID without another `/v1/live` 404

## Boundaries

This change does not alter model selection, Responses routing, browser automation, or credential storage. The relay fails closed when the incoming Codex bearer credential is absent.

### DIFF
--- a/src/native-realtime-passthrough.ts
+++ b/src/native-realtime-passthrough.ts
@@ -1,0 +1,100 @@
+const NATIVE_REALTIME_URL = "wss://api.openai.com/v1/live";
+const NATIVE_REALTIME_HANDSHAKE_TIMEOUT_MS = 35_000;
+
+const HANDSHAKE_HEADERS = new Set([
+  "connection",
+  "content-length",
+  "host",
+  "sec-websocket-extensions",
+  "sec-websocket-key",
+  "sec-websocket-protocol",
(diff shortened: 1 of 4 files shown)


miuuyy:
Thanks. This proxies realtime WebSocket traffic, but issue #258 is HTTP WebRTC call creation and this handler rejects that request. Closing because it does not fix the reported bug and the credential-proxy surface is not justified by a separate reproduction.

Issue #204, opened 2026-08-29
stream disconnected before completion: Launcher browser control channel failed: HTTP 400: connectorIdentity requires conversationKey


4biddencode:
When I selected the Model ChatGPT Web - Lunar, it retried 5 times then gave me the error stream disconnected before completion: Launcher browser control channel failed: HTTP 400: connectorIdentity requires conversationKey. Logs are attached.

[logs.txt](https://github.com/user-attachments/files/31598180/logs.txt)

miuuyy:
Thanks — the attached log isolates this precisely. This is not a ChatGPT DOM, authentication, rate-limit, or network failure.

The launcher smoke completes successfully with Luna. The failure starts only after Full MCP setup: Luna intentionally has no retained `conversationKey`, while the browser worker still sends `connectorIdentity` because local tools are enabled. The launcher correctly rejects that impossible contract with `connectorIdentity requires conversationKey`, and Codex retries the same deterministic failure.

I am marking this **P1** because it blocks every Full-mode Luna turn for the affected account shape. No further diagnostics are needed right now; the issue should stay open until that lifecycle contract is corrected and tested without weakening the retained-conversation invariant.

Dinesh-Das:
Yup faced similar issue on my run.

Kirkland-Pickles:
same error : stream disconnected before completion: Launcher browser control channel failed: HTTP 400: connectorIdentity requires conversationKey

br0k3nb:
Same here!

miuuyy:
Fixed and released in [v4.0.6](https://github.com/miuuyy/codex-chatgpt-web/releases/tag/v4.0.6). Launcher browser control now sends connectorIdentity only together with its required conversationKey, removing the ordinary Luna HTTP 400 path. Closing this as completed.

4biddencode:
Hey @miuuyy I just updated codex web gpt, still facing the same issue

4biddencode:
Yes confirming this. Even after restarting, ChatGPT, Codex Web GPT, and my mac im still facing these issues.

miuuyy:
Thanks for confirming. I rechecked this against the original trace and the actual published macOS artifacts, not only the repository source.

[...]

PR #253, opened 2026-08-31
Run managed runtime upgrades from the packaged runtime

nokiyliao:
## Summary

- execute `upgradeManagedRuntime()` with the new launcher's embedded runtime
- add regression coverage that asserts the upgrade transaction selects the embedded invocation

## Why

The launcher detects that its managed runtime belongs to an older release, but the upgrade setup command currently runs through the configured old runtime. That old runtime writes its own version and command back into `config.json`, so the launcher appears upgraded while the running daemon remains on the predecessor release.

Using the packaged runtime makes the upgrade monotonic: the newly installed launcher writes its own exact runtime identity into the managed configuration.

## Validation

`node --test launcher/tests/runtime-host.test.cjs launcher/tests/runtime-install.test.cjs launcher/tests/update.test.cjs` (44 passed)

## Boundaries

This does not change update discovery, version comparison, external-runtime ownership, or rollback behavior. It only selects the already packaged runtime for the existing managed upgrade transaction.

### DIFF
--- a/launcher/electron/runtime.cjs
+++ b/launcher/electron/runtime.cjs
@@ -902,6 +902,7 @@ class RuntimeHost {
       message: `Upgrading launcher runtime from ${existing.config.releaseVersion} to ${currentVersion}`,
       successMessage: `Launcher runtime upgraded to ${currentVersion}`,
       timeoutMs: existing.mode === "full" ? MCP_SETUP_TIMEOUT_MS : CORE_SETUP_TIMEOUT_MS,
+      embedded: true,
     });
     return {
       updated: true,
--- a/launcher/tests/runtime-host.test.cjs
+++ b/launcher/tests/runtime-host.test.cjs
@@ -23,11 +23,13 @@ function hostFor(existingConfig) {
     },
   });
   let invocation;
-  host.runSetup = async (name, args) => {
+  let runOptions;
+  host.runSetup = async (name, args, options) => {
     invocation = { name, args };
+    runOptions = options;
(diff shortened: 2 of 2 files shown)


miuuyy:
Thanks for the contribution. I traced the packaged upgrade path and reproduced the invocation selection independently, and the premise of this change does not hold for the launcher-managed runtime.

Before this patch, `RuntimeHost.command()` already calls `runtimeRootProvider()`. That provider runs `ensurePackagedRuntime()`, which validates the new application's version, platform/architecture, and bundle identity, installs that exact bundle under the versioned runtime directory, and returns it before the setup command is constructed. In a reproduction with an older configured runtime and a newer packaged application, the existing code selected the newer durable runtime and its newer `cli.js` without this change.

Using `embedded: true` is not a harmless equivalent. It bypasses the verified versioned runtime path and can cause setup to persist a command that references the replaceable application bundle (or a system Bun executable) instead of the launcher-owned durable runtime under `~/.codex-chatgpt-web/versions`. The added test only asserts that a mocked `runSetup` received the new flag; it never observes which executable or entrypoint actually runs, so it does not validate the claimed regression.

This also cannot explain the follow-up report in #204: the fixed v4.0.6/v4.0.7 browser-helper bundles contain the guard, and a current launcher advertises its current helper even to an older daemon. I am closing this PR because it changes a correct upgrade path based on an unverified root cause and weakens the durable-runtime invariant. No revision is requested for this approach.

PR #255, opened 2026-08-31
fix: avoid send activation timeout

qeeqez:
## Summary

- stop Playwright from waiting on post-send navigation after ChatGPT activates Send
- keep send activation bounded by the bridge stage abort signal
- continue proving submission through existing turn/MCP evidence
- add a regression test that simulates the composer detaching immediately after Send and reproduces the old 30s locator.press timeout unless the fixed options are used

## Verification

- bun test tests/browser-worker-contract.test.ts (105 pass)
- bun run typecheck
- git diff --check

### DIFF
--- a/src/adapters/chatgpt-web/browser-worker.ts
+++ b/src/adapters/chatgpt-web/browser-worker.ts
@@ -2571,7 +2571,11 @@ export class ChatGptBrowserWorker {
     await captureDiagnostic?.("send-ready");
     const initialToolBatchRevision = externalProgress?.snapshot().lastToolBatchRevision ?? 0;
     await submissionLifecycle?.onSendActivated?.();
-    await sendButton.press("Enter");
+    // ChatGPT may replace the composer or schedule a history/navigation update as soon as Enter
+    // activates Send. Waiting for Playwright's post-action navigation can therefore time out even
+    // though the message was already submitted. Submission is proven below from turn/MCP evidence,
+    // so keep activation bounded by the browser stage instead of Playwright's default action timeout.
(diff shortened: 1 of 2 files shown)


qeeqez:
Original issue received:
```
stream disconnected before completion: ChatGPT Send was activated, but acceptance could not be proven; the prompt will not be resent: locator.press: Timeout 30000ms exceeded.
Call log:
[2m  - waiting for locator('[data-testid="prompt-textarea"], #prompt-textarea, [contenteditable="true"][data-lexical-editor="true"]').filter({ visible: true }).first().locator('xpath=ancestor::form[1]').getByTestId('send-button')[22m
[2m    - locator resolved to <button aria-disabled="false" interestfor="_r_1a9_" aria-label="Send prompt" data-testid="send-button" aria-describedby="_r_1a9_" id="composer-submit-button" class="composer-submit-btn composer-submit-button-color h-9 can-hover:relative can-hover:after:absolute can-hover:after:-inset-x-1 can-hover:after:inset-y-0 can-hover:after:content-[''] w-9">…</button>[22m
[2m  - elementHandle.press("Enter")[22m
```

miuuyy:
Thanks — the reported root cause is valid, and we implemented the fix locally rather than merging this version as-is. The issue here is `timeout: 0`: it disables the Playwright action timeout completely, and the added test also makes that unbounded behavior part of the contract.

Our patch keeps `noWaitAfter: true` and the abort signal, but uses the existing bounded `browserStageTimeouts.send` budget (20 seconds). This avoids the false post-send navigation timeout without allowing the Send action itself to hang indefinitely. The local version passes all 104 browser-worker contract tests and typecheck on Bun 1.4.0.

PR #256, opened 2026-08-31
fix: recover destroyed retained browser tabs

qeeqez:
## Summary

- remove retained ChatGPT tab ownership when its Electron webContents is destroyed
- defensively discard stale destroyed retained tabs before a later Codex provider round
- let ordinary continued tasks allocate a fresh ChatGPT surface and use their full canonical prompt instead of reusing a dead surface id
- preserve fail-closed behavior for flows that explicitly require the retained conversation

## Regression coverage

- destroyed retained surface is discarded and a later provider round receives a fresh surface
- external webContents destruction immediately removes the retained ownership record

## Verification

- launcher tests: 227 passed, 0 failed, 1 skipped
- launcher TypeScript typecheck passed
- git diff --check passed

### DIFF
--- a/launcher/electron/browser-host.cjs
+++ b/launcher/electron/browser-host.cjs
@@ -515,6 +515,15 @@ class BrowserHost {
 
   bindTurnContents(tab) {
     const contents = tab.view.webContents;
+    contents.once("destroyed", () => {
+      if (this.turnTabs.get(tab.id) !== tab) return;
+      this.logger.warn("browser.tab_destroyed", {
+        tabId: tab.id,
+        traceId: tab.traceId,
+        status: tab.status,
+      });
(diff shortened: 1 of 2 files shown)


qeeqez:
I have noticed that sometimes continue does not work on a session, which browser window has been closed long time ago. 
The fix - is what codex found

miuuyy:
Thanks. The reported Continue failure has no trace or reproduction, and this patch is not scoped to the Codex Continue action; it affects every later provider round. The new destroyed listener can also run during BrowserHost shutdown and recreate the descriptor after teardown removed it, leaving stale state. Closing rather than requesting a revision; please open an issue with a Safe Log if the failure repeats.

Issue #205, opened 2026-08-29
Feature: first-class external router/provider mode that never replaces Codex routing


ventianima-lab:
## Problem

Codex Web GPT currently behaves primarily as the Codex route owner. Users who already use another OpenAI-compatible model router need CGW to run only as one provider behind that router.

This is not only a setup-time conflict. On Windows with Codex Web GPT 4.0.5, I reproduced the launcher changing the active Codex route again during startup even though both persisted configurations said external-provider:

- launcher state: integrationMode = external-provider
- runtime config: codexIntegrationMode = external-provider
- external router endpoint before CGW startup: http://127.0.0.1:10100/v1
- CGW provider endpoint: http://127.0.0.1:17841/v1
- launcher startup log then ran bridge-connect / route connect
- Codex openai_base_url was changed from the external router to http://127.0.0.1:17841/v1

The launcher state file did not contain bridgeEnabled, so state loading defaulted it to true. Disconnecting the direct route and persisting bridgeEnabled = false restored the intended topology, but this should be a first-class supported mode rather than a local state repair.

## Requested feature

Please add an explicit router/integration mode with two clear choices:

1. Direct route: CGW owns the Codex route, as it does today.
2. External router/provider: CGW runs its browser host, proxy, and tunnel, but is only an OpenAI-compatible provider behind another router.

In External router/provider mode, CGW should:

- never call route connect or replace Codex model_provider/openai_base_url;
- preserve that rule across launcher startup, autostart, updates, repair, setup, and doctor;
- expose the CGW provider endpoint and model metadata needed by a generic external router;
- show in the launcher UI that routing is managed externally;
- fail closed if it cannot verify the external-router arrangement, rather than silently taking ownership of the Codex route;
- migrate older launcher state so external-provider implies bridgeEnabled = false even when that field is absent.

This should be router-agnostic. OpenCodex is one example, but the contract should work with any local OpenAI-compatible router that can forward the cgw models to the CGW endpoint.

## Regression test suggestion

Use a persisted launcher state fixture where integrationMode is external-provider and bridgeEnabled is absent, plus a runtime config where codexIntegrationMode is external-provider. Starting the launcher must leave route status inactive and must not change the existing Codex base URL.

## Why this matters

Without this mode, starting CGW can break native or third-party model routes even though users only want to add ChatGPT Web models. This is adjacent to the native-model reconnect reports in #146, but the requested fix is a distinct routing-ownership contract.

[...]

PR #281, opened 2026-09-01
feat: opt-in coexistence routing behind an existing Codex router

thirdbase1:
## Summary

Setup currently refuses to run whenever `openai_base_url` is already configured, naming the other wrapper and stopping there. That is the right default, but it means users of local model routers (codex-router, OpenCodex, Headroom, …) had to disconnect their router entirely to use the bridge — even though the two compose naturally: ChatGPT Web turns go to the browser, everything else forwards to the router.

This PR adds an opt-in chaining mode. When setup replaces a non-empty `openai_base_url`, the previous value is stored privately as `nativePassthroughBaseUrl` and used as the upstream for every non-ChatGPT-Web request:

- `POST /v1/responses`
- `POST /v1/responses/compact`
- `POST /v1/alpha/search`

Native model discovery (`GET /v1/models`) stays on the official backend, and ChatGPT Web turns are byte-for-byte unchanged. With no configured value the official backend is used, exactly as before — the field is optional and nothing changes for existing installs.

## Changes

- `src/config.ts` — optional `nativePassthroughBaseUrl`; `normalizeNativePassthroughBaseUrl()` accepts only HTTP loopback URLs with no credentials, query, or fragment, so the existing same-OS-user trust boundary is preserved rather than widened. Validated in `parseConfig`.
- `src/codex-integration-document.ts` / `codex-integration.ts` — `readCodexOpenAiBaseUrl()` reads the previous route for capture.
- `src/setup.ts` — setup captures the pre-existing `openai_base_url` (when it differs from the bridge's own route) into the config field; the value participates in `meaningfulRuntimeChange` so daemon restarts track it.
- `src/native-passthrough.ts` — `forwardNativeCodexRequest()` accepts an optional upstream base (defaults to the official backend).
- `src/server.ts` — non-Web responses, compaction, and search forward to the configured value; `/v1/models` is untouched.
- tests — loopback/credential validation, exact Router URL construction for all three forwarded endpoints, and route capture from a fixture Codex config.

## Notes for review

- **Security:** the stored URL must be loopback and credential-free. No secrets enter command lines or logs; the value is a route, not a credential, and follows the same user-only storage as the rest of the app config.
- **Restorability:** unchanged — the integration journal still owns only `openai_base_url` and restores the prior value byte-for-byte on disconnect/uninstall; `model_provider` / `model_catalog_json` continue to be preserved, never removed.
- **Verified against a real router:** with codex-router on `127.0.0.1:4202`, `codex-chatgpt-web setup --browser-only --replace-codex-route …` captured the router URL, and a packaged build served a ChatGPT Web Luna turn (smoke test: all 11 checkpoints, `turn completed`) plus a routed-provider turn that reached the router and completed `200`.

## Testing

- [x] `bun test tests/native-passthrough.test.ts tests/runtime-layout.test.ts tests/codex-integration.test.ts tests/server-models.test.ts tests/server-compaction.test.ts` — 71 pass, 0 fail
- [x] `tsc --noEmit` clean
- [x] End-to-end on Windows: browser-only setup against a live codex-router; Web turn and router-chained turn both completed

### DIFF
--- a/src/codex-integration-document.ts
+++ b/src/codex-integration-document.ts
@@ -108,6 +108,12 @@ export function readCodexModelContextOverride(): CodexModelContextOverride | und
   return contextWindow === undefined ? undefined : { contextWindow };
 }
 
+export function readCodexOpenAiBaseUrl(): string | undefined {
+  const path = getCodexConfigPath();
+  if (!existsSync(path)) return undefined;
+  return findTopLevelAssignment(splitLines(readFileSync(path, "utf8")), "openai_base_url").value;
+}
+
 export function assignments(lines: string[]): Record<ManagedAssignmentKey, PreviousAssignment> {
   return {
     openai_base_url: findTopLevelAssignment(lines, "openai_base_url"),
--- a/src/codex-integration.ts
+++ b/src/codex-integration.ts
@@ -80,7 +80,7 @@ export {
   getCodexJournalRecoveryPath,
   getCodexModelsCachePath,
 } from "./codex-integration-shared";
-export { readCodexModelContextOverride } from "./codex-integration-document";
+export { readCodexModelContextOverride, readCodexOpenAiBaseUrl } from "./codex-integration-document";
 export type {
   CodexIntegrationJournal,
   CodexModelContextOverride,
--- a/src/config.ts
+++ b/src/config.ts
@@ -82,6 +82,8 @@ export interface AppConfig {
   experimentalBiggerContext: boolean;
   /** Optional adapter-silence budget for the Responses watchdog. */
   stallTimeoutSec?: number;
+  /** Existing loopback Responses route used for every non-Web model. */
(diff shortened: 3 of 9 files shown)


miuuyy:
Thanks for the implementation effort and the live router validation. I am closing this because it does not implement #205's ownership contract.

The requested mode must leave the external router as Codex's sole entrypoint and must never replace or restore `openai_base_url` during setup, startup, repair, doctor, or upgrade. This PR instead captures the old URL, installs CGW as the front route, and forwards non-Web requests back to the old router. That inverse topology still makes CGW the Codex route owner and does not solve the reported external-provider lifecycle.

The forwarding tests are useful for the topology implemented here, but they cannot prove the required no-route-mutation invariant. We will implement the owner-correct external-provider mode separately rather than iterate this PR into a different architecture.

PR #291, opened 2026-09-02
feat: add Dock and taskbar visibility settings

### LINKED ISSUE #286: [Feature] Hide ChatGPT Web helper icon from macOS Dock
### User problem

On macOS, running codex-chatgpt-web shows a second ChatGPT-style icon in the Dock alongside the official ChatGPT app. Since the two icons look nearly identical, it is difficult to tell which one is ChatGPT and which one is the ChatGPT Web connection.

Image

### Current workflow or workaround

I have to remember each app’s Dock position or hover over the icons to identify them. This is easy to mix up when both are running.

### Proposed behavior

Add a macOS setting to hide the ChatGPT Web app from the Dock while keeping the connection running and accessible from the menu bar.

This would let users keep ChatGPT Web active without a second ChatGPT-like Dock icon. The menu-bar icon should remain available for status and controls. If fully hiding the Dock icon is incompatible with the app lifecycle, using a clearly distinct helper icon would be a reasonable fallback.

Before I start implementation, would maintainers be open to a small PR? I’m happy to submit one once the preferred approach is confirmed.

### Primary surface

Packaging or installer / macOS launcher behavior.

### Safety and compatibility boundaries

This should be macOS-specific and must not affect authentication, browser automation, model routing, or connection behavior. It only changes how the running helper appears in the Dock and menu bar.

### Scope checklist

- [x] I searched existing issues and Discussions for this request.
- [x] This request does not rely on bypassing authentication, access controls, or account limits.
- [x] I understand that opening an issue does not guarantee implementation or acceptance of a pull request.

wpfyorg:
## Summary

- add a macOS setting to hide Codex Web GPT from the Dock
- add Windows and Linux settings to hide the launcher from the taskbar
- persist and apply platform-specific icon visibility immediately and on startup
- require an available menu bar/system tray entry before runtime hiding, and restore the app icon if tray creation fails
- add English, Chinese, and Japanese copy plus regression coverage

## Verification

- `npx -y bun@1.4.0 run verify`
- 522 core tests passed
- 250 launcher tests passed with 1 expected platform-specific skip
- launcher typecheck and production renderer build passed
- dependency audits reported no vulnerabilities
- relocatable runtime smoke passed
- Windows/Linux behavior is covered by automated wiring/state tests but was not manually verified on those operating systems

Fixes #286


miuuyy:
Thanks for the contribution. I’m closing this because it conflicts with the current v5 launcher work and is too risky to merge during the current release cycle.

Issue #248, opened 2026-08-31
[Bug] ChatGPT Web turns fail with missing cwd context and multipart send timeout


1saifj:
## Summary

Two additional ChatGPT Web failure modes are occurring during Codex turns:

1. A turn fails immediately because the ChatGPT Web request is missing `cwd` in the trusted Codex environment context.
2. Another turn repeatedly reconnects and then fails because the ChatGPT browser stage times out at `multipart_stage_2_send`.

These appear related to Web-turn context/transport reliability but are distinct from the partial-edit recovery problem reported in #247.

### Missing trusted `cwd`

After the model changes from **GPT-5.6 Sol** to **ChatGPT Web — Pro**, the turn reaches repeated reconnect attempts and ends with:

> stream disconnected before completion: ChatGPT web turn is missing cwd in trusted Codex environment context

The composer remains available, but the turn does not proceed.

### Multipart browser-stage timeout

A separate turn shows reconnect attempts and ends with:

> stream disconnected before completion: ChatGPT browser stage timed out: multipart_stage_2_send

The composer still appears to be in an active/thinking state, including a visible stop button, which makes it unclear whether the browser turn is still running or has already failed.

## Expected behavior

- Every ChatGPT Web turn should receive a complete trusted Codex environment context, including `cwd`, before the browser request is accepted.
- If required context is missing, the bridge should fail before starting the Web turn with a clear actionable diagnostic.
- Multipart-stage timeouts should terminate cleanly and expose a stable failed state instead of leaving the composer looking active.
- Reconnect attempts should not leave the user uncertain whether pressing Stop, retrying, or starting a new turn is safe.

## Steps to reproduce

1. Select a ChatGPT Web model in Codex, including ChatGPT Web — Pro.
2. Submit a prompt from a repository-backed Codex task.
3. Observe repeated reconnect attempts.
4. Reproduce either failure mode:
   - missing `cwd` in the trusted Codex environment context; or
   - timeout at `multipart_stage_2_send`.
5. Check whether the turn is cancelled, whether the composer remains active, and whether retrying changes the outcome.

## Impact

These failures prevent normal ChatGPT Web turns from completing and provide insufficient lifecycle state during recovery. The missing-`cwd` error may indicate that the request cannot be safely associated with the current trusted workspace, while the multipart timeout can leave the user unsure whether the prompt was sent or whether a retry could duplicate work.

## Evidence

Two screenshots show the exact error messages and reconnect states:

[...]

PR #296, opened 2026-09-02
fix: recover cwd from first workspace root when resumed turn omits it (fixes #288)

### LINKED ISSUE #288: [Bug] Codex 0.150.0-alpha.8 omits <cwd> from <environment_context> when resuming a thread after a model switch ("missing cwd in trusted Codex environment context")
## Summary

On macOS, turns sent by the ChatGPT desktop app's bundled Codex CLI (`0.150.0-alpha.8`) fail with

> stream disconnected before completion: ChatGPT web turn is missing cwd in trusted Codex environment context

whenever the turn **resumes an existing thread after the user switched the model to a ChatGPT Web model**. Fresh sessions and terminal `codex exec` turns from the same alpha binary work fine. This looks like the same signature as #248 ("After the model changes from GPT-5.6 Sol to ChatGPT Web — Pro"), with a now-verified client-side root cause.

## Environment

- codex-chatgpt-web 4.0.7 (I also checked the v4.0.8 `app/cli.js`: the `eu()`-equivalent environment parser is unchanged, so 4.0.8 is affected too)
- Codex CLI `0.150.0-alpha.8` (bundled with ChatGPT.app 151.0.7922.170, macOS 26.2 arm64)
- Terminal Codex CLI `0.144.1` is NOT affected

## Root cause (verified from rollout records)

In the failing turn, the `<environment_context>` block contains **no `<cwd>` element at all**:

```xml
<environment_context>
  <current_date>2026-09-02</current_date>
  <timezone>Asia/Shanghai</timezone>
  <filesystem><workspace_roots><root>/Users</root><root>/Users/evander/.codex/visualizations/2026/08/31/<thread-id></root></workspace_roots><permission_profile type="disabled"><file_system type="unrestricted" /></permission_profile></filesystem>
</environment_context>
```

Working turns (fresh `exec` sessions from the same `0.150.0-alpha.8` binary, and 0.144.1 turns) do include `<cwd>…</cwd>` as an additional `input_text` part of the same user message. The bridge fail-closes (correctly, per the design note in #248) because it must not invent a cwd — but the client gave it nothing to recover.

## Suggested fix

When `<cwd>` candidates are empty, recover cwd from the **first `<workspace_roots><root>` entry** (Codex's convention: first root = primary workspace), then continue the existing validation chain unchanged:

- the absolute-path check still applies to the recovered value;
- the "cwd is outside the trusted Codex workspace roots" containment check passes trivially since the value comes from `workspace_roots`;
- the sandbox parser already recognizes this shape (`permission_profile type="disabled"` + `file_system type="unrestricted"` → danger-full-access), so no further changes are needed;
- if no `<workspace_roots>` roots exist either, keep throwing `MissingTrustedCodexEnvironmentError` (fail-closed preserved).

I applied exactly this as a local patch (in `eu()`: when the cwd candidate list is empty, `n = [firstRoot]` before the `Jl(n,"cwd")` dedup) and the previously-failing turns now pass both a byte-level replay of the failing payload and normal end-to-end turns.

## Related

#248 (same error signature after model switch), #59, #153, #177, #178.

samuelfaj:
Fixes #288

## Summary

Resumed turns from Codex CLI `0.150.0-alpha.8` (bundled in ChatGPT.app `151.0.7922.170` / macOS 26.2 arm64) omit `<cwd>` from `<environment_context>` whenever the thread is resumed after the user switches the model to a ChatGPT Web model. The bridge correctly fail-closes with:

> `ChatGPT web turn is missing cwd in trusted Codex environment context`

but the turn then stays stuck in `Reconnecting… / waiting for network` and never completes. Fresh `codex exec` sessions from the same binary and `0.144.1` turns are unaffected — see the verified rollout payload in #288.

This implements the suggested fix verbatim: when cwd candidates are empty, recover cwd from the **first `<workspace_roots><root>` entry** (Codex's convention: first root = primary workspace) and keep the entire existing validation chain.

## Root cause

In the failing turn the `<environment_context>` contains no `<cwd>` at all:

```xml
<environment_context>
  <filesystem><workspace_roots><root>/Users/…</root><root>/Users/…/.codex/visualizations/…/<thread-id></root></workspace_roots><permission_profile type="disabled"><file_system type="unrestricted" /></permission_profile></filesystem>
</environment_context>
```

Working turns do include `<cwd>…</cwd>` as an additional `input_text` part of the same user message. The previous parser threw before it could derive the trusted envelope.

## Change

`src/adapters/chatgpt-web/environment.ts` — `extractChatGptTurnEnvironment`:

```ts
const rootMatches = [...text.matchAll(/<workspace_roots>[\s\S]*?<\/workspace_roots>/g)]
  .flatMap(section => [...section[0].matchAll(/<root>([^<]+)<\/root>/g)].map(match => match[1] ?? ""));
const cwdMatches = environmentCwdMatches(text, clientMetadataWorkspaceRoots(parsed));
const cwdCandidates = uniqueAbsolutePaths(
  cwdMatches.length > 0 ? cwdMatches : rootMatches.slice(0, 1),
  "cwd",
);
```

* `roots` is now parsed before `cwdCandidates` so the fallback source exists.
* If `cwdMatches` is non-empty the path is unchanged; if empty we pass `rootMatches.slice(0, 1)` — exactly one candidate — to the same `uniqueAbsolutePaths(..., "cwd")` dedup.
* If neither `<cwd>` nor `<workspace_roots>` exist, the empty array still throws `MissingTrustedCodexEnvironmentError` — fail-closed is preserved.

Why this is safe:

* the absolute-path check still applies to the recovered value
* `cwd is outside the trusted Codex workspace roots` passes trivially since the value comes from `workspace_roots`
* `sandboxTypeFromEnvironment` already recognizes this shape (`permission_profile type="disabled"` / `file_system type="unrestricted"` → `dangerFullAccess`), so no sandbox change is needed

## Tests

`tests/environment.test.ts` — new case `uses the first workspace root when a resumed turn omits cwd` replays the #288 payload byte-for-byte (two roots, no `<cwd>`, `dangerFullAccess`) and asserts:

```ts
{ cwd: root, roots: [root, visualizationsRoot], writableRoots: [root, visualizationsRoot], sandboxPolicy: { type: "dangerFullAccess" } }
```

```
bun test ./tests/environment.test.ts — 29 pass 0 fail
bun test — 523 pass 0 fail
```

## How to verify

1. Rebuild the runtime (`bun run build` requires Bun `1.4.0`) — `dist/runtime/app/cli.js` should contain the fallback `r.length>0?r:n.slice(0,1)` path.
2. Promote the bundle to `~/.codex-chatgpt-web/versions/<version>-<platform>-<arch>/` (atomic copy via `launcher/electron/runtime-install.cjs` semantics) or reinstall via the launcher.
3. Replay the failing payload or do a real end-to-end resume-after-model-switch on macOS — previously failing turns now complete.

Closes #288. Related #248.

## Checklist

- [x] Rebuild verified with Bun 1.4.0 — `dist/runtime/manifest.json` `bundleId` matches file records
- [x] No new dependencies
- [x] Dist is gitignored; only `src/` + `tests/` are in the PR

### DIFF
--- a/src/adapters/chatgpt-web/environment.ts
+++ b/src/adapters/chatgpt-web/environment.ts
@@ -510,13 +510,16 @@ function matchesPath(root: string, path: string): boolean {
 
 export function extractChatGptTurnEnvironment(parsed: CodexParsedRequest): ChatGptTurnEnvironment {
   const text = trustedEnvironmentText(parsed);
+  const rootMatches = [...text.matchAll(/<workspace_roots>[\s\S]*?<\/workspace_roots>/g)]
+    .flatMap(section => [...section[0].matchAll(/<root>([^<]+)<\/root>/g)].map(match => match[1] ?? ""));
   const cwdMatches = environmentCwdMatches(text, clientMetadataWorkspaceRoots(parsed));
-  const cwdCandidates = uniqueAbsolutePaths(cwdMatches, "cwd");
+  const cwdCandidates = uniqueAbsolutePaths(
+    cwdMatches.length > 0 ? cwdMatches : rootMatches.slice(0, 1),
+    "cwd",
+  );
   if (cwdCandidates.length !== 1) throw new Error("ChatGPT web turn has conflicting trusted Codex cwd values");
   const cwd = cwdCandidates[0]!;
 
-  const rootMatches = [...text.matchAll(/<workspace_roots>[\s\S]*?<\/workspace_roots>/g)]
-    .flatMap(section => [...section[0].matchAll(/<root>([^<]+)<\/root>/g)].map(match => match[1] ?? ""));
   const roots = rootMatches.length > 0 ? uniqueAbsolutePaths(rootMatches, "workspace_roots") : [cwd];
   if (!roots.some(root => matchesPath(root, cwd))) {
     throw new Error("ChatGPT web cwd is outside the trusted Codex workspace roots");
--- a/tests/environment.test.ts
+++ b/tests/environment.test.ts
@@ -104,6 +104,28 @@ describe("trusted current Codex environment envelope", () => {
     });
   });
 
+  test("uses the first workspace root when a resumed turn omits cwd", () => {
+    const visualizationsRoot = resolve(root, "visualizations", "thread_current");
+    const resumedEnvironment = `<environment_context>
+  <current_date>2026-09-02</current_date>
+  <timezone>America/Asuncion</timezone>
+  <filesystem><workspace_roots><root>${root}</root><root>${visualizationsRoot}</root></workspace_roots>${dangerFullAccessProfileXml}</filesystem>
+</environment_context>`;
+    const request = currentWire({ environmentXml: resumedEnvironment });
+    const body = request._rawBody as { input: Array<Record<string, unknown>> };
+    for (const item of body.input) {
+      item.internal_chat_message_metadata_passthrough = { turn_id: "turn_current" };
+    }
+
+    expect(extractChatGptTurnEnvironment(request)).toEqual({
+      cwd: root,
+      roots: [root, visualizationsRoot],
+      writableRoots: [root, visualizationsRoot],
+      sandboxPolicy: { type: "dangerFullAccess" },
+      tools: [],
+    });
+  });
+
   test("accepts either canonical provenance form on an intervening developer message", () => {
     for (const developer of [
       {


miuuyy:
Thanks for the careful report and focused reproduction. We are not going to merge this revision.

The first-workspace-root fallback is safe only for the exact Codex 0.150 filesystem-only envelope. In this implementation it runs whenever parsed cwd candidates are empty, which also makes malformed or ambiguous input look recoverable: malformed `<cwd>` markup, malformed `<root>` markup, and multiple `<workspace_roots>` sections are not distinguished from a genuinely omitted cwd. The added coverage proves the positive two-root case but not those fail-closed boundaries.

Our local 5.0 work already contains a stricter implementation that we will commit shortly. It:

- keeps an explicit cwd authoritative;
- uses the first root only when there is exactly one well-formed `workspace_roots` section and no cwd tag at all;
- rejects malformed cwd/root markup and ambiguous envelopes;
- covers multi-folder and projectless turns.

I am closing this PR to avoid maintaining two competing implementations of the same fix. The issue report and reproduction were still useful—thank you.

PR #298, opened 2026-09-02
fix: support native v2 cross-backend subagents

### LINKED ISSUE #297: Native V2 cross-backend delegation arrives as unreadable encrypted payload
## Problem

With the **Native** subagent protocol selected, a native Codex model can expose MultiAgent V2 collaboration tools whose `message` arguments are marked `encrypted: true`. The official backend then returns provider-encrypted function arguments. A routed ChatGPT Web child cannot decrypt that cross-backend payload, so the turn fails before opening a browser with:

> ChatGPT Web cannot read this encrypted cross-backend subagent payload.

This blocks native-to-Web delegation and can also poison later continuations of the same V2 task. Restarting and starting a new task is necessary after changing the protocol, but it does not solve the transport mismatch while the forwarded tool schema still requests encryption.

## Reproduction

1. Install v4.0.8 in Full mode and select `codex-chatgpt-web subagents native`.
2. Fully restart Codex Web GPT and Codex, then start a new task.
3. From a native Codex model, call V2 `spawn_agent` with a `chatgpt-web/*` child model.
4. Observe the encrypted cross-backend payload error.

## Expected behavior

Native mode should support readable V2 delegation across native and ChatGPT Web backends while Compatibility V1 remains the conservative default.

## Proposed focused fix

For authenticated native Responses passthrough **only when Native mode is selected**:

- recognize V2 collaboration surfaces in both ordinary `tools` and Responses Lite `additional_tools`;
- remove `encrypted: true` only from the `message` schemas of `spawn_agent`, `send_message`, and `followup_task`;
- add Codex's explicit `encrypted_function_args: []` plaintext-delivery marker only to matching plaintext function calls in SSE or JSON responses;
- preserve genuinely encrypted calls, unrelated tool schemas, compaction, model discovery, and every other native request unchanged;
- document that cross-agent prompts are locally readable in Native mode and that an already-encrypted task cannot be recovered.

Regression coverage should include namespaced, configured-namespace, unnamespaced, Responses Lite, SSE, JSON, encrypted-preservation, and Compatibility V1 cases.

Irvinwop:
## What this changes

Fixes #297.

Native mode now rewrites only the MultiAgent V2 collaboration message boundary that must cross providers:

- authenticated native Responses requests remove `encrypted: true` only from the `message` schemas of `spawn_agent`, `send_message`, and `followup_task`;
- ordinary `tools`, Responses Lite `additional_tools`, configured collaboration namespaces, and unnamespaced surfaces are covered;
- matching plaintext function calls receive `encrypted_function_args: []` in SSE and JSON responses so Codex delivers readable arguments;
- genuinely encrypted calls, unrelated schemas, model discovery, compaction, and Compatibility V1 remain unchanged;
- documentation now explains local plaintext history exposure and the required launcher/Codex restart plus new-task boundary.

## Evidence

Before this change, a native V2 spawn targeting a `chatgpt-web/*` child forwards an encrypted message schema. The official backend returns provider-private function arguments and the routed Web child fails before browser construction with the encrypted cross-backend payload error.

Focused request/response fixtures now prove schema rewriting and plaintext markers for namespace, configured-namespace, unnamespaced, Responses Lite, SSE, and JSON shapes. Negative cases prove that encrypted arguments, unrelated tool fields, and Compatibility V1 remain untouched.

`bun run smoke:subagents` exercises a stock Codex binary through nested root/child/grandchild spawn, wait, and follow-up lifecycles:

```text
CODEX_SUBAGENT_V1_LIFECYCLE_SMOKE_OK
CODEX_SUBAGENT_V2_LIFECYCLE_SMOKE_OK
```

## Scope and invariants

## Verification

- [x] I ran `bun install --frozen-lockfile` in the repository root and `launcher/` (Bun 1.4.0; no changes).
  - The gate currently stops at unchanged dependency audits: root reports 4 high and 2 moderate advisories in `fast-uri`/`qs`; launcher reports 4 high and 1 moderate in `fast-uri`/`@xmldom/xmldom`. No package or lockfile is changed by this PR.
  - Running the remaining stages directly produced 530/530 core tests, 248/248 runnable launcher tests (1 platform skip), clean root/launcher typechecks, a successful launcher production build, runtime bundle, and third-party notices.
  - The final relocated-runtime smoke reached its macOS browser check and failed because `/Applications/Google Chrome.app/Contents/MacOS/Google Chrome` is absent on this host.
- [ ] I manually tested the affected behavior.
  - The stock-Codex V1/V2 lifecycle smoke is green; a fresh account-bound browser task after a full launcher/Codex restart remains to be run.
- [ ] If this changes local tools, MCP execution, or the outer Codex agent loop, I tested it through a real installed Codex integration; DEV mode alone is acceptable only when execution is not affected.
  - A signed macOS arm64 package was installed and its authenticated launcher browser/protocol checks passed, but the post-restart cross-backend task is not yet claimed.
- [x] If this changes ChatGPT browser UI handling, I included observed DOM evidence and a reproducible fixture instead of broadening selectors speculatively. (Not applicable: no browser selectors or UI handling changed.)
- [x] If this changes the launcher, I preserved macOS, Windows, and Linux packaging and named the platform packages actually built below. (Not applicable: no launcher source changed.)

- macOS arm64, Full mode: packaged application installed; Native protocol is active; authenticated embedded-browser check passed.
- macOS arm64: stock Codex V1 and V2 nested lifecycle smoke passed.
- Account-bound browser-backed V2 delegation after full restart: not run yet.
- Windows and Linux packages: not run; no launcher code changed.

### DIFF
--- a/README.ja.md
+++ b/README.ja.md
@@ -160,9 +160,12 @@ ChatGPT のツール呼び出しを現在の Codex タスクへ接続します�
 アンインストール前の Codex 統合削除も行えます。すべてのブラウザーチェックポイントでスクリーンショットが必要な場合にのみ、
 `CODEX_CHATGPT_WEB_BROWSER_DIAGNOSTICS=1` を設定してください。
 
-新規インストールでは、クロスバックエンドのサブエージェントに **Compatibility V1** を使用します。
-**Native** は Codex 独自の機能設定を維持し、プレーンテキストの Web-to-Web V2 delegation を有効にします。
-プロトコル変更後は Codex を再起動し、新しいタスクを開始してください。
+新規インストールでは、保守的なクロスバックエンド方式として **Compatibility V1** を使用します。
+**Native** は Codex 独自の機能設定を維持し、ネイティブモデルと ChatGPT Web モデル間の V2 delegation を有効にします。
+Native モードでは、ローカルの認証済みプロキシが `spawn_agent`、`send_message`、`followup_task` の
+メッセージ引数だけをプレーンテキストにするため、どちらのバックエンドでも読み取れます。そのため、これらの委任プロンプトは
+プロバイダー専用の暗号化データではなく、ローカルで読み取り可能になります。プロトコル変更後は Codex Web GPT と Codex を再起動し、
+新しいタスクを開始してください。
 
 ```bash
 codex-chatgpt-web subagents status
--- a/README.md
+++ b/README.md
@@ -163,9 +163,12 @@ Use **Activity** for safe local diagnostics and **Settings → Run doctor** for 
 Settings can also cancel a retained browser turn or remove the Codex integration before uninstall.
 Set `CODEX_CHATGPT_WEB_BROWSER_DIAGNOSTICS=1` only when every browser checkpoint needs a screenshot.
 
-New installs use **Compatibility V1** for cross-backend subagents. **Native** preserves Codex's own
-feature settings and enables plaintext Web-to-Web V2 delegation. Restart Codex and start a new task
-after changing the protocol:
+New installs use **Compatibility V1** for the conservative cross-backend path. **Native** preserves
+Codex's own feature settings and enables V2 delegation between native and ChatGPT Web models. In
+Native mode, the authenticated local proxy makes only `spawn_agent`, `send_message`, and
+`followup_task` message arguments plaintext so either backend can read them; those delegation
+prompts are therefore locally readable instead of provider-opaque. Restart Codex Web GPT and Codex,
+then start a new task after changing the protocol:
 
 ```bash
 codex-chatgpt-web subagents status
--- a/README.zh-CN.md
+++ b/README.zh-CN.md
@@ -152,8 +152,11 @@ bun run app
 取消保留的浏览器任务，或在卸载前移除 Codex 集成。仅在需要为每个浏览器检查点保存截图时设置
 `CODEX_CHATGPT_WEB_BROWSER_DIAGNOSTICS=1`。
 
-新安装默认使用 **Compatibility V1** 以支持跨后端 subagent。**Native** 会保留 Codex 自身的
-功能设置，并启用明文 Web-to-Web V2 委派。切换协议后，请重启 Codex 并创建新任务：
+新安装默认使用较保守的 **Compatibility V1** 跨后端方案。**Native** 会保留 Codex 自身的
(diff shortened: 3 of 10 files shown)


miuuyy:
Thanks for investigating the Native V2 failure and documenting the plaintext tradeoff. We are declining this implementation because it changes a security-sensitive native transport boundary without enough authoritative proof.

The main blockers are:

- V2 collaboration is inferred heuristically from a namespace name or the co-occurrence of three tool names, rather than from an explicit upstream capability/ABI.
- The response transform recursively walks arbitrary response objects and can mutate any matching `function_call` shape, instead of operating on a closed set of known response events.
- Removing argument encryption and adding plaintext markers changes authenticated native passthrough semantics and local confidentiality.
- The fixtures prove the proposed rewrite itself, but the affected account-bound native-to-Web delegation was not tested end to end.

That combination is too risky for the core passthrough path: an unrelated future tool shape could be silently reclassified and rewritten. Compatibility V1 remains the conservative path while Native V2 cross-backend transport lacks an explicit, verified contract.

Closing this PR rather than shipping a heuristic protocol rewrite. Thank you for doing the investigation and for being clear about what was not live-tested.

Irvinwop:
Rebased the branch onto upstream v5.0.0 and force-updated the existing head branch to [`feebe02`](https://github.com/Irvinwop/codex-chatgpt-web/commit/feebe025043ce63fd98030496623fe10c6768c28). GitHub currently leaves this closed PR's displayed head pinned to its closure-time commit, so the linked fork commit/branch is the current code.

The v5 rebase also exposed an interaction with the older local cwd fallback: v5 now recovers Codex 0.150+ filesystem-only environment diffs from the first ordered workspace root, but the older fallback could bypass v5's malformed-`cwd` / malformed-`root` safeguards. The rebased branch removes that redundant fallback and adds project-plus-Codex-visualization-root regression coverage for the reported `ChatGPT web turn is missing cwd in trusted Codex environment context` failure.

Local verification on macOS arm64 with the pinned Bun 1.4.0 toolchain is green:

- full `scripts/verify.ts` release gate
- 603 runtime tests passing
- 278 launcher tests passing (1 platform skip)
- relocatable runtime smoke passing
- packaged macOS arm64 launcher smoke passing
- ZIP/DMG produced from source and the staged app passes strict deep code-signature validation

The Native V2 compatibility patches remain on the branch. I have not performed the account-bound end-to-end V2 flow after this rebase because the running app could not be restarted during the work, so this update does not claim to resolve the maintainer's stated security/contract objections or ask that the closed PR be reopened as-is.

PR #301, opened 2026-09-03
fix: resolve slider-overflow tiers through their model radio row

### LINKED ISSUE #300: [Bug] Pro tier: "ChatGPT effort slider does not expose item index 4 (min=0; max=3)" after chatgpt.com picker redesign
### Codex Web GPT version

v4.0.8

### Codex version

0.144.1 (terminal CLI; also reproduces from the desktop integration)

### Platform

macOS arm64

### OS version

macOS 26.2 (Darwin 25.2.0)

### ChatGPT account tier

Pro

### Integration mode

Full

### Selected model

chatgpt-web/pro

### What happened?

ChatGPT redesigned the composer effort picker. The reasoning-effort slider now exposes only four positions (`aria-valuemin=0`, `aria-valuemax=3`), and tiers beyond the slider range are rendered as separate model radio rows inside the same effort menu. Observed DOM (zh-CN UI, temporary chat, Pro account):

```html
<div tabindex="0" data-disabled="" aria-label="Pro" role="menuitemradio" aria-disabled="true" aria-checked="false" data-state="unchecked" data-orientation="vertical" data-radix-collection-item="" class="group __menu-item cursor-not-allowed!">…</div>
```

The menu contains the slider plus the model rows `GPT-5.6 Sol` (checked), `GPT-5.5`, `Pro`. The helper still maps the `chatgpt-web/pro` tier (adapter effort `max`) to slider index 4, so `targetValue = min + 4 = 4 > max = 3` and every Pro turn fails before the prompt is attached.

Notes from the live session:

- Pressing Enter on the `Pro` row does not activate it (the row stays `aria-checked="false"`); the rows only respond to a real click.
- The `Pro` row can be `aria-disabled="true"` / `data-disabled` (`cursor-not-allowed`). I observed it disabled in both Temporary Chat and a normal chat root, so it appears to be account-side gating (likely Pro quota), not a Temporary Chat restriction. When it is disabled the helper must fail closed instead of clicking.

### Minimal reproduction

1. Configure `chatgpt-web/pro` as the model.
2. Start any Codex turn on chatgpt.com current UI.
3. The browser stage fails at effort selection.

### Complete final error

```text
stream disconnected before completion: ChatGPT effort slider does not expose item index 4 (min=0; max=3)
```

### Privacy-safe diagnostics

Trace checkpoints for a failing turn: `effort-control-ready` → `effort-menu-open-requested` → `effort-slider-visible` → `turn-failed` with the error above; `effortItems` snapshot lists the three `menuitemradio` rows (GPT-5.6 Sol checked, GPT-5.5, Pro) and no fifth slider position. No raw browser state or credentials included.

### Other Codex wrappers or route owners

None

### Preflight checklist

- [x] I tested the latest release and, for catalog or turn failures, fully restarted Codex including its background process.
- [x] I stated whether the failure also occurs in a fresh Codex task, or explained why that does not apply.
- [x] I attached only a privacy-safe export, or explained why the exporter could not produce one; I removed any remaining private content.
- [x] I searched existing issues for this exact final error.

littlerodents:
## What this changes

When the requested effort tier's slider position exceeds the slider's `aria-valuemax`, `selectModelAndEffort` now resolves the tier through its model radio row in the same effort menu instead of failing outright: it waits briefly for a row matching `[role="menuitemradio"][aria-label="<displayLabel>"]`, fails closed (non-retryable, explicit error) when the row is `aria-disabled`/`data-disabled`, otherwise clicks it and confirms `aria-checked` (treating the menu closing as a committed switch). Tiers within the slider range behave exactly as before. Fixes #300.

## Evidence

Reproduction (v4.0.8, chatgpt.com current UI, Pro account, model `chatgpt-web/pro`): every turn fails at effort selection with

```text
ChatGPT effort slider does not expose item index 4 (min=0; max=3)
```

Observed DOM (captured from the live page, zh-CN Temporary Chat; also reproduced on the normal chat root):

```html
<div tabindex="0" data-disabled="" aria-label="Pro" role="menuitemradio" aria-disabled="true" aria-checked="false" data-state="unchecked" data-orientation="vertical" data-radix-collection-item="" class="group __menu-item cursor-not-allowed!">…</div>
```

The menu exposes a 4-position effort slider (`aria-valuemin=0`, `aria-valuemax=3`) plus the model rows `GPT-5.6 Sol` / `GPT-5.5` / `Pro`; diagnostic checkpoints show `effort-slider-visible` → `turn-failed` (issue #300). Enter/Space does not activate these rows — observed live: pressing Enter left the row `aria-checked="false"` — so the fallback uses a real click, then confirms `aria-checked` or the menu closing before returning.

With the equivalent change running in my installed helper: a slider-path turn (`chatgpt-web/extra-high`) completed end to end (`effort-selected` → `turn-completed`), and a `chatgpt-web/pro` turn while the row is `aria-disabled` failed fast with the new explicit error (checkpoint `pro-choice-disabled` → non-retryable adapter error). The enabled-row click path is pinned by the new contract tests; I could not observe it live because the account currently gates the row.

## Scope and invariants

## Verification

- [x] I ran `bun run verify` with the Bun version pinned by `package.json`.

Note on `bun run verify`: every step passes except the `audit` step, which fails on this commit for me exactly as it does on `main` — 6 pre-existing advisories in the committed lockfile (fast-uri via `ajv`, qs via `@modelcontextprotocol/sdk > express > body-parser`), none introduced by this branch (no dependency or lockfile changes). `typecheck`, the full `test` suite (525 → 528 tests, 0 failures), `launcher:typecheck`, `launcher:test`, `launcher:build`, the runtime bundle build, and `smoke-release` all pass on this branch.

The launcher is not touched by this change.

### DIFF
--- a/src/adapters/chatgpt-web/browser-worker.ts
+++ b/src/adapters/chatgpt-web/browser-worker.ts
@@ -133,6 +133,8 @@ const CHATGPT_SMOKE_EXPECTED = "CODEX WEB GPT READY";
  */
 export const CHATGPT_UI_SETTLE_MS = 250;
 export const CHATGPT_SEND_ENABLE_GRACE_MS = 5_000;
+const CHATGPT_EFFORT_TIER_ROW_TIMEOUT_MS = 5_000;
+const CHATGPT_EFFORT_TIER_ACTIVATION_TIMEOUT_MS = 10_000;
 
 const CHATGPT_DOM_REVISION_ATTRIBUTES = [
   "aria-hidden",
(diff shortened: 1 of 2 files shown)


miuuyy:
Thanks for trying to help. I’m closing this because it does not fix the root cause, and we already have our own local changes in progress for the complete fix.

PR #303, opened 2026-09-03
fix: extend Pro compaction handoff timeout

JulianZJN:
## What this changes

- keep the existing five-minute structured compaction handoff budget for non-Pro efforts
- give Pro and Zero Risk Pro compaction handoffs a 50-minute default and maximum budget
- preserve an explicitly configured shorter safety ceiling
- add focused regression coverage for the effort-specific timeout policy

The previous hard five-minute ceiling caused long Pro checkpoint handoffs to fail with:

`ChatGPT compaction did not fully settle within 300000ms`

## Evidence

Before the change, a requested 60-minute handoff was unconditionally clamped to 300,000 ms. The new regression test proves that:

- non-Pro compaction remains capped at 300,000 ms
- Pro compaction defaults to and is capped at 3,000,000 ms
- a shorter explicit timeout still wins

## Scope and invariants

## Verification

- [x] I ran `bun run verify` with Bun 1.4.0.
- [x] I added a focused regression test for the behavior change.
- [ ] I manually exercised a real 5-to-50-minute Pro compaction timeout.
- [ ] I tested the changed outer Codex compaction loop through a real installed integration.
- [x] This does not change ChatGPT browser UI handling.
- [x] This does not change launcher packaging.

- macOS arm64: complete automated verification passed
- Pro account, Full mode: live 50-minute boundary not run; the installed integration will be exercised separately
- Packaged builds: not run; launcher code and packaging are unchanged

### DIFF
--- a/src/adapters/chatgpt-web/compaction-handoff.ts
+++ b/src/adapters/chatgpt-web/compaction-handoff.ts
@@ -127,7 +127,18 @@ function currentToolResults(
   return results;
 }
 
-export const MAX_COMPACTION_HANDOFF_TIMEOUT_MS = 5 * 60_000;
+export const DEFAULT_COMPACTION_HANDOFF_TIMEOUT_MS = 5 * 60_000;
+export const MAX_COMPACTION_HANDOFF_TIMEOUT_MS = 50 * 60_000;
+
+export function resolveCompactionHandoffTimeoutMs(
+  proCompaction: boolean,
+  configuredTimeoutMs?: number,
+): number {
+  const modelTimeoutMs = proCompaction
+    ? MAX_COMPACTION_HANDOFF_TIMEOUT_MS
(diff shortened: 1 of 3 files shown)


miuuyy:
Thanks for the focused patch, but I do not want to extend the Pro compaction handoff budget from 5 minutes to 50 minutes. The model can legitimately think for a while, but in my real Pro testing a structured compaction handoff that has not settled within 5 minutes is already an unhealthy or stuck handoff, not normal latency. Waiting up to 50 minutes would hide that failure and keep the browser, turn, and compaction ownership alive far too long. The proposed 5-to-50-minute path was also not exercised through a real installed integration. I am closing this rather than asking for revisions; if a reproducible healthy handoff genuinely needs more than 5 minutes, we should investigate that specific trace instead of widening the global Pro timeout.

PR #320, opened 2026-09-04
perf: capture+polling speedups; Codex 0.153.x env_context carrier

jaivial:
Three small perf/robustness fixes for the ChatGPT web adapter.

### 1. Fire-and-forget diagnostic captures

`ChatGptBrowserDiagnostics.capture()` runs `page.screenshot({timeout: 5000})` and writes PNG+JSON files to disk. The polling loop awaits every checkpoint, so a slow render can stall the whole turn for 5s waiting on a screenshot that is purely diagnostic.

Split into `capture()` (enqueue + return immediately) and `captureBody()` (the previous body). Writes still run serially through `captureQueue` so sequence numbers stay monotonic and `NN-*.png` / `NN-*.json` pairs never overlap.

### 2. Adaptive poll cadence

Three hardcoded `setTimeout(..., 250)` sleeps in the main loop become a single `pollSleep()` helper that reuses the existing `responseDomCache.snapshot` flag:
- 100ms when a snapshot exists (ChatGPT is streaming — minimize delta latency)
- 250ms otherwise (give the DOM time to settle, avoid CPU burn on empty scans)

### 3. Hard timeout for stalled turns

If the ChatGPT browser stalls past 5 minutes after submit, the polling loop now throws — the existing `try/finally` block then calls `notifyLauncherTurn({phase: end, status: failed})` and releases the broker channel. Without this, a stalled turn leaks a never-ending heartbeat and locks the broker token (this is the 'stuck in working' symptom).

### Bonus: Codex CLI 0.153.x env_context carrier

`environmentBeforeUser()` previously rejected any candidate whose `itemTurnId` didn't equal the active user's. Codex CLI 0.153.x sends the per-turn `<environment_context>` as a server-owned user message with no `internal_chat_message_metadata_passthrough` — so its identity is the server `id`, not the turn id of the active instruction.

Accept the carrier when either its turn id matches the active user or it's a server-owned item (`candidate.id` is a non-empty string). User-authored env_context inside an attacker-controlled user message wouldn't carry a server id, so the trust signal still holds.

---

Upstream already includes equivalent changes in v5.0.0+ — this branch targets the v4.0.x line for anyone running that release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,22 @@ bun run app
 
 This source path requires Bun 1.4.0. The command installs locked dependencies and opens the app.
 
+## Headless CLI mode (Linux servers)
+
+Full lifecycle from a terminal — no launcher, no Electron, no persistent Xvfb:
+
+```bash
+codex-chatgpt-web setup --full --browser-host managed-chrome --headless \
+  --chrome /usr/bin/google-chrome --login-headless auto --acknowledge-unofficial
+codex-chatgpt-web login             # headless-first; auto-falls back to ephemeral Xvfb
+codex-chatgpt-web service install   # systemd user unit (+ linger) for daemon and tunnel
+codex-chatgpt-web doctor            # xvfb, linger, browser-host, headless checks
+```
+
+- `--browser-host managed-chrome --headless` runs turns through a Playwright-managed headless Chrome.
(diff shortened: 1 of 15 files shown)


jaivial:
Closing - upstream v5.0.2 already includes equivalent fixes (captureQueue, pollSleep, hard timeout, env_context carrier). The cherry-picked fix commits have been applied directly to the fork's headless-cli-mode branch instead.

PR #329, opened 2026-09-05
fix(browser): support localized ChatGPT tool confirmation prompts

lixiangwelding:
### Summary
When ChatGPT Web is used with a Simplified or Traditional Chinese UI, tool confirmation cards (e.g. `允许 ChatGPT 使用 <app>？` or `允許 ChatGPT 使用 <app>？`) fail to be detected because `resolveChatGptToolConfirmation` in `src/adapters/chatgpt-web/browser-worker.ts` only matched the English string `Allow ChatGPT to use ${appName}?`.

Additionally, the auto-approval button matcher only recognized `/^Allow(?: once)?$/`, and the timeout denial action only recognized `Deny`.

When an unhandled approval dialog appears in Chinese ChatGPT Web:
1. The browser worker does not recognize the confirmation card.
2. ChatGPT halts waiting for approval.
3. The turn eventually times out after 60s with `stream disconnected before completion: ChatGPT stopped responding after the task started. Check the ChatGPT tab before continuing.`

### Changes
- **Dialog Title Matching**: Updated `resolveChatGptToolConfirmation` to use `chatGptToolConfirmationMatcher`, matching English (`Allow ChatGPT to use <app>?`), Simplified Chinese (`允许 ChatGPT 使用 <app>？`), and Traditional Chinese (`允許 ChatGPT 使用 <app>？`), safely escaping `appName`.
- **Approval Button Matching**: Expanded the anchored button regex to match one-shot approval actions across locales (`Allow once` / `Allow`, `允许一次` / `允许`, `允許一次` / `允許`) while keeping it strictly anchored to prevent matching persistent permissions like `Always allow` or `Allow low-risk actions` (`允许低风险操作`).
- **Deny Button Matching**: Updated the deny fallback matcher to recognize localized buttons (`Deny` / `Decline` / `拒绝` / `拒絕`).
- **Test Coverage**: Added contract tests in `tests/browser-worker-contract.test.ts` covering Simplified and Traditional Chinese approval and denial paths, as well as regex title matching.

### Verification
- `bun test tests/browser-worker-contract.test.ts` passed (120 pass, 0 fail).
- `bun run verify` passed cleanly (version sync, security audits, typecheck, all 654 tests, launcher build & runtime smoke tests).

### DIFF
--- a/src/adapters/chatgpt-web/browser-worker.ts
+++ b/src/adapters/chatgpt-web/browser-worker.ts
@@ -769,6 +769,13 @@ export async function throwIfChatGptTerminalErrorAlert(scope: ChatGptTextScope):
   );
 }
 
+export function chatGptToolConfirmationMatcher(appName: string): RegExp {
+  const escapedAppName = appName.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
+  return new RegExp(
+    `(?:Allow ChatGPT to use|允许\\s*ChatGPT\\s*使用|允許\\s*ChatGPT\\s*使用)\\s*${escapedAppName}[?？]`,
+  );
+}
+
 export async function resolveChatGptToolConfirmation(
   page: Page,
   appName: string,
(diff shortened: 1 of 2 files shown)


miuuyy:
Thanks. I reviewed our existing confirmation flow again. The documented Full-mode setup uses Allow all actions, so a confirmation card appearing with that setting would be unexpected. However, we deliberately also support per-call handling: --auto-approve-tool-calls opts into Allow once, and without that option the worker waits for a manual decision before denying an unanswered request. Localized recognition can therefore matter for that existing path.

Please confirm which connector permission setting you used and whether --auto-approve-tool-calls was enabled. Could you include the actual Simplified/Traditional Chinese dialog labels or a privacy-safe DOM sample from the failing case? The PR currently has matcher fixtures but no captured UI evidence. I am keeping it open pending that clarification rather than dismissing the confirmation path as unsupported.

miuuyy:
Closing this implementation. Optional per-call approval is supported, so the localization problem is worth retaining. However, the requested permission mode and actual Chinese card/button labels or privacy-safe DOM were not supplied; matcher fixtures built from the proposed labels do not verify the rendered UI. The branch also conflicts with main.

The report and evidence requirements are now tracked in #356.

Please coordinate in #356 before submitting a replacement. A new PR is not guaranteed to be accepted: local work may already be underway by then or may supersede this approach.

Issue #59, opened 2026-08-06
Skill invocation breaks chatgpt-web turns: "missing cwd in trusted Codex environment context"


Purin1410:
### Summary

On v2.0.0, any turn that invokes a **Codex skill** against a `chatgpt-web/*` model fails immediately with:

```
stream disconnected before completion: ChatGPT web turn is missing cwd in trusted Codex environment context
```

The same prompt in the same directory with the same model **succeeds when no skill is invoked**, so this is specific to the extra message Codex appends for a skill.

### Environment

- codex-chatgpt-web **2.0.0** (Linux x64, launcher AppImage, `mode: full`)
- Codex CLI **0.146.1**
- Ubuntu 24.04.4 LTS
- Model: `chatgpt-web/high`
### Reproduction

```bash
# succeeds
codex exec -m chatgpt-web/high "reply with exactly: ok"

# fails with "missing cwd in trusted Codex environment context"
codex exec -m chatgpt-web/high '$my-skill reply with exactly: ok'
```

Deterministic on my machine - 100% of skill-invoking turns fail, 100% of non-skill turns succeed.

### Root cause

`rawEnvironmentText()` in `src/adapters/chatgpt-web/environment.ts` locates `<environment_context>`
**positionally**: it only probes the item at `activeUserIndex - 1`.

When a skill is invoked, Codex appends an extra `<skill>…</skill>` **user** message *after* the user's
prompt, all within the same turn. The captured request body looks like this:

```
[0] developer  additional_tools
[1] developer  (base instructions)
[2] developer  <permissions instructions> …        turn_id=019fd7bd-13c4-…
[3] user       AGENTS.md + <environment_context>   turn_id=019fd7bd-13c4-…   ← cwd lives here
[4] user       (the actual prompt)                 turn_id=019fd7bd-13c4-…
[5] user       <skill>…</skill>                    turn_id=019fd7bd-13c4-…   ← appended by the skill
```

`activeUserIndex` is `5`, so `environmentBeforeUser(input, 5, turnId)` inspects item `[4]`, which has
no `<environment_context>`. None of the later fallbacks recover it either:

- `canonicalMetadataEnvironmentBeforeUser` - same positional assumption.
- the `_replayPrefixLen` loop — not a replay.
- the final historical loop — explicitly **skips same-turn items**
  (`if (!historicalTurnId || historicalTurnId === currentTurnId) continue;`), and here the
[...]

PR #336, opened 2026-09-05
fix: bind skill-injected user messages to native provenance

ncarvalhodev:
## Summary

Skill-injected Codex user messages can shift the active user item after the native environment/prompt pair. This fix recovers the trusted environment only when the injected item carries the native `skills.selected_skill_instructions` provenance marker.

## Root cause

Skill instructions are injected as a server-owned `user` item. The previous recovery path could cross an assistant response based on turn metadata alone, making skill-shaped user content insufficiently distinguished from ordinary user-authored input.

## Fix

- Use `internal_chat_message_metadata_passthrough.content_item_kinds` and require `skills.selected_skill_instructions` for skill recovery.
- Recover only from the earlier same-turn canonical environment/prompt pair.
- Preserve canonical turn, workspace, and sandbox provenance checks.
- Keep fail-closed behavior for arbitrary `<skill>` content without native provenance.

## Validation

- Focused environment tests: 49 passed, 0 failed.
- `bun run verify`: 282 passed, 0 failed, 1 skipped.
- Typecheck, audits, launcher build, and runtime smoke passed.
- `git diff --check` passed.

No versioning or release artifacts changed.

Context/follow-up: #59

### DIFF
--- a/src/adapters/chatgpt-web/environment.ts
+++ b/src/adapters/chatgpt-web/environment.ts
@@ -92,6 +92,13 @@ function itemTurnId(value: unknown): string | undefined {
   return typeof turnId === "string" ? turnId : undefined;
 }
 
+function isSelectedSkillInstruction(value: Record<string, unknown>): boolean {
+  const metadata = record(value.internal_chat_message_metadata_passthrough);
+  const kinds = metadata?.content_item_kinds;
+  return Array.isArray(kinds)
+    && kinds.some(kind => kind === "skills.selected_skill_instructions");
+}
+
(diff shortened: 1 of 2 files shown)


miuuyy:
Thanks for the contribution. We are closing this without merging because the environment recovery is not reliable enough as submitted. In review, it accepted an earlier environment across an assistant response without establishing that the environment belongs to the current turn; the released code rejects that case.

A revised submission would need a fresh review and may still not be accepted. Please do not treat this as a commitment to merge a follow-up.

PR #337, opened 2026-09-05
Add macOS Dock icon visibility setting

emreertunc:
Adds a macOS-only Settings toggle to hide the Dock icon while keeping Codex Web GPT available from the menu bar and preserving the background runtime/browser host. The preference persists, defaults off, and only hides the Dock icon when the tray is available.

Validation:
- `bun run test` (283 passed, 1 skipped)
- `bun run typecheck`
- `bun run build:renderer`


miuuyy:
Thanks for the contribution. We are not adding this setting in the current scope, so I am closing the PR without merging. The Dock visibility request in #286 can remain in the feature backlog.

emreertunc:
Thank you. Why not tho? I am using codex chatgpt web with full harness setup and dock icon makes unnecessary clutter since I don't open the app very often. Please consider adding it.

PR #340, opened 2026-09-06
fix(chatgpt-web): recover V1 subagents with null agent paths

### LINKED ISSUE #314: [Bug] [Agent Spawn]: ChatGPT web turn is missing cwd in trusted Codex environment context
### Codex Web GPT version

5.0.0

### Codex version

26.901.31953

### Platform

Windows x64

### OS version

Windows 11 25H2

### ChatGPT account tier

Pro

### Integration mode

Full harness (MCP)

### Selected model

ChatGPT Web - Pro Ultra

### What happened?

I am using ChatGPT Web — Pro Ultra through codex-chatgpt-web in Full Harness mode.

A normal ChatGPT Web turn can start successfully, and launcher checks / runtime verification are green. However, when Codex attempts to create or delegate to an agent, the ChatGPT Web route fails instead of spawning/executing the delegated agent through the configured Web/MCP route.

The final Web-route error is:

`stream disconnected before completion: ChatGPT web turn is missing cwd in trusted Codex environment context`

After this failure, Codex appears to continue execution using a native Codex model route (in my case 5.6 Luna Max) rather than remaining on the selected `ChatGPT Web — Pro Ultra` route.

This is especially problematic because my native Codex/Work quota is already exhausted. The UI then reports:

`You've hit your usage limit. Upgrade your plan or add credits to continue, or try again at Sep 11, 2026, 10:42 AM.`

Observed behavior:
- ChatGPT Web — Pro Ultra is selected.
- Launcher/runtime checks are healthy.
- The parent Web turn can begin normally.
- Agent creation/delegation begins.
- The Web stream disconnects with the missing-`cwd` error.
- The delegated work does not reliably continue through the Web/MCP route.
- Codex then appears to use a native Luna route, which immediately hits my exhausted Codex/Work quota.

Expected behavior:
- Agent/sub-agent creation should preserve the trusted Codex environment context, including `cwd`.
- Delegated agent work should remain on the configured ChatGPT Web / MCP route when that route is selected.
- A Web-route failure should be surfaced as a route failure rather than silently/implicitly falling back to a native Codex model that consumes a different quota.

### Minimal reproduction

1. Fresh Codex task.
2. Full Harness enabled and runtime verification green.
3. Select `ChatGPT Web — Pro Ultra`.
4. Run any task that causes Codex to create a sub-agent.
5. Wait for the agent creation/delegation step.
6. The Web stream fails with: `stream disconnected before completion: ChatGPT web turn is missing cwd in trusted Codex environment context`

Image

Image

### Complete final error

```text
First agent:
Reconnecting 5/5

Second agent:
You've hit your usage limit. Upgrade your plan or add credits to continue, or try again at Sep 11, 2026, 10:42 AM.
```

### Privacy-safe diagnostics

I reproduced the issue on Codex Web GPT v5.0.0 for Windows.

The issue template asks for `Activity → Export safe log`, but I could not find an `Export safe log` option in the current launcher UI. I checked both Activity and Settings.

I therefore could not attach the requested privacy-safe export. I attached screenshots showing the final error and the selected route/model state instead.

### Other Codex wrappers or route owners

none

### Preflight checklist

- [x] I tested the latest release and, for catalog or turn failures, fully restarted Codex including its background process.
- [x] I stated whether the failure also occurs in a fresh Codex task, or explained why that does not apply.
- [ ] I attached only a privacy-safe export, or explained why the exporter could not produce one; I removed any remaining private content.
- [x] I searched existing issues for this exact final error.

b-y-f:
Fixes the V1 `spawn_agent` compatibility case behind #314.

Codex 0.153.4 can create V1 thread-spawn children with `agent_path = null` while Responses metadata carries `agent_name = \"/root\"`. The existing child-lineage check required `agent_name` to match `/root/...`, so this V1 sentinel was rejected before canonical rollout recovery could run. That could surface as `ChatGPT web turn is missing cwd in trusted Codex environment context` even though the matching child rollout contained a valid cwd and workspace roots.

This change treats `/root` as the no-concrete-agent-path sentinel, while preserving provenance checks for the exact child thread, parent thread, open spawn edge, current turn, workspace/sandbox authority, and null agent path consistency across SQLite and rollout session metadata. Concrete `/root/<agent>` paths still require exact equality.

Regression coverage reproduces the Codex 0.153.4 V1 shape and verifies mismatched non-null agent paths are rejected.

Validation:
- `bun test tests/environment.test.ts`: 49 pass, 0 fail
- `bunx tsc --noEmit`: pass
- `git diff --check`: pass

Fixes #314.

### DIFF
--- a/src/adapters/chatgpt-web/codex-rollout-environment.ts
+++ b/src/adapters/chatgpt-web/codex-rollout-environment.ts
@@ -50,6 +50,10 @@ function contains(root: string, path: string): boolean {
   return rel === "" || (!rel.startsWith("..") && !isAbsolute(rel));
 }
 
+function matchesAgentPath(value: unknown, expected?: string): boolean {
+  return expected === undefined ? value == null : value === expected;
+}
+
 function canonicalRolloutName(name: string, threadId: string): boolean {
   const escapedThreadId = threadId.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
   return new RegExp(
@@ -102,7 +106,8 @@ function indexedRollout(
     if (!row) return { kind: "absent" };
     const child = "parentThreadId" in identity;
     const matchesOwner = child
-      ? row.agent_path === identity.agentName && row.parent_thread_id === identity.parentThreadId && row.status === "open"
+      ? matchesAgentPath(row.agent_path, identity.agentName)
+        && row.parent_thread_id === identity.parentThreadId && row.status === "open"
       : row.parent_thread_id == null && (row.agent_path == null || row.agent_path === "/root");
     if (typeof row.rollout_path !== "string" || !matchesOwner) {
       throw new Error(`Codex state does not authenticate the requested ${child ? "subagent" : "root thread"} rollout`);
@@ -249,10 +254,10 @@ function validateSessionMeta(
   if (item.type !== "session_meta"
     || payload?.id !== lineage.threadId
     || payload.parent_thread_id !== lineage.parentThreadId
-    || payload.agent_path !== lineage.agentName
+    || !matchesAgentPath(payload.agent_path, lineage.agentName)
     || payload.thread_source !== "subagent"
     || spawn?.parent_thread_id !== lineage.parentThreadId
-    || spawn.agent_path !== lineage.agentName) {
+    || !matchesAgentPath(spawn.agent_path, lineage.agentName)) {
     throw new Error("Codex rollout session metadata does not authenticate the requested subagent");
   }
 }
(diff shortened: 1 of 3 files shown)


miuuyy:
Thanks for identifying the V1 null-agent-path case. Your complete patch is already included in our local stabilization work for #314, together with the additional cold-fork and native V2 fixes.

I'm closing this PR without a merge while consolidating that work into one patch against the current local tree, which has diverged substantially from the published base. This is an integration decision, not a rejection of your diagnosis or contribution. The changes are still local, not released.

#314 remains the tracker. Please coordinate there before preparing another PR; local work may already cover it, and a replacement submission is not guaranteed to be accepted.

PR #343, opened 2026-09-06
launcher: localize connector diagnostics in Japanese

cXSHoz:
## Summary

- Localize MCP connector verification progress and successful Doctor diagnostics when the launcher language is Japanese.
- Preserve dynamic values such as the Responses proxy endpoint and connector name.
- Keep English, Chinese, unknown backend messages, and backend/CLI Doctor output unchanged.

## Tests

- `node --test tests/localization.test.cjs tests/renderer-wiring.test.cjs` — 23 passed
- `node --test tests/*.test.cjs` — 284 passed, 2 skipped
- `bun run typecheck`
- `bun run build:renderer`

### DIFF
--- a/launcher/src/App.tsx
+++ b/launcher/src/App.tsx
@@ -9,7 +9,7 @@ import {
   type ReactNode,
 } from "react";
 import { createPortal } from "react-dom";
-import { copyFor, type Copy } from "./i18n";
+import { copyFor, localizeRuntimeMessage, type Copy } from "./i18n";
 import { Icon, type IconName } from "./icons";
 import type {
   BrowserInteractionMode,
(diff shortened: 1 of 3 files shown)


miuuyy:
The Japanese diagnostics localization idea is preserved in #358. I am declining the current implementation because dynamic connector names are not preserved literally: running the PR function with `Native $&` produces `Native {name}` in Japanese, English, and Chinese. That also breaks the stated promise to leave English/Chinese output unchanged.

A replacement must preserve literal names/endpoints and unknown messages while changing only the intended localized UI text.

Please coordinate in #358 before submitting a replacement. A new PR is not guaranteed to be accepted: local work may already be underway by then or may supersede this approach.

PR #349, opened 2026-09-06
fix(chatgpt-web): classify stopped thinking as upstream failure

marcelokarval:
## Summary

- classify a persistent ChatGPT `Stopped thinking` state as a non-retryable upstream interruption instead of a client cancellation
- retain the terminal outcome across native reconnects without sending the prompt to the browser a second time
- verify the Responses SSE payload keeps `server_error`, `chatgpt_stopped_thinking`, and `retryable: false`

## Scope

This does not change the five-second grace window or the MCP-progress protection. Linux/X11 launcher visibility work is intentionally excluded.

## Verification

- `bun test tests/browser-worker-contract.test.ts tests/chatgpt-web-harness.test.ts tests/bridge-stall-timeout.test.ts`
- `bun run typecheck`
- `bun test`

All passed locally.

### DIFF
--- a/src/adapters/chatgpt-web/adapter-error.ts
+++ b/src/adapters/chatgpt-web/adapter-error.ts
@@ -43,11 +43,11 @@ export function chatGptTurnSupersededError(): ChatGptWebAdapterError {
 
 export function chatGptStoppedThinkingError(): ChatGptWebAdapterError {
   return new ChatGptWebAdapterError(
-    "ChatGPT remained in 'Stopped thinking' for 5 seconds, so the Codex turn was cancelled.",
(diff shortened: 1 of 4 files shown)


miuuyy:
I am declining this change as submitted. The visible `Stopped thinking` state alone does not prove an upstream failure; manual Stop and upstream interruption still need to be distinguished. Replacing the error with HTTP 502 and advising a new Codex task does not establish the cause or justify abandoning the current task.

I have retained the diagnostic distinction under #351. This is a review follow-up, not a claim that its reporter's generic stopped-responding error was caused by this particular UI state.

Please coordinate in #351 before submitting a replacement. A new PR is not guaranteed to be accepted: local work may already be underway by then or may supersede this approach.

PR #354, opened 2026-09-07
fix(launcher): rediscover uncached tunnel health before monitoring

DJNgoma:
## What this changes

Discover an existing tunnel's loopback health endpoint when the launcher has not cached it yet, before monitoring falls back to runtime inventory. This lets the existing MCP diagnostics check detect a wedged dispatcher even when inventory reports readiness.

The change reuses the existing endpoint validation, diagnostics probes, and restart policy. It also adds platform-specific recovery guidance and links from the English, Japanese, and Chinese READMEs. There are no protocol, provider, packaging, or dependency changes.

## Evidence

Previously, an uncached endpoint made local probes unobservable and could leave monitoring dependent on inventory readiness. The new regression supplies a real isolated loopback diagnostics server with a recent internal MCP HTTP 502 event: monitoring discovers the endpoint and returns `ready: false`, `statusKnown: true`, and `fatal: true`.

Additional regressions cover cached discovery, command failure, malformed JSON, and rejection of a non-loopback endpoint. When both discovery and inventory are unavailable, the observation remains unknown rather than claiming recovery.

## Scope and invariants

- [x] Read `CONTRIBUTING.md`; remaining account-bound acceptance gates are explicitly recorded below.
- [x] One focused fix, regression coverage, and related recovery documentation only.
- [x] Model, route, effort, connector, and capability selection remain unchanged.
- [x] Full mode retains its existing turn-bound capability; Browser-only gains no broker or connector.
- [x] No credentials, browser state, tunnel IDs, raw logs, private paths, generated artifacts, version changes, or unrelated dependencies are committed.

## Verification

Fresh verification on 7 September 2026, on this exact branch based on upstream `c648c09`, using Bun 1.4.0 and isolated home, Codex, and application-state directories:

- [x] `bun install --frozen-lockfile` in the root and `launcher/`.
- [x] `bun run verify`: **655 runtime tests passed; 286 launcher tests passed, 1 platform-specific test skipped; 0 failures**.
- [x] Root and launcher audits: no vulnerabilities found.
- [x] Version consistency, both TypeScript checks, launcher production build, runtime bundle, and third-party notices generation passed.
- [x] Relocated runtime smoke: `RELOCATABLE_RUNTIME_SMOKE_OK`.
- [x] `git diff --check`; one commit with no unrelated fork commits.
- [ ] Installed Codex Full-mode recovery acceptance.
- [ ] Native macOS, Windows, and Linux package build/smoke validation.

Automated verification: macOS arm64. The deterministic regression uses an isolated loopback server, not a live account-bound tunnel. Installed-app recovery and cross-platform packaged acceptance were **not run**; no active runtime was replaced and no user service was restarted. This PR remains a draft for those acceptance checks and upstream CI review.

### DIFF
--- a/README.ja.md
+++ b/README.ja.md
@@ -160,6 +160,11 @@ ChatGPT のツール呼び出しを現在の Codex タスクへ接続します�
 アンインストール前の Codex 統合削除も行えます。すべてのブラウザーチェックポイントでスクリーンショットが必要な場合にのみ、
 `CODEX_CHATGPT_WEB_BROWSER_DIAGNOSTICS=1` を設定してください。
 
+ChatGPT の通常の応答は届くのに Codex Native2 のツール呼び出しが `502` を返す場合は、
+モデル経路とは別にツール経路で問題が発生している可能性があります。OS 別の復旧手順は
+[トラブルシューティング](TROUBLESHOOTING.md#every-codex-native2-tool-call-returns-502-but-chatgpt-text-still-works)
+を参照してください。トンネルを再起動する前に、実行中のタスクを完了またはキャンセルしてください。
+
 新規インストールでは、クロスバックエンドのサブエージェントに **Compatibility V1** を使用します。
 **Native** は Codex 独自の機能設定を維持し、プレーンテキストの Web-to-Web V2 delegation を有効にします。
 プロトコル変更後は Codex を再起動し、新しいタスクを開始してください。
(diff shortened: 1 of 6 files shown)


miuuyy:
Thanks for identifying the uncached diagnostics path. We have now implemented our own fix: every tunnel start/adoption discovers and verifies the current MCP diagnostics endpoint before monitoring begins, and unknown diagnostics no longer become healthy merely because inventory reports ready.

This addresses the startup cause rather than discovering the endpoint only after monitoring has started. The implementation has passed read-only checks on both real local profiles and the supervisor regression suite, and is included in the upcoming v5.0.5 patch. Tracking: #372.

Closing this draft in favor of that implementation. Thanks for the investigation; we are not planning to merge this alternative patch, and a replacement submission may overlap work already completed locally.

Issue #324, opened 2026-09-05
Scope the Codex Integration to a model_provider Instead of Overriding openai_base_url Globally


JulianZJN:
## Summary

`setup` installs the bridge by writing a global `openai_base_url` into `~/.codex/config.toml`. Every Codex request then travels through the local Responses proxy, including requests for first-party models that have no relationship to the ChatGPT web session.

The result is a hard dependency. When the bridge is unavailable for any reason, the user does not lose the Web models and keep the rest. Codex stops working entirely until the bridge is repaired.

I would like to propose scoping the integration to a named `model_provider` instead. I have verified that this works against Codex 0.153.0, and the measurements are below.

## Why the Global Override Hurts

A global base URL couples two independent things: which models are offered, and whether Codex can reach any model at all. The bridge is a single user-space process with no supervisor installed by default, so its availability is far lower than the availability of the official endpoint it replaces. Three failure modes hit me on one machine within roughly 36 hours.

**The launcher owns the proxy and nothing restarts it.** `service status` reports `installed: false, loaded: false` in launcher mode, so the Responses proxy lives and dies with the Electron app. Quitting that app, or having it exit for any other reason, closes port 17841. Codex then shows a reconnect banner and no model works. Recovery requires relaunching the app and restarting Codex, because a TCP-level reconnect alone leaves Codex paired with the previous proxy instance.

**Upstream connections do not survive a network change.** After moving between a campus network and a phone hotspot, every request returned:

```
502 upstream_server_error
The socket connection was closed unexpectedly.
url: http://127.0.0.1:17841/v1/responses
```

The process was alive, port 17841 was listening, `doctor` reported `Doctor result: ready`, and the launcher's own browser smoke test completed a full ChatGPT turn, because Electron does its own networking. Only the proxy's `fetch` layer held dead sockets. Restarting the launcher fixed it. Nothing short of that did.

This failure is difficult to diagnose because every local health signal is green. The only reliable discriminator I found was calling the proxy with a real bearer token from `~/.codex/auth.json`:

```
GET http://127.0.0.1:17841/v1/models?client_version=<v>
```

A request without a bearer always returns `502 Native Codex passthrough requires the incoming Bearer authorization`, so an unauthenticated probe cannot distinguish a healthy proxy from a broken one. With a valid bearer, 200 means the upstream is fine and 502 means the connection pool is stale. Consider surfacing this distinction in `doctor`, since the current output reports ready in both states.

[...]

PR #360, opened 2026-09-07
feat: provider-only route mode defers Codex routing to an external router

sushmitxo:
Implements the external-provider route mode from the CGW_OPENCODEX handoff. This is an unsupported local fork patch; upstream defaults are unchanged (managed).

## Summary
- New `codexRouteMode` setting (`managed` default | `external-provider`)
- Provider-only mode NEVER reads or writes `~/.codex/config.toml`; setup, launcher bridge reconcile/connect/restore, and subagent mutations fail closed
- Setup route-mode transition unwinds an existing journal before switching ownership
- Server serves the Web model catalog locally when no native bearer exists and rejects non-Web native turns locally
- CLI `--codex-route-mode` / route status + doctor surface the mode

## Tests
- 5 new src tests (codex-integration) + 4 new launcher tests (runtime-host), all green
- src typecheck passes; launcher runtime.cjs/main.cjs parse cleanly

Full detail is delivered in `CGW_OPENCODEX_PROVIDER_IMPLEMENTATION_HANDOFF.md` (not committed).

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -172,7 +172,14 @@ ChatGPT endpoint so Voice session creation never falls through to the Responses-
 assignments are journaled and restored exactly on disconnect or uninstall; a conflicting existing
 Voice route requires explicit `--replace-codex-route` ownership. The daemon forwards the
 authenticated official model catalog and appends only the routed models owned by the
-`chatgpt-web/` namespace; no static catalog is installed. Subagent protocol selection is explicit,
+`chatgpt-web/` namespace; no static catalog is installed. An advanced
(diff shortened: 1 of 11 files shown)


miuuyy:
Thanks for documenting the provider-only implementation. We are not taking external-provider routing into the upstream project; this direction was already declined in #205 and #324.

This adds another route-ownership contract across setup, recovery and model discovery. It is outside the current supported architecture and is not needed to consolidate the stabilization patch, so I am closing it without a merge. The proposal remains recorded in those issues and this PR.

Please coordinate before submitting another implementation of this direction; a replacement PR would not imply acceptance.

PR #361, opened 2026-09-07
feat(launcher): add macOS menu bar only mode

### LINKED ISSUE #286: [Feature] Hide ChatGPT Web helper icon from macOS Dock

TheCrazyLex:
## Summary

Adds a macOS-only **Hide Dock icon** preference so Codex Web GPT can run from the menu bar alone, plus the two fixes that mode depends on to be safe.

Most of the plumbing already existed — the tray, `keepRunningOnClose` hiding the window instead of quitting, and `--hidden` startup. This adds the missing piece and repairs what breaks once the Dock icon can go away.

Fixes #286.

## The preference

**`launcher/electron/dock-visibility.cjs` (new)** — a serialized controller with injected `dock` / `hasTray` / clock, so it is unit-testable without Electron.

Electron's `DockShow` refreshes `last_dock_show_` on every call, and `DockHide` is a documented no-op within one second of a show. So showing an already-visible Dock icon silently swallows the next hide and leaves the toggle disagreeing with reality. The controller therefore:

- skips redundant shows, and defers a hide until the guard interval has elapsed
- coalesces concurrent requests, dropping a deferred hide that has been superseded
- re-probes the tray after the wait, and refuses to hide without one
- verifies each transition with `dock.isVisible()` rather than assuming the call took

**`launcher/electron/main.cjs`** — the preference is written **only after macOS applies the change**, and the Dock rolls back if that write fails, so the switch and the Dock cannot disagree in either direction. A saved preference that cannot be applied at startup is kept and logged, leaving the Dock icon reachable.

It is deliberately independent of `keepRunningOnClose`: a tray-accessible app that quits on window close is a legitimate configuration, and the existing close handler already shuts down cleanly there.

## Two prerequisite fixes

**The macOS menu bar icon was an empty image.** `trayImage()` built the glyph from an `image/svg+xml` data URL, but `nativeImage` decodes only PNG and JPEG — so the status item had nothing to draw. Survivable while a Dock icon exists; a lockout once it can be hidden, and it made the tray guard hollow. The glyph now ships as a template PNG (`assets/trayTemplate.png` + `@2x`, same geometry), and `createTray` refuses an image that fails to decode.

**macOS never started hidden at login.** `args` is documented Windows-only and `openAsHidden` stopped working on macOS 12, below the app's 13.0 floor — so `--hidden` never reached argv and `startHidden` was never true on a login launch. Startup now reads `getLoginItemSettings().wasOpenedAtLogin`; Windows and Linux are unchanged. Relatedly, restoring a maximized window at `ready-to-show` calls `NativeWindowMac::Maximize`, which shows a hidden window, so that restore is deferred until the window is actually presented.

## Scope

Nothing changes in the browser host, runtime supervisor, or turn lifecycle. Hiding the Dock icon changes the macOS activation policy, not window visibility — and the hidden-window path this relies on is the one `keepRunningOnClose` has exercised by default all along. Windows/Linux taskbar hiding from #291 is left out to keep this reviewable.

## Verification

`bun run verify` — exit 0. Core **655 pass / 0 fail**; launcher **299 pass / 0 fail / 1 expected platform skip**; typecheck, renderer build, runtime bundle, notices and relocatable runtime smoke all pass; both audits clean.

## Credit

@emreertunc wrote the original macOS Dock preference in #337 — the state key, IPC shape, Settings row and "never hide without a tray" ordering are his design and I kept them. @wpfyorg reported #286 and wrote the first cross-platform attempt in #291.


TheCrazyLex:
@miuuyy, could you take a look at this one? Menu-bar mode is widely requested, and this is a robust implementation of it that supersedes the closed #337.

cc @emreertunc @wpfyorg

miuuyy:
Thanks for the detailed work on menu-bar mode, including the tray and login-start behavior. We are deferring this feature while consolidating the current stabilization patch. This PR also changes Dock transitions, tray assets and startup behavior, so it needs a separate macOS review and acceptance pass rather than being folded into the bug-fix release.

I'm closing it without a merge to keep the active PR queue focused. The feature and its prerequisites remain tracked in #286, with this PR preserved as a reference. This is a scope/timing decision, not a claim that the implementation is broken.

Please coordinate on #286 before opening a replacement: our local tree has diverged substantially and we may implement it differently; another submission is not guaranteed to be accepted.

Issue #321, opened 2026-09-04
[Bug] Error running remote compact task: stream disconnected before completion: ChatGPT did not complete the context handoff. Retry the task.


hankok:
### Codex Web GPT version

5.0.2

### Codex version

26.901.22334

### Platform

Windows x64

### OS version

Windows 11 Pro

### ChatGPT account tier

Plus

### Integration mode

Full harness (MCP)

### Selected model

High

### What happened?

- Codex stuck on compact task (stucks 8-10 minutes). 
`Context automatically compacting`
`Error running remote compact task: stream disconnected before completion: ChatGPT did not complete the context handoff. Retry the task.`

The above loop repeats if retry chats. 

- Sometimes if lucky, the compact task completes before the error hits. 

### Minimal reproduction

as described in "What happened?"

### Complete final error

```text
Error running remote compact task: stream disconnected before completion: ChatGPT did not complete the context handoff. Retry the task.
```

### Privacy-safe diagnostics

[codex-web-gpt-diagnostics-2026-09-04.json](https://github.com/user-attachments/files/31854653/codex-web-gpt-diagnostics-2026-09-04.json)

### Other Codex wrappers or route owners

NA

### Preflight checklist

- [x] I tested the latest release and, for catalog or turn failures, fully restarted Codex including its background process.
- [x] I stated whether the failure also occurs in a fresh Codex task, or explained why that does not apply.
- [x] I attached only a privacy-safe export, or explained why the exporter could not produce one; I removed any remaining private content.
- [x] I searched existing issues for this exact final error.

miuuyy:
I found the actual v5.0.2 run in your attachment: `1426c7570120_fallback`. Parts 1 and 2 were acknowledged; the final send failed on a five-second DOM observation timeout. Earlier duplicate-turn and five-minute errors in this file came from v5.0.1, before the 23:11 UTC upgrade.

The existing safe log is sufficient to investigate why observation recovery is missing in this compaction path. To understand the underlying renderer stall, please attach the privacy-reviewed JSON files `21-prompt-attachment-complete`, `22-file-attachment-complete`, `23-send-ready`, and `24-turn-failed` from `.codex-chatgpt-web/diagnostics/browser-turns/1426c7570120_fallback-9c54d7af`. If that directory has already rotated away, please say so; no need to recreate the full long task just for those files.

hankok:
[...]

PR #362, opened 2026-09-07
fix: read the current environment envelope per content part

MaxKerkula:
## The failure

A thread dies permanently at its first context compaction. The handoff itself succeeds; the
continuation immediately after does not:

```
15:32:04.883  broker trace=0903cffd4623 accepted structured compaction handoff
15:32:05.475  browser turn 0903cffd4623 failed: ChatGPT web turn aborted
15:32:07.291  trusted environment unavailable (thread_id=present, turn_id=present,
              previous_response_id=none, replay_prefix_items=0, context_messages=38)
   ... x6 with backoff, then Codex gives up on the turn
```

Seen on v5.0.4, Windows, Full mode, `chatgpt-web/pro`.

## Cause

`extractChatGptContinuationEnvironmentClaim` matched the envelope against `rawMessageText(item)` —
every content part joined with a newline — while `environmentBeforeUser` and
`canonicalMetadataEnvironmentBeforeUser` matched it against each part on its own.

That disagreement is invisible until compaction. When Codex installs a compacted history it
rebuilds its preamble into one server-owned user item. From the captured `replacement_history`,
item 34 has three parts:

| part | content |
|---|---|
| 0 | `<recommended_plugins>…` |
| 1 | `# AGENTS.md instructions <INSTRUCTIONS>…` |
| 2 | `<environment_context><cwd>…</cwd>…</environment_context>` |

Before compaction the envelope is its own item and matches. After, the *joined* text starts with
`<recommended_plugins>`, so the anchored regex finds zero claims and the parser throws
`"Compaction continuation requires one current native environment claim"` from
`ChatGptThreadEnvironmentStore.resolve` (thread-environment.ts:158) — before the native rollout
recovery at lines 165-181, which had the correct environment in the turn's own `turn_context`.

Replaying the real 37-item `replacement_history` through the exported parser:

```
v5.0.4    CLAIM THREW: Compaction continuation requires one current native environment claim
patched   CLAIM OK: cwd C:\...\AATS | workspaceWrite | roots: 2
```

The recovered authority matches the `turn_context` Codex wrote for that turn, so the
`sameAuthority` cross-check passes and the continuation proceeds.

## The change

Read envelopes one content part at a time through a single helper the three call sites share, so
they cannot drift apart again.

Two deliberate details, both from adversarial review of an earlier draft of this patch:

- **The helper reads a content array only.** `rawMessageText` also flattened a bare string
  `content`. Carrying that into the helper let a whole-message string reach `environmentBeforeUser`,
  which is on the primary authority path and has no metadata/workspace binding — a shape that path
  never accepted. Verified as a real widening before it was removed: with the string branch, an
  input claiming `<cwd>C:\</cwd>` with `dangerFullAccess` was granted despite client metadata
  declaring a different workspace. The claim parser keeps its own string handling, because a claim
  is only cross-checked against the rollout at thread-environment.ts:175 and never becomes
  authority on its own.
- **Identical claims are deduplicated.** Per-part reading finds envelopes the joined text used to
  hide, so one turn can surface the same envelope twice — once from Codex, once quoted back in a
  current-turn message. Identical text is one claim. Genuinely different claims stay ambiguous and
  still fail closed. I could not reach the duplicate shape through `resolve()` with a realistic
  payload (the primary path resolves first in every ordering I tried), so its regression test calls
  the exported parser directly rather than pretending otherwise.

## Tests

Three regression tests in `tests/environment.test.ts`, each verified to fail without the specific
line it guards:

| test | fails against |
|---|---|
| current environment part beside Codex's preamble parts | v5.0.4 as shipped |
| whole-message string envelope is not authority-bearing | the draft with the string branch |
| duplicate claims are one claim, conflicting ones fail closed | the draft without dedupe |

## Verification

- `bun x tsc --noEmit` — clean.
- `bun test tests/environment.test.ts` — 51 pass, 0 fail.
- `bun test tests/*.test.ts` — 521 pass, 135 fail. Those 135 are **pre-existing on the pristine
  `v5.0.4` tag** in my Windows environment; I captured the failure set before and after and diffed
  it — byte-identical, no new failures. `bun run verify` fails for that same pre-existing reason,
  so I did not use it as evidence.

## Not fixed here, but noticed

`contextualUserMessage` (environment.ts:135) still classifies on joined text, so Codex's multi-part
preamble is not recognized as contextual and `extractChatGptTurnUserRevision` will hand it back as
the human instruction whenever that item is the last user item. The compaction ordering (preamble
before the replayed source) hides it today. It is the same joined-vs-per-part disagreement in a
different recognizer, but fixing it means teaching the contextual vocabulary about
`<recommended_plugins>` and AGENTS.md blocks, which is a larger change than this bug needs. Happy
to open it separately if you want it.

### DIFF
--- a/src/adapters/chatgpt-web/environment.ts
+++ b/src/adapters/chatgpt-web/environment.ts
@@ -240,6 +240,31 @@ export function isChatGptCompactionContinuation(parsed: CodexParsedRequest): boo
     && isAcceptedCompactionContinuation(parsed, identity, revision);
 }
 
+const ENVIRONMENT_ENVELOPE = /^<environment_context>[\s\S]*<\/environment_context>$/;
+
+function environmentEnvelope(text: unknown): string[] {
+  if (typeof text !== "string") return [];
+  const trimmed = text.trim();
+  return ENVIRONMENT_ENVELOPE.test(trimmed) ? [trimmed] : [];
+}
+
+/**
+ * Return the environment envelopes one message carries, one content part at a time.
+ *
+ * Codex always sends the envelope as its own part. When it rebuilds the preamble for a compacted
+ * history it keeps that part but adds sibling parts around it (recommended plugins, AGENTS.md
+ * instructions), so the joined message text is not an envelope and must never be the unit of
+ * recognition; every caller here has to agree on the part, or a valid current update disappears.
+ *
+ * Only a real content array is read. Codex never sends the envelope as a bare message string, and
+ * that shape has never carried filesystem authority on these paths, so it must not start now.
+ */
+function environmentEnvelopeParts(item: Record<string, unknown>): string[] {
+  return Array.isArray(item.content)
+    ? item.content.flatMap(part => environmentEnvelope(record(part)?.text))
+    : [];
+}
+
 /** Parse a claim only: the caller must compare it with this turn's native rollout authority. */
 export function extractChatGptContinuationEnvironmentClaim(parsed: CodexParsedRequest): ChatGptTurnEnvironment {
   const turnId = extractChatGptTurnIdentity(parsed).turnId;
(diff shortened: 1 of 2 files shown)


miuuyy:
Thanks for reporting and investigating the continuation failure. The underlying grouped-environment problem is already covered by our own per-content-part implementation, which was in our local release work and has now shipped in v5.0.5.

I am closing this PR because merging it would duplicate that fix. The remaining refactor and duplicate-claim behavior do not have a demonstrated additional native failure that we need to address. Existing compaction reports remain tracked in #321; this is not a claim that every compaction failure has the same cause.

Thanks again. If you find a distinct failure on the released implementation, please open an issue with the native request shape or safe diagnostics before preparing another patch.

PR #365, opened 2026-09-07
fix(chatgpt-web): normalize Windows rollout path identity

Ahmet1991:
Windows Codex state can store an extended-length rollout path (`\\?\...`) while the bridge derives the sessions root with the normal drive spelling. Comparing those forms directly can reject the active rollout even though both paths identify the same file.

This normalizes both sides with `toNamespacedPath()` before containment/identity checks and adds a Windows regression test covering a namespaced `state_5.sqlite` rollout path.

Validation:
- `git diff --check` passes.
- Local Bun verification could not run because Bun is not installed/on PATH on this machine; the PR CI matrix should exercise the regression on Windows, macOS, and Linux.

### DIFF
--- a/src/adapters/chatgpt-web/codex-rollout-environment.ts
+++ b/src/adapters/chatgpt-web/codex-rollout-environment.ts
@@ -10,7 +10,7 @@ import {
   readdirSync,
   realpathSync,
 } from "node:fs";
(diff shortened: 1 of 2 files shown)


miuuyy:
Thanks for investigating this. The fix is now covered by our local v5.0.5 implementation, so I am closing this alternative and shipping the audited local patch. Tracking: #376.

PR #366, opened 2026-09-07
fix(scripts): make verification portable on Windows

Ahmet1991:
Windows packaged runtimes can provide `bun.exe` without a `bunx` shim, so `bun run verify` stops at `bunx tsc --noEmit`. PowerShell also does not expand `tests/*.test.ts` before Bun receives it.

This keeps the scripts shell-independent by using Bun's local binary resolution (`tsc --noEmit`) and passing the test directory directly (`bun test ./tests/`).

Validation on Windows with the packaged Bun 1.4.0 runtime:
- `bun run typecheck` completed successfully through `bun run verify`.
- `bun run test` now executes the suite instead of matching zero files; 654 tests passed before the known installed-launcher discovery test failed because this machine has the launcher installed. That separate purity bug is submitted independently.
- `git diff --check` passes.


miuuyy:
Thanks. This changes the development workflow and needs to be evaluated across the full verification setup. I am keeping it outside the current runtime patch and preserving the proposal in #378.

miuuyy:
@Ahmet1991 Following up on your coordination message in #378: both package-script changes are already covered by the implementation released in [v5.0.6](https://github.com/miuuyy/codex-chatgpt-web/releases/tag/v5.0.6). There is no need to rebase or resubmit a PR for this scope. Thanks for checking it on Windows.

Replying here because #378 was converted to a discussion, which has since been removed.

PR #367, opened 2026-09-07
fix(dev-profile): keep injected launcher discovery pure

Ahmet1991:
`installedLauncherCandidates()` accepts an injected platform/environment for deterministic path discovery, but the Windows branch still queried the real machine registry. On a development machine with Codex Web GPT installed, that overrides the injected `LOCALAPPDATA` and makes the supposedly synthetic lookup impure.

Only consult the real Windows registry when the caller is using the real `process.env`. Production behavior is unchanged; injected test/dev environments stay deterministic.

Validation:
- `bun test tests/dev-profile.test.ts` — 5 pass, 0 fail.
- The previously failing `installed launcher discovery has explicit platform candidates` case passes on a machine where the launcher is actually installed.
- `git diff --check` passes.

### DIFF
--- a/src/dev-chat/profile.ts
+++ b/src/dev-chat/profile.ts
@@ -133,7 +133,11 @@ export function installedLauncherCandidates({
     );
   } else if (platform === "win32") {
     const registeredLocation = windowsInstallLocation?.trim()
-      || (process.platform === "win32" ? registeredWindowsLauncherInstallLocation() : undefined);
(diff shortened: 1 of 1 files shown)


miuuyy:
Thanks for identifying the discovery-purity issue. This needs to be handled consistently across isolated and ordinary profile discovery. I am not taking this implementation into the current patch; the case is preserved in #379.

PR #368, opened 2026-09-07
fix(browser-host): invalidate retained chats after navigation

Ahmet1991:
A retained manual/Native2 conversation can remain marked reusable after its tab navigates. The next resume then sends only the incremental prompt into a page that no longer represents the retained conversation.

Invalidate the retained conversation on navigation so the next resume rebuilds the full context instead of reusing stale page ownership.

Validation:
- `node --test launcher/tests/browser-host.test.cjs` — 96 pass, 0 fail.
- Includes a focused regression: `navigating a retained manual chat invalidates it before the next resume`.
- `git diff --check` passes.

### DIFF
--- a/launcher/electron/browser-host.cjs
+++ b/launcher/electron/browser-host.cjs
@@ -869,6 +869,15 @@ class BrowserHost {
 
   bindManualTurnContents(tab) {
     const contents = tab.view.webContents;
+    const invalidateRetainedConversation = () => {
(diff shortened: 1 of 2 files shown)


miuuyy:
Thanks for the report. We now have our own local v5.0.5 implementation covering the retained-page lifecycle, including active continuations. Closing this alternative; tracking: #377.

PR #369, opened 2026-09-07
fix(chatgpt-web): refuse an archived rollout record, not the turn

Ahmet1991:
Codex can move a session into `archived_sessions` while its state index still points at the moved rollout. The bridge correctly refuses to read outside `sessions/`, but currently turns that refused record into a fatal turn error even when the same thread already has trusted cached authority.

Treat an out-of-sessions candidate as an unusable record and fall back to previously proven thread authority. The archived file is never opened. Fresh/subagent recovery still fails closed when no trusted authority exists.

Validation:
- `bun test tests/environment.test.ts` — 49 pass, 0 fail.
- Regression covers an indexed archived rollout and confirms outside-path recovery still fails closed.
- `git diff --check` passes.

### DIFF
--- a/src/adapters/chatgpt-web/codex-rollout-environment.ts
+++ b/src/adapters/chatgpt-web/codex-rollout-environment.ts
@@ -578,8 +578,24 @@ export function resolveCurrentCodexRolloutEnvironment(options: {
   }
 
   const matching: ChatGptTurnEnvironment[] = [];
+  let unusableCandidates = 0;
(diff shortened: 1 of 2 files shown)


miuuyy:
Thanks. The archived-record case is worth tracking, but the proposed handling does not sufficiently distinguish it from other lookup failures or preserve all existing validation boundaries. Closing this implementation and keeping the proposal in #380.

PR #370, opened 2026-09-07
fix(chatgpt-web): preserve authority across ambient environment diffs

Ahmet1991:
Codex can emit a cwd-less `<environment_context>` when only ambient facts change, for example when the local date rolls over mid-turn. The bridge currently notices that diff as a current environment claim but cannot extract a cwd from it, which can discard the thread authority established moments earlier and fail the turn.

Only treat envelopes that actually state a cwd as current authority claims. Cwd-less ambient diffs may keep the already-proven thread authority, while their workspace roots are still checked so a widened root remains fail-closed.

Validation:
- `bun test tests/environment.test.ts` — 52 pass, 0 fail.
- Four focused regressions cover trusted fallback, widened roots, untrusted threads, and explicit cwd claims.
- `git diff --check` passes.

### DIFF
--- a/src/adapters/chatgpt-web/environment.ts
+++ b/src/adapters/chatgpt-web/environment.ts
@@ -111,12 +111,25 @@ export function hasRawChatGptEnvironmentContext(parsed: CodexParsedRequest): boo
   });
 }
 
+const ENVELOPE_PATTERN = /<\/?environment_context\b/i;
+const CWD_TAG_PATTERN = /<\/?cwd\b/i;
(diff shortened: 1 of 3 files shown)


miuuyy:
I rechecked this against current native Codex requests. The ordinary cwd-less/date-update case passes with our existing handling; the reported runtime failure was not reproduced. I am not accepting a broader authority fallback without a matching failing native trace. Related earlier cwd-less context tracking: #288.

Issue #308, opened 2026-09-04
[Bug] Chatgpt freeze after search


MaxAndolini:
### Codex Web GPT version

5.0.0

### Codex version

Latest

### Platform

Windows x64

### OS version

Windows 11 25H2

### ChatGPT account tier

Plus

### Integration mode

Full harness (MCP)

### Selected model

ChatGPT Web - High

### What happened?

After typing in the prompt, chatgpt freezes and eventually gives a timeout error. Sometimes it works without problems, but when this happens, I've tried closing and reopening it several times and it doesn't work at all. I've tried turning Bigger context off and on, but the result is the same.

### Minimal reproduction

1. Run a Codex task with ChatGPT Web.
2. Let the task run for a long time until the Web response is cancelled or becomes unresponsive.

### Complete final error

```text
https://prnt.sc/ssvU8F3xJNAC

stream disconnected before completion: ChatGPT stopped responding after the task started. Check the ChatGPT tab before continuing.
```

### Privacy-safe diagnostics

[codex-web-gpt-diagnostics-2026-09-04.json](https://github.com/user-attachments/files/31828682/codex-web-gpt-diagnostics-2026-09-04.json)

### Other Codex wrappers or route owners

none

### Preflight checklist

- [x] I tested the latest release and, for catalog or turn failures, fully restarted Codex including its background process.
- [x] I stated whether the failure also occurs in a fresh Codex task, or explained why that does not apply.
- [x] I attached only a privacy-safe export, or explained why the exporter could not produce one; I removed any remaining private content.
- [x] I searched existing issues for this exact final error.

MaxAndolini:
@miuuyy 

I tried with the latest version, v5.0.2, and the situation is still the same. If I do something with the prompt in a chat window, it doesn't work later; the chatgpt web screen freezes and gives an error in the browser. However, I can copy the same chat with Markdown and run it in a new chat. It seems like there's a problem with the system, caching, or browser. Also, even if I verify it on the MCP side, it always appears as if it hasn't been verified in the interface. Could you please fix that as well? It's annoying.

<img width="868" height="712" alt="Image" src="https://github.com/user-attachments/assets/7f9c9e63-5b8a-4005-854c-fe55b11202fd" />

miuuyy:
I found why the MCP wizard can show `Verify runtime` again after leaving and reopening the section: its displayed verification result is local component state. That part does not need another log.

[...]

Issue #307, opened 2026-09-04
[Bug][Windows] Cancelling a ChatGPT Web turn leaves the same Codex task stuck on the next message


Nubfall:
## Summary

On Windows Full Harness, cancelling an in-progress ChatGPT Web response once can leave the same Codex conversation unrecoverable. Sending the next message in that conversation makes the ChatGPT Web browser surface remain stuck or unresponsive, followed by repeated reconnect attempts. Restarting Codex does not recover the task; creating a new Codex conversation and manually importing the previous context is the only reliable workaround.

## Environment

- Windows 11 x64
- Codex Web GPT 5.0.0 runtime
- Official packaged launcher
- Full Harness / MCP
- Connector: Codex Native2
- ChatGPT Web model (High / Extra High)
- No Cockpit mux or other wrapper involved

## Reproduction

1. Start a long-running Codex task through ChatGPT Web in an existing Codex conversation.
2. While the ChatGPT Web response is still in progress, force-cancel/interrupt the Codex response once.
3. Send a new message in the same Codex conversation.
4. Observe that the embedded ChatGPT Web page remains stuck in a thinking/unresponsive state.
5. Wait for the automatic reconnect attempts to finish.

## Actual result

Codex eventually reports:

~~~
stream disconnected before completion: ChatGPT stopped responding after the task started. Check the ChatGPT tab before continuing.
~~~

The error repeats after reconnect attempts. Restarting Codex does not restore the affected conversation. A new Codex conversation works only after manually carrying over the old context.

## Local lifecycle evidence

A sanitized read of the launcher log around the affected run shows the following pattern:

1. A prior browser turn is aborted and its browser tab is released.
2. The next turn creates a fresh browser turn and accepts the Web prompt.
3. The next response path enters response_page_rebind_1 and times out after approximately 60 seconds, or is aborted while the page is still in the response lifecycle.
4. The broker then rejects release/claims associated with a retired turn.
5. Later recovery attempts create additional browser turns; a global cleanup reports multiple active objects at once (http=2, browser=7) before cancelling them.

Representative runtime records include:

~~~
browser turn ... failed: ChatGPT web turn aborted
browser turn ... stage=response_page_rebind_1 failed ... ChatGPT browser stage timed out: response_page_rebind_1
broker rejected release ... retiredTurn=...
runtime.active_turns_cancelled ... http=2,browser=7
~~~

[...]

PR #371, opened 2026-09-07
fix(launcher): hand off running retained turns safely

Ahmet1991:
When a follow-up/steering turn arrives while the retained Automatic conversation is still running, the launcher currently cannot hand that tab to the waiting turn. A failed owner can also strand waiters indefinitely.

This adds a bounded FIFO handoff for the same conversation key, drains waiters on every terminal status (`completed`, `failed`, `aborted`), and clears handoff timers on takeover, timeout, helper exit, and tab destruction.

Validation:
- `node --test launcher/tests/browser-host.test.cjs` — 98 pass, 0 fail.
- Regressions cover same-conversation steering reuse and failed-turn handoff.
- `git diff --check` passes.
- `bun test tests/retained-compaction.test.ts` hits the known Windows single-file named-pipe scheduling hang documented by the original change; the launcher suite completes cleanly.

### DIFF
--- a/launcher/electron/browser-host.cjs
+++ b/launcher/electron/browser-host.cjs
@@ -352,6 +352,7 @@ class BrowserHost {
     this.visible = false;
     this.surfaceActive = true;
     this.turnTabs = new Map();
+    this.automaticTurnHandoffs = new Map();
     this.closedTurnOwners = new Map();
     this.userCancelledTurnOwners = new Map();
     this.manualTerminalSignals = new Map();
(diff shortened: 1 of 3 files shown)


miuuyy:
Thanks. This changes active-turn handoff without sufficiently accounting for the surrounding cancellation, cleanup and retention lifecycle. I am keeping our current implementation and closing this proposal. The underlying interrupted-task continuation case is already tracked in #307 and #308.

PR #373, opened 2026-09-07
fix(browser): serialize heavy page-driving phases across concurrent turns

Ahmet1991:
Five ChatGPT browser tabs can run concurrently, but the expensive page-driving phases share one Electron renderer pool. Concurrent DOM setup/rebind/multipart-send work can starve another turn until its DOM probe times out even when both turns are otherwise healthy.

Add a single-owner FIFO lock only around the heavy page-driving stages. Waiting for model output remains outside the lock, so normal concurrent reasoning still proceeds. The lock is re-entrant per turn, reclaims dead helpers, skips dead waiters, releases on every terminal path, and sizes its wait budget for the full five-tab field.

This also includes the follow-up fixes that make the lock effective: expose the heavy acquire/release control routes, match numbered rebind/multipart stages correctly, and cover connector catalog refresh/final effort selection.

Validation:
- `node --test launcher/tests/browser-host.test.cjs launcher/tests/control-server.test.cjs` — 111 pass, 0 fail.
- `bun test tests/browser-worker-contract.test.ts tests/launcher-browser-host.test.ts` — 135 pass, 0 fail.
- `bun x tsc --noEmit` passes.
- `git diff --check` passes.

### DIFF
--- a/launcher/electron/browser-host.cjs
+++ b/launcher/electron/browser-host.cjs
@@ -30,6 +30,10 @@ const IDLE_BROWSER_URL = "data:text/html;charset=utf-8,%3C!doctype%20html%3E%3Ch
 const PRIMARY_VIEW_BOOTSTRAP_TIMEOUT_MS = 10_000;
 const MAX_BROWSER_VIEW_DIMENSION = 16_384;
 const MAX_BROWSER_TABS = 5;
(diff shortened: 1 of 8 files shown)


miuuyy:
Thanks. A global scheduling change needs evidence that it addresses the actual stalls and preserves cancellation and independent-turn progress. Those conditions are not established here. Closing this implementation; the investigation is preserved in #381.

Issue #353, opened 2026-09-07
[Bug] Legacy remote compact 404s: non-chatgpt-web model passthrough to chatgpt.com /responses/compact which does not exist


wpfyorg:
### Codex Web GPT version

5.0.4 (launcher + daemon 5.0.4)

### Codex version

codex-cli 0.153.2 (spawns with `openai_base_url = http://127.0.0.1:17841/v1`)

### Platform

macOS arm64 (Apple Silicon)

### OS version

macOS 26.5.1

### ChatGPT account tier

Plus — bridge config reports `solAvailable: true, proAvailable: false`, `experimentalBiggerContext: true`

### Integration mode

Full harness, automatic browser host ("Codex Native2")

### Selected model

`chatgpt-web/high`

### What happened?

Compaction is hard-failing on this non-Pro account, and after the first failure the retry path produces a misleading 404. Chain of events (all visible in the attached log):

**Step 1 — chatgpt-web compaction attempt fails on non-Pro Bigger Context limits.** The session's structured context handoff needs more than the non-Pro staging budget. Two variants seen:

```text
ChatGptWebAdapterError: No ChatGPT effort available to this account can carry a Bigger Context stage with 80,169 estimated tokens and 276,680 characters.
  status: 400, errorType: "invalid_request_error", code: "context_length_exceeded"
```
(2026-09-06T15:55:57Z, again at 16:47 with 80,172 tokens. For a non-Pro account the effort selector only has `low` (32k token cap) and `medium` (80,000 token cap) — the largest indivisible stage part lands 169–172 tokens over the 80,000 cap.)

```text
ChatGptWebAdapterError: This Bigger Context transaction is estimated at 278,561 input tokens, which exceeds its experimental 270,000-token three-part ceiling. Run /compact, then retry.
```
(2026-09-07T04:11:21Z. 270,000 = the non-Pro `medium` 90,000-token context window × 3 under `experimentalBiggerContext`.)

This is the same failure class as #321 (Plus/non-Pro accounts).

**Step 2 — codex retries compaction with a different model.** codex-cli 0.153.2's compaction model fallback (`codex-rs/core/src/compact_model_fallback.rs`) retries when the error is `ContextWindowExceeded` / `UnexpectedStatus` / `InvalidRequest` / etc. The retry request carries a model slug that is **not** a `chatgpt-web/*` profile model.

[...]

Issue #346, opened 2026-09-06
[Bug] The message you submitted was too long, please edit it and resubmit.


ralphchristianeclipse:
### Codex Web GPT version

5.0.4

### Codex version

0.153.4

### Platform

Windows x64

### OS version

WIN 11 24H2

### ChatGPT account tier

Plus

### Integration mode

Full harness (MCP)

### Selected model

ChatGPT Web - High

### What happened?

<img width="564" height="132" alt="Image" src="https://github.com/user-attachments/assets/b55413ca-3079-44b4-aaf3-7ebdc6549a31" />

### Minimal reproduction

Just upgrade to v5.0.4

normal codex runs

<img width="315" height="108" alt="Image" src="https://github.com/user-attachments/assets/41534894-42cc-47c9-adef-1aeb348f0f73" />

### Complete final error

```text
The message you submitted was too long, please edit it and resubmit.
```

### Privacy-safe diagnostics

[codex-web-gpt-diagnostics-2026-09-06.json](https://github.com/user-attachments/files/31881933/codex-web-gpt-diagnostics-2026-09-06.json)

### Other Codex wrappers or route owners

none

### Preflight checklist

- [x] I tested the latest release and, for catalog or turn failures, fully restarted Codex including its background process.
- [x] I stated whether the failure also occurs in a fresh Codex task, or explained why that does not apply.
- [x] I attached only a privacy-safe export, or explained why the exporter could not produce one; I removed any remaining private content.
- [x] I searched existing issues for this exact final error.

hankok:
see the same issue multiple times

MohammedGhazal09:
This can be solved if you just make message sent in a .txt file @miuuyy

miuuyy:
I checked the attached safe log. It shows several successful multipart submissions, but it does not contain the exact “message you submitted was too long” alert, so I cannot yet tie the screenshot to a particular turn.

Could you attach a safe log exported immediately after this happens (Launcher → Activity → Export safe log), and give the failure time with your timezone? Please also say whether it happened on a fresh Codex task, while continuing an existing task, or during compaction, and whether the corresponding ChatGPT tab showed the same alert. A screenshot of that error area would help; please leave out private conversation text.

I have reproduced separate local partitioning/preflight problems and am fixing them, but I do not want to claim they explain your failure without matching the actual turn. No full conversation, cookies or credentials are needed.

ralphchristianeclipse:
[...]

PR #375, opened 2026-09-07
fix(chatgpt-web): guard browser message capacity before send

Ahmet1991:
ChatGPT browser page capacity is a per-message constraint, separate from the model context window. A turn can fit the model window but still freeze/fail when one browser message is too large.

This adds a pre-send capacity guard: keep inline transport when it fits, automatically use the existing 2/3-part transport when a normal Automatic turn exceeds the page limit, and fail explicitly with `browser_message_too_large` when no allowed transport fits. Manual/Zero Risk and Luna remain fail-closed instead of silently changing their transport contract.

While rebasing this onto current `main`, I fixed one compatibility regression in the original change: `experimentalBiggerContext=true` must still run through the existing Bigger Context policy, including Luna's explicit rejection, even when the inline message is below the page limit.

Validation:
- focused capacity + preparation regression run: 4 pass, 0 fail.
- `bun x tsc --noEmit` passes.
- `git diff --check` passes.
- A whole-file `chatgpt-web-harness.test.ts` run was started but did not exit on this Windows machine after the broker/named-pipe portion; I stopped it and am not counting it as a pass.

### DIFF
--- a/src/adapters/chatgpt-web/capacity.ts
+++ b/src/adapters/chatgpt-web/capacity.ts
@@ -1,0 +1,82 @@
+import type { CodexParsedRequest } from "../../types";
+import { ChatGptWebAdapterError } from "./adapter-error";
+import {
+  compiledChatGptWebMaxMessageChars,
+  DEFAULT_CHATGPT_WEB_MAX_MESSAGE_CHARS,
+} from "./input-tokens";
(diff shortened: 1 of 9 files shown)


miuuyy:
Thanks. This introduces a separate fixed page-capacity policy whose general limit is not established by the supplied evidence, and it changes transport selection beyond the current patch. We already have local message-budget and multipart fixes for #346 and #353, so I am closing this implementation and keeping the remaining capacity questions in those issues.

Issue #351, opened 2026-09-07
[Bug] stream disconnected before completion: ChatGPT stopped responding after the task started. Check the ChatGPT tab before continuing.


anlvdt:
### Codex Web GPT version

Version 5.0.4 (5.0.4)

### Codex version

Version 26.901.51231

### Platform

macOS arm64

### OS version

macOS 26.6.2

### ChatGPT account tier

Plus

### Integration mode

Full harness (MCP)

### Selected model

ChatGPT Web - High

### What happened?

After update to this Version 5.0.4 (5.0.4), i continusously got this error message, please help to fix. Thank you.

### Minimal reproduction

[Bug] stream disconnected before completion: ChatGPT stopped responding after the task started. Check the ChatGPT tab before continuing.

### Complete final error

```text
[Bug] stream disconnected before completion: ChatGPT stopped responding after the task started. Check the ChatGPT tab before continuing.
```

### Privacy-safe diagnostics

[Bug] stream disconnected before completion: ChatGPT stopped responding after the task started. Check the ChatGPT tab before continuing.

### Other Codex wrappers or route owners

[Bug] stream disconnected before completion: ChatGPT stopped responding after the task started. Check the ChatGPT tab before continuing.

### Preflight checklist

- [x] I tested the latest release and, for catalog or turn failures, fully restarted Codex including its background process.
- [x] I stated whether the failure also occurs in a fresh Codex task, or explained why that does not apply.
- [x] I attached only a privacy-safe export, or explained why the exporter could not produce one; I removed any remaining private content.
- [x] I searched existing issues for this exact final error.

miuuyy:
I've seen a similar “Thinking failed” state after heavy Pro usage. A temporary ChatGPT-side cooldown could be involved. Please pause requests and try again in about 30 minutes, with one fresh task.

Your report lists High, so we can't assume it is the same cause. If it persists, please attach the exact error shown inside the ChatGPT tab and a fresh **Activity → Export safe log** captured after the failure.

zxniuniu:
我也遇到的相同的问题。

miuuyy:
@zxniuniu Thanks for confirming. Could you add your launcher/Codex versions, OS, ChatGPT account tier and selected Web model, and attach a safe log from **Launcher → Activity → Export safe log** immediately after one failure? Please include the failure time and timezone.

[...]

PR #395, opened 2026-09-08
fix: a visible stop button disproves "Stopped thinking"

MaxKerkula:
## The failure

A turn is cancelled while ChatGPT is still generating:

```
18:16:20.371  broker completed call_p7mXWBrgHCks pending=0   <- last tool result
              ... 65s of heartbeats, no tool calls ...
18:17:25.910  failed: ChatGPT remained in 'Stopped thinking' for 5 seconds
```

The turn had been running six minutes and was driving tool calls throughout. The launcher sampled
`running=true` at three points before the cancellation, and the model returned another tool call
after each one.

## Cause

Both wait loops reached this decision **before** they had read the stop button, and never consulted
it — even though every other DOM-health verdict in the same loops takes `running` as an input.

Once the last tool result aged past `CHATGPT_RESPONSE_DOM_GRACE_MS` (60s),
`chatGptExternalProgressSuppressesDomHealth` went false, the window stopped being cleared, and
`CHATGPT_STOPPED_THINKING_GRACE_MS` (5s) later the guard threw. 60 + 5 against an observed 65.5.

So a model that pauses longer than a minute between tool calls is cancelled while a stale label
says it stopped.

## The change

Read the stop button before deciding, through one shared helper rather than the identical copy each
loop carried.

Proven MCP activity and a visible stop button are deliberately **not** treated as the same evidence:

| evidence | strength | effect |
|---|---|---|
| completed tool call | proves work happened | forgets the window outright |
| visible stop button | the renderer describing itself, which is what a wedged tab gets wrong | *suspends* under a ceiling |

The ceiling is not belt-and-braces, it is the only bound that exists: every other
`ChatGptTurnDomHealthTracker` verdict requires `!running`, `turnTimeoutMs` is undefined unless a
caller sets it, the main response loop runs outside any stage timeout, and the bridge's stall budget
is a silence timer this adapter refreshes with its own 10s heartbeat. Suppressing the guard without
a ceiling would trade a 5s false cancellation for a turn that never ends.

A failed stop-button read now reads as `undefined` rather than `false`. The call is wrapped in
`.catch()`, and treating a fault as "not running" re-armed the guard precisely when the page handle
was least trustworthy, such as during a rebind.

## Tests

`tests/browser-worker-contract.test.ts` replays the production timeline against the real constants
at the loop's real 250ms cadence, driving the real gate rather than asserting on constants:

| scenario | v5.0.4 | with this change |
|---|---|---|
| still generating, 65s | **cancels at 65,000 ms** | survives |
| still generating, 9 min | cancels | survives |
| wedged tab, 30 min | cancels at 65,000 ms | cancels at 660,000 ms |
| ChatGPT genuinely stopped | cancels at 65,000 ms | unchanged |
| stop button unreadable | cancels | suspended, same bound |

Each test was verified to fail against the specific line it guards.

`bun x tsc --noEmit` clean. `tests/browser-worker-contract.test.ts` passes. The full suite has a
large pre-existing failure set on the `v5.0.4` tag in my Windows environment; I captured it before
and after and diffed — identical, no new failures. `bun run verify` fails for that same
pre-existing reason, so I did not use it as evidence.

## Note on an existing test

`turn cancellation heuristics defer to proven MCP progress in both wait loops` is a source grep. Its
counts are updated for the shared helper. It cannot prove either loop passes the freshly read stop
button rather than a constant — I confirmed that by sabotaging the main loop with `running: false`
and getting a clean run — so the helper's behaviour is covered by unit tests instead, and the
comment no longer claims otherwise.

### DIFF
--- a/src/adapters/chatgpt-web/browser-worker.ts
+++ b/src/adapters/chatgpt-web/browser-worker.ts
@@ -1507,11 +1507,23 @@ export function chatGptExternalProgressSuppressesDomHealth(
     && age < CHATGPT_EXTERNAL_PROGRESS_STALL_CEILING_MS;
 }
 
+/**
+ * How long a still-generating ChatGPT may hold the turn without any proven progress.
+ *
+ * A visible stop button is the renderer describing itself, so it is weaker evidence than a
+ * completed tool call and cannot be trusted without end. This mirrors
+ * {@link CHATGPT_EXTERNAL_PROGRESS_STALL_CEILING_MS}, which exists because a call that never
+ * returns would otherwise hold a turn open forever; a stop button that never clears does the same.
+ * Past this bound the DOM decides again, so a wedged tab fails instead of hanging.
+ */
+export const CHATGPT_STOPPED_THINKING_RUNNING_CEILING_MS = 10 * 60_000;
+
 export class ChatGptStoppedThinkingTracker {
   private visibleSince?: number;
+  private suppressedSince?: number;
 
   /**
-   * Forgets an in-progress "Stopped thinking" window.
+   * Forgets an in-progress "Stopped thinking" window and the suppression holding it open.
    *
    * Suppressing only the throw let the window keep accruing while a tool call was outstanding, so
    * the first observation after progress ended cancelled the turn instantly. Proven activity must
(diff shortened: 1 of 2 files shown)


miuuyy:
Thanks for investigating this. We are closing this version because the ten-minute suppression still introduces false cancellations: after that window, the first appearance of the label can terminate a turn even while the Stop button remains visible. An unreadable Stop state also still becomes “not running” in the other completion checks.

“Stopped thinking” is not a dependable liveness signal. We have implemented a different local policy: an actual status in the current response is surfaced directly to Codex as a ChatGPT error, with no added waiting window or automatic resend. A ChatGPT Web usage limit may be one cause, but the label alone does not establish that. Quoted answer/reasoning text is excluded.

The broader response failures remain tracked in #351. Thanks for the useful investigation; a replacement PR is not reserved or guaranteed to be accepted while local work continues.

PR #396, opened 2026-09-08
fix: read ChatGPT's deferral notice before calling a turn stalled

MaxKerkula:
> Stacked on #395 — the first commit here is that PR's. Review or merge #395 first; the second
> commit is this change.

## The failure

A turn does 28 tool calls over 28 minutes, produces **no assistant message at all**, and ends with:

```
ChatGPT stopped generating but did not expose its completed-turn action; the ChatGPT DOM may have changed
```

Codex surfaces that as `stream disconnected before completion: ChatGPT stopped responding after the
task started`, discards the turn, and the user is left with collapsed tool cards and an error box —
never a reply.

## Cause

The DOM did not change. `button[data-testid="copy-turn-action-button"]` is intact; I probed a live
page and it appears exactly as expected on a healthy turn.

What is missing is any handling of ChatGPT's own deferral notice:

> Our systems are thinking a bit more about this request before responding. You can retry with a
> faster model for a quicker response, though it may be less capable of handling complex requests.

```
$ grep -ril "thinking a bit more" src/     ->  (no matches)
$ grep -ril "retry with a faster model" src/  ->  (no matches)
```

The adapter knows about the rate-limit dialog and `Something went wrong`, but not this. The notice
renders **outside** the assistant turn, so the response-DOM snapshot never sees it — the diagnostics
from a failing turn show a response element of only 12–26 characters. To every stall verdict in the
loop the turn is indistinguishable from a model that gave up, which is the exact opposite of what
ChatGPT is saying.

## The change

Read the notice before any stall verdict fires. While it shows, forget the stall windows, and say so
once through commentary so the wait is visible rather than silent. Past
`CHATGPT_DEFERRED_RESPONSE_CEILING_MS` the turn fails with what ChatGPT actually did, not with a
guess about the DOM.

The probe is deliberately lazy — it only runs on a turn that already looks stalled
(`!completionActionVisible && visibleText.length === 0`). The page is large (I measured 817,935
characters of body text on a real failing turn), so querying it every 250ms would be a real cost. A
turn producing text or showing its completion action is never deferred, so it never pays for this.

## Tests

- The deferral is bounded, forgotten as soon as ChatGPT starts producing, and the ceiling is finite —
  the same shape as every other liveness signal here.
- The error names the real condition; it no longer blames the DOM.
- A source assertion pins the short-circuit ordering: the probe must precede the stall verdict, and
  must sit behind the cheap guard conditions.

`bun x tsc --noEmit` clean, `tests/browser-worker-contract.test.ts` passes, full-suite failure set
identical to the `v5.0.4` baseline in my environment.

## What this does not fix

I could not confirm the notice was on screen during the specific 28-minute failure that prompted
this — the diagnostics do not capture text outside the response element, which is precisely why the
condition was invisible. So this is a real unhandled state with a plausible link to that failure,
not a proven root cause for it, and I would rather say so than overclaim.

### DIFF
--- a/src/adapters/chatgpt-web/browser-worker.ts
+++ b/src/adapters/chatgpt-web/browser-worker.ts
@@ -769,6 +769,61 @@ export async function throwIfChatGptTerminalErrorAlert(scope: ChatGptTextScope):
   );
 }
 
+/**
+ * ChatGPT's own notice that it has queued the request behind its normal latency.
+ *
+ * It renders outside the assistant turn, so the response-DOM snapshot never sees it: the turn just
+ * looks like a model that produced no text and exposed no completion action. Every stall verdict
+ * here means "ChatGPT stopped producing this turn", and this notice is ChatGPT stating the exact
+ * opposite, so it has to be read before any of them fire.
+ */
+const chatGptDeferredResponseNotice = (scope: ChatGptTextScope): Locator => scope
+  .getByText(/systems are thinking a bit more about this request/i)
+  .last();
+
+export async function chatGptDeferredResponseVisible(scope: ChatGptTextScope): Promise<boolean> {
+  return chatGptDeferredResponseNotice(scope).isVisible().catch(() => false);
+}
+
+export function chatGptDeferredResponseError(): ChatGptWebAdapterError {
+  return new ChatGptWebAdapterError(
+    "ChatGPT deferred this request (\"Our systems are thinking a bit more about this request before "
+    + "responding\") and never started answering. Retry, or pick a faster reasoning effort.",
+    { status: 504, errorType: "server_error", code: "chatgpt_response_deferred", retryable: true },
+  );
+}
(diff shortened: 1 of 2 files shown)


miuuyy:
Thanks for the investigation, but this implementation does not fit the failure it is intended to address. The new probe runs only when visible response text is empty, whereas the missing-completion-action error quoted here is emitted for a nonempty response. The page-wide text lookup also recognizes a user quoting the notice as if ChatGPT had displayed it.

The notice was not confirmed during the reported incident, and this PR also includes the suppression policy from #395 that we are not adopting. I am closing this version. The underlying stopped-response failures remain tracked in #351; the deferral notice is a hypothesis to verify against an actual failure, not an established cause.

Please attach that evidence to the issue before preparing a replacement. Local work may supersede another PR, so acceptance is not guaranteed.

PR #399, opened 2026-09-08
fix(chatgpt-web): normalize Windows namespaced rollout paths; require exactly one native environment_context claim

mhsalomar:
## Summary

Two Windows-only correctness fixes in the chatgpt-web adapter:

1. **Rollout path containment false-negative on Windows.** `pathIdentity()` lowercased the raw `resolve()`d path for comparison, but `fs.realpathSync` on Windows can return an extended-length (`\\?\C:\...`) or UNC (`\\?\UNC\server\share\...`) form. Comparing that verbatim against a normally-formed sessions root always failed containment, throwing `Codex rollout path escapes the sessions directory` for legitimate rollouts. Fix strips the `\\?\` / `\\?\UNC\` prefix before lowercasing; a path genuinely outside the sessions root is still rejected.
2. **Compaction continuation broke on an embedded or duplicated `<environment_context>` claim.** Both the "is this a native environment message" check and the continuation-claim extractor required the *entire* trimmed message to be exactly one `<environment_context>...</environment_context>` block (anchored `^...$`). A claim embedded in a larger message (e.g. with surrounding project context) was invisible to matching, and a message containing two concatenated claims collapsed to a single anchor-mismatch instead of the intended "zero or multiple claims" rejection. Fix matches the tag as a substring/`g`-extracted list instead of the whole string; the "exactly one" invariant (`Compaction continuation requires one current native environment claim`) is unchanged and still enforced.

## Root cause

Both bugs share a root cause: environment-claim/path recognition assumed a single canonical string form (case, delimiter, prefix) that Windows APIs and larger message payloads don't guarantee. Neither fix loosens the actual security/authentication invariant (sessions-directory containment; exactly-one-claim authentication) — both only fix the string normalization/matching around it.

## Tests

`tests/environment.test.ts` adds:
- A `<environment_context>` claim embedded inside a larger message resolves normally.
- Two concatenated claims in one message throws `Compaction continuation requires one current native environment claim` (previously this exact input hit a different, ambiguous failure path).
- (Windows-only) A `\\?\`-namespaced rollout path resolves against the sessions root correctly.

`bun test tests/environment.test.ts`: 49/49 pass. `bun run typecheck`: clean.

### DIFF
--- a/src/adapters/chatgpt-web/codex-rollout-environment.ts
+++ b/src/adapters/chatgpt-web/codex-rollout-environment.ts
@@ -42,7 +42,13 @@ function record(value: unknown): Record<string, unknown> | undefined {
 
 function pathIdentity(value: string): string {
   const normalized = resolve(value);
-  return process.platform === "win32" ? normalized.toLowerCase() : normalized;
+  if (process.platform !== "win32") return normalized;
+  const comparable = normalized.startsWith("\\\\?\\UNC\\")
+    ? `\\\\${normalized.slice(8)}`
+    : normalized.startsWith("\\\\?\\")
+      ? normalized.slice(4)
+      : normalized;
+  return comparable.toLowerCase();
 }
 
 function contains(root: string, path: string): boolean {
--- a/src/adapters/chatgpt-web/environment.ts
+++ b/src/adapters/chatgpt-web/environment.ts
@@ -134,7 +134,7 @@ export function hasCurrentChatGptEnvironmentContext(parsed: CodexParsedRequest):
 
 function contextualUserMessage(value: Record<string, unknown>): boolean {
   const text = rawMessageText(value).trim();
-  return /^<environment_context>[\s\S]*<\/environment_context>$/.test(text)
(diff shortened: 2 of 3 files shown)


mhsalomar:
Closing as superseded by v5.0.5 (0b053b6).

- Windows rollout-path fix: v5.0.5 already fixes this via `toNamespacedPath()` in `pathIdentity()` (different implementation, same effect). Confirmed by reading the shipped v5.0.5 diff.
- `Compaction continuation requires one current native environment claim`: verified with a dedicated regression test (dedicated environment_context content part, grouped-sibling-parts preamble, zero claims, multiple claims, plain-string content) against a clean v5.0.5 checkout with no patch applied -- all 6 cases pass. v5.0.5's per-content-part anchored matching already handles the real Codex Desktop wire shape (environment_context always arrives as its own content part, never merged with preamble text in the same part).

No code change needed on top of v5.0.5 for either issue this PR was opened for. Thanks for shipping v5.0.5.

PR #403, opened 2026-09-08
fix(browser): resolve Playwright launchPersistentContext attach timeout on Windows

Wladefant:
## Summary
Resolves the issue where `bun run src/cli.ts login` opens a second visible Chrome window and times out with `launchPersistentContext: Timeout 180000ms exceeded`, accompanied by an unsupported `--no-sandbox` warning banner.

### Root Causes
1. **Windows Bun stdio Pipe Incompatibility**: When running under Bun on Windows, `child_process.spawn` does not inherit extra pipe file descriptors (`stdio[3]` and `stdio[4]`) to non-Bun Windows executables (`chrome.exe`). Playwright passes `--remote-debugging-pipe`, but Chrome never receives the pipe handles from Bun, so Playwright hangs on connection.
2. **Process Singleton Re-use / Collision**: When the manual Chrome window closes, lingering child processes (Crashpad handler, GPU process, or graceful shutdown) still hold the `user-data-dir` singleton lock. Calling `launchPersistentContext` on the same profile path causes Chrome to delegate to the lingering instance and exit with code 0.
3. **Unsupported `--no-sandbox` Flag**: Passed by default on Chromium, triggering Chrome's unsupported flag warning banner on Windows.

### Fix
- Terminate any lingering Chrome processes associated with the profile directory and remove stale lock files (`SingletonLock`, `SingletonCookie`, `SingletonSocket`, `lockfile`).
- Create an isolated temporary copy of the profile for persistent context extraction, ensuring Chrome's process singleton mutex never collides.
- Drop `--no-sandbox` via `ignoreDefaultArgs` on Windows.
- On Windows under Bun, delegate Playwright persistent context extraction and capability verification to a dedicated Node helper (`src/browser-playwright-worker.cjs`), running headlessly with `--disable-blink-features=AutomationControlled` and headed fallback.
- Streamline operator workflow: operator only sees and closes one normal window, subsequent extraction and verification complete headlessly in ~3 seconds.

### DIFF
--- a/src/adapters/chatgpt-web/browser-helper-main.ts
+++ b/src/adapters/chatgpt-web/browser-helper-main.ts
@@ -13,7 +13,11 @@ interface RunMessage {
   id: string;
   config: {
     appName: string;
-    browserHostDescriptorPath: string;
+    browserHost?: "launcher" | "managed-chrome";
+    browserHostDescriptorPath?: string;
+    chromeExecutablePath?: string;
+    storageStatePath?: string;
+    headed?: boolean;
     browserDiagnosticsPath?: string;
     turnTimeoutMs: number;
     autoApproveToolCalls: boolean;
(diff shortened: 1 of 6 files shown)


miuuyy:
Thanks for investigating the Windows login timeout. Closing this implementation because it broadens process/profile handling, weakens login verification, and leaves helper-packaging gaps. Those changes are not justified by a verified current failure.

The known Bun extra-pipe write bug matching this symptom was fixed upstream and the fix is included in our pinned Bun 1.4.0. The ordinary launcher also uses a separate Electron/Node browser path. This does not prove that your standalone CLI case is resolved; the PR does not establish a reproduction on the pinned runtime.

The report is preserved in https://github.com/miuuyy/codex-chatgpt-web/issues/405 for that specific verification. We will handle a confirmed remaining cause locally without taking the broader changes here. A replacement PR is not guaranteed to be accepted, especially if local work has already started. Thank you for the report.

Issue #276, opened 2026-09-01
Full Harness: raw exec can bypass wait_agent polling contract and expire turn token


ventianima-lab:
## Summary

In Full Harness mode, a ChatGPT Web turn can bypass the advertised transport-safe `wait_agent` schema by calling `wait_agent` inside the freeform `exec` gateway. A long nested wait then exceeds the MCP invocation deadline, the binding is retired as designed, and a later tool claim fails with:

`turn token is invalid or expired`

This is distinct from #274: the browser turn had been accepted and remained in its native tool loop. The failure was triggered by an overlong nested native call, not by browser completion racing a delayed initial claim.

## Environment

- codex-chatgpt-web: v4.0.7
- Windows 11
- Full Harness mode
- `cgw/chatgpt-web-pro`

## Reproduction

1. Start a tool-capable ChatGPT Web turn.
2. From ChatGPT, invoke the top-level freeform `exec` tool.
3. Inside the JavaScript input, call a nested agent wait with a long timeout, for example:

```js
const result = await tools.multi_agent_v1__wait_agent({
  targets: [{ agent_id: "..." }],
  timeout_ms: 180000,
});
text(result);
```

4. The outer MCP request exceeds the 90-second transport deadline. Its binding is retired to prevent an abandoned native mutation from overlapping later work.
5. A later claim using the same turn token fails with `turn token is invalid or expired`.

## Expected behavior

The freeform `exec` gateway must not bypass the same transport contract advertised by the direct `multi_agent_v1__wait_agent` / `multi_agent_v2__wait_agent` tools. The safe options are:

- reject nested `wait_agent` calls in `exec` and direct the model to use the dedicated tool, or
- enforce the exact 10-second polling contract before invoking the native harness.

The binding should continue to be retired after a real MCP deadline; extending token lifetime or leaving a timed-out binding active would risk overlapping an abandoned command with later work.

## Local regression proof

A local guard that rejects nested `wait_agent` calls in the freeform `exec` gateway passes the existing outer-native MCP bridge test. The same token remains usable for a subsequent direct `wait_agent` call using `timeout_ms=10000`. TypeScript type-check also passes.

No user prompt is included here, and the affected prompt was not resent.

miuuyy:
Fixed in [v4.0.8](https://github.com/miuuyy/codex-chatgpt-web/releases/tag/v4.0.8).

[...]

PR #404, opened 2026-09-08
fix: bound collaboration agent waits

ogyrec-o:
## What this changes

Adds collaboration__wait_agent to the existing 10 second wait guard used by the older multi-agent wait tools.

The same guard now applies to both direct calls and the raw exec gateway. It also drops any inherited timeout_ms default before exposing the schema, so the contract stays fixed at 10000 ms.

Related: #276

## Evidence

Added regression coverage for direct and raw calls. A 180000 ms wait is rejected before the outer tool runs, while a 10000 ms wait is forwarded. The exposed schema requires 10000 ms and has no default.

Focused harness test: 81 passed, 0 failed.

bun run verify passes.

Also tested with Codex 0.153.4 against the packaged 5.0.5 runtime built from this commit. A spawned collaboration child completed after repeated 10 second polls, and the parent could still run a tool afterward.

## Scope and invariants

- [x] I read and followed CONTRIBUTING.md.

## Verification

- [x] I ran bun install --frozen-lockfile in the repository root and launcher/.
- [x] I ran bun run verify with the Bun version pinned by package.json.

macOS arm64
Plus
Full harness / MCP
Codex CLI 0.153.4
Packaged 5.0.5 runtime

Browser-only: not run
Windows: not run
Linux: not run
Other account tiers: not run

### DIFF
--- a/src/adapters/chatgpt-web/mcp-server.ts
+++ b/src/adapters/chatgpt-web/mcp-server.ts
@@ -27,9 +27,10 @@ const BRIDGE_TOOL_NAMES = new Set([
   "codex_turn_complete",
 ]);
 
-const GATEWAY_AGENT_WAIT_TOOL_NAMES = new Set([
+const AGENT_WAIT_TOOL_WIRE_NAMES = new Set([
   "multi_agent_v1__wait_agent",
   "multi_agent_v2__wait_agent",
+  "collaboration__wait_agent",
 ]);
 
 const turnTokenSchema = z.string().min(20).max(256);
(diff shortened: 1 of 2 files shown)


miuuyy:
Thanks for identifying the missing collaboration wait guard. Closing this PR in favor of the local implementation we have now prepared as part of our existing transport work (#276).

The submitted regression uses the legacy target-list contract, but native `collaboration.wait_agent` accepts a timeout and waits for mailbox activity; the installed Codex rejects that target-list argument. The inherited target-based guidance also needs to match the actual native tool.

Our local version covers direct calls and both exec paths, preserves the native arguments/result, and distinguishes a wait timeout from agent completion. It has passed the full verification suite and a real installed-Codex/MCP check, but is not released yet. Thank you for flagging the gap. A replacement PR is not guaranteed to be accepted because this work is already being handled locally.

ogyrec-o:
@miuuyy Thanks for the clarification and for checking this against the current native collaboration contract.

Issue #342, opened 2026-09-06
[Bug] Frequent codex_exec safety-status rejections block development and verification


MohammedGhazal09:
## Problem

Frequent safety-classification rejections interrupt ordinary authorized development and verification through ChatGPT Web's Codex Native tools. The user reports repeated occurrences across coding continuations, especially repository inspection and test-related work. The latest report includes both a `codex_exec` listener-inspection rejection and a separate browser-review rejection in the same continuation.

This is a bug report about tool reliability and failure handling. The rejection rate has not been measured, and the underlying cause has not been localized to OpenAI classification, connector metadata, or bridge request construction.

## Exact error

Both actions reportedly returned:

```text
This tool call was blocked by OpenAI because we couldn't determine the safety status of the request.
```

## Latest observed impact

The user supplied the Web model's end-of-turn report:

- `codex_exec` rejected a local listener-inspection command. Shell execution then stopped for that response.
- An independent browser-review action through `codex_tool_call` → `mcp__cua_repl__js` was also rejected.
- Before those blockers, the model reported 24 focused tests passing across six files and changes to browser assertions for mobile visibility and action counts.
- The changed browser tests, final diff, typecheck and lint remained unverified. The multi-phase GSD pipeline stopped with the current plan still in progress.

These are user-supplied model reports, not independently replayed tool traces. The exact rejected command and browser arguments are absent from the supplied report and must not be reconstructed. Earlier successful tests do not establish verification of the later edits.

## Reproduction workflow

This is the reported workflow; a minimal deterministic reproduction is still needed:

1. Run a coding task through the bridge in full-harness/MCP mode, with an authorized local repository and native tools.
2. Perform source work and focused verification.
3. Request a local listener inspection with `codex_exec` before rendered UI verification.
4. Observe the safety-status rejection above. A separate browser-review action was also reported rejected in this continuation.
5. Observe the remaining verification and pipeline work left pending.

The user has reported recurrence across new sessions/continuations. The exact selected Web mode, binary versions and original arguments for this latest occurrence have not been captured, so this is not claimed as a release-specific or Pro-only regression.

## Expected behavior

Authorized benign inspection and verification should execute when policy permits. If classification cannot establish safety, the outcome should be diagnosable with the exact affected action and a privacy-safe failure class. Independent permitted work should continue when its prerequisites are satisfied. Actual refusals must remain respected.

## Known bridge contribution

[...]

PR #430, opened 2026-09-10
fix: expose read-only Codex task access

oenderg:
## What this changes

A read-only `mcp__codex_app__read_thread` request currently has to pass through the generic `codex_tool_call` action, whose public MCP annotations correctly describe arbitrary calls as destructive and open-world. ChatGPT can therefore reject a harmless task read before it reaches the bridge.

This adds a dedicated, turn-bound `codex_read_thread` action. It forwards only to the exact advertised `mcp__codex_app__read_thread` tool, preserves the native argument surface, reports an explicit error when that tool is unavailable, and advertises read-only, non-destructive, idempotent, closed-world annotations. The action is available through both the automatic Full harness and Zero Risk contracts.

Related to #342.

## Evidence

Before this change, ChatGPT Web High rejected both the generic task-read request and a minimal retry with only the task ID. The rejection happened before connector dispatch.

After installing the patched runtime and explicitly refreshing the existing `Codex Native2` connector, ChatGPT exposed `codex_read_thread`. A real installed Codex task successfully used it to retrieve one turn from a referenced Codex task with `includeOutputs: false`; the request reached the outer `read_thread` tool and returned its structured result.

The connector ABI hash changes because this is an additive public action. The existing ABI test documents explicit connector refresh or identity migration as the two supported responses; this patch was accepted end-to-end after an explicit refresh. It intentionally does not add a broader connector-identity migration.

## Scope and invariants

## Verification

- [ ] I ran `bun run verify` with the Bun version pinned by `package.json`. It currently stops at the unchanged lockfile audits: three existing Hono advisories in the root and one existing `js-yaml` advisory in `launcher/`. No dependency or lockfile changes are included. Every later verification stage was run individually and passed.
- [x] If this changes ChatGPT browser UI handling, I included observed DOM evidence and a reproducible fixture instead of broadening selectors speculatively. N/A: no browser UI handling changed.
- [x] If this changes the launcher, I preserved macOS, Windows, and Linux packaging and named the platform packages actually built below. N/A: no launcher code changed.

Additional results:

- `bun run typecheck`
- `bun run test`: 693 passed, 1 platform-specific test skipped, 0 failed
- `bun run launcher:typecheck`
- `bun run launcher:test`: 295 passed, 1 skipped, 0 failed
- `bun run launcher:build`
- runtime bundle build, third-party notice generation, and relocatable runtime smoke test
- focused Full/Zero Risk MCP tests: 88 passed, 0 failed
- `git diff --check`

- macOS 26 Tahoe, Apple silicon
- ChatGPT Plus
- Automatic Full harness with ChatGPT Web High: passed after explicitly refreshing `Codex Native2`
- Zero Risk: unit-tested request-ID routing; live account flow not run
- Browser-only: not affected
- Windows and Linux: not run
- Packaged platform installers: not built; no launcher or packaging code changed

### DIFF
--- a/src/adapters/chatgpt-web/mcp-server.ts
+++ b/src/adapters/chatgpt-web/mcp-server.ts
@@ -22,6 +22,7 @@ const BRIDGE_TOOL_NAMES = new Set([
   "codex_write_stdin",
   "codex_apply_patch",
   "codex_view_image",
+  "codex_read_thread",
   "codex_tool_inventory",
   "codex_tool_call",
   "codex_turn_complete",
@@ -116,6 +117,10 @@ function wireName(tool: CodexTool): string {
 
 function exactTool(environment: ChatGptTurnEnvironment, name: string): CodexTool | undefined {
   return environment.tools.find(tool => !tool.namespace && tool.name === name);
+}
+
+function exactCodexAppTool(environment: ChatGptTurnEnvironment, name: string): CodexTool | undefined {
+  return environment.tools.find(tool => tool.namespace === "mcp__codex_app" && tool.name === name);
 }
 
 function gatewayToolNameIsValid(name: string): boolean {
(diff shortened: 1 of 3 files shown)


miuuyy:
Thanks for the installed test, but a dedicated MCP wrapper for this individual native action does not fit the product direction. I want to diagnose the underlying refusals in #342 rather than add separate public actions around them. I'm closing this proposal.

PR #432, opened 2026-09-11
fix: include Pro usage-limit retry date in the adapter error

gmoroz:
## What this changes

When ChatGPT has hidden Pro behind its weekly usage cap, selecting Pro currently fails with only:

`ChatGPT model controls are unavailable. Reload ChatGPT and retry the task.`

The picker still shows a `Pro` row. Hovering it mounts a portal with the retry date, for example:

`Limit reached. Try again after Sep 15, 2026.`

This change Playwright-hovers that `Pro` menuitemradio and appends the visible retry text to the client-visible adapter error. Extra High is unchanged. Missing tooltip keeps the existing error with no fabricated date.

## Evidence

**Before.** Requested effort `max` while the slider range is `0..3`. Error message is only `ChatGPT model controls are unavailable. Reload ChatGPT and retry the task.` The date is not in `document.body.innerText` or the accessibility tree until hover.

**Observed DOM.** Intelligence picker `role="menuitemradio"` whose accessible text is exactly `Pro`. Hover mounts a portal whose visible text matches `Try again after Mon DD, YYYY`. Screenshot of the live control: tooltip `Limit reached. Try again after Sep 15, 2026.` next to the `Pro` row while Extra High is selected (`4 of 4`).

**After.** Same failure path throws that tooltip text in `ChatGptWebAdapterError.message`. Fixture: hover the Pro row, wait for the visible retry date, expect `rejects.toThrow("Limit reached. Try again after Sep 15, 2026.")`.

## Scope and invariants

## Verification

- [x] I ran `bun run verify` with the Bun version pinned by `package.json`.

`bun run verify` (Bun 1.4.0): `check-version` OK; `bun audit` fails on existing `hono@4.12.34` moderate advisories (not this diff).

- `bun run typecheck`: pass
- `bun test tests/chatgpt-session.test.ts`: 15 pass
- `bun test ./tests`: **694 pass, 1 skip, 3 fail** (698 total). Failures: two Zero Risk compaction tests and `explicit setup still refuses changed hooks` — same class of environment/launcher failures reported on #402, not this change.
- `git diff --check`: pass
- Red/green: the hover→tooltip test fails without `userVisibleDetail` and passes after the source change.

- Linux, ChatGPT Pro account, observed the live Pro tooltip on 2026-09-11. Did not replay the installed helper turn after this patch: the date is gone from the DOM until hover, and a live Pro click would fight an in-use Extra High session. Fixture covers the error path.
- Browser-only / Full / packaged launcher builds: not run.

### DIFF
--- a/src/adapters/chatgpt-web/browser-worker.ts
+++ b/src/adapters/chatgpt-web/browser-worker.ts
@@ -193,9 +193,9 @@ function chatGptModelControlUnavailableError(diagnostic: string): Error {
   return new Error(CHATGPT_MODEL_CONTROL_UNAVAILABLE_MESSAGE, { cause: new Error(diagnostic) });
 }
 
-function chatGptModelControlUnavailableAdapterError(diagnostic: string): ChatGptWebAdapterError {
+function chatGptModelControlUnavailableAdapterError(diagnostic: string, userVisibleDetail?: string): ChatGptWebAdapterError {
   return new ChatGptWebAdapterError(
-    CHATGPT_MODEL_CONTROL_UNAVAILABLE_MESSAGE,
+    userVisibleDetail ? `${CHATGPT_MODEL_CONTROL_UNAVAILABLE_MESSAGE} ${userVisibleDetail}` : CHATGPT_MODEL_CONTROL_UNAVAILABLE_MESSAGE,
     {
       status: 502,
       errorType: "server_error",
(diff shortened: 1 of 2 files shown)


miuuyy:
Thanks for the tooltip observation. Reviewing and adapting this separate implementation would cost more time and model quota than keeping a focused version in the local patch with support for other languages. I've added that version: it reads only the tooltip linked to the disabled Pro option and preserves ChatGPT's localized wording. Closing in favor of the local change tracked in #500; no reset date is guessed.

PR #435, opened 2026-09-11
fix: use existing hooks.json for the Codex Interrupt hook

aolin480:
## What this changes

The bridge now uses an existing `$CODEX_HOME/hooks.json` for its Interrupt hook. This avoids the mixed hook-source warning caused by adding a hook to `config.toml` when JSON hooks are already present. If `hooks.json` is absent, setup keeps the existing TOML behavior and does not create the JSON file.

Version 11 journals track JSON hook ownership and TOML trust state. Setup migrates bridge-owned v10 TOML hooks into an existing JSON file, preserves user hooks and symlinks, and restores owned changes on disconnect or uninstall. Recovery handles interrupted JSON writes during first install and runtime-command upgrades. Launcher checkpoints also include the JSON file.

This PR remains a draft pending maintainer review and dependency audit resolution.

## Evidence

- Isolated Codex discovery with JSON hooks and TOML trust state produces no mixed-source warning on Codex 0.154.0 and 0.153.4.
- The real Codex interruption smoke passes with JSON storage selected.
- The built CLI reconnects simulated v10 installations into v11 with both TOML-only and existing-JSON configurations. Disconnect restores the original configuration in both cases.
- Regression tests cover user-hook preservation, legacy migration, symlinks, path tampering, setup after disconnect, and interrupted writes.

## Scope and invariants

- [x] I read `CONTRIBUTING.md`.
- [x] The change is limited to hook storage, lifecycle ownership, rollback, and regression evidence.
- [x] Model, route, effort, connector, and capability selection retain their existing behavior.
- [x] Browser-only gains no broker or connector.
- [x] No dependency updates, generated artifacts, version changes, credentials, or private paths are included.

## Verification

The following checks used the pinned Bun 1.4.0 on macOS arm64.

- [x] `bun install --frozen-lockfile` in the repository root and `launcher/`.
- [x] `bun run check-version` and `bun run typecheck`.
- [x] `bun run test`: 706 pass, one Windows-specific skip.
- [x] `bun run launcher:typecheck` and `bun run launcher:test`: 295 pass, one skip.
- [x] `bun run launcher:build`.
- [x] `bun run build` and `bun run smoke`: `RELOCATABLE_RUNTIME_SMOKE_OK`.
- [x] `bun run smoke:interrupt`: `NATIVE_CODEX_INTERRUPT_LIFECYCLE_SMOKE_OK`.
- [x] `bun run scripts/smoke-codex-hooks-json.ts <codex-executable>` against Codex 0.154.0 and 0.153.4.
- [x] Packaged CLI v10-to-v11 reconnect/disconnect simulation for both storage modes.
- [ ] `bun run verify` completes successfully. Existing dependency audit findings block it; the remaining checks were run separately.

The root audit reports three moderate advisories against the unchanged Hono 4.12.34 pin. The launcher audit reports one high advisory against unchanged js-yaml 4.3.1. Both remain unresolved and outside this change.

The relocatable macOS arm64 runtime and launcher renderer were built and exercised. Codex smoke tests used an isolated Codex home and a local fixture server.

Not run: native macOS installer packaging, Windows/Linux packages, or authenticated ChatGPT account tests. Account-tier and authenticated Browser-only/Full coverage remain unverified.

### DIFF
--- a/launcher/electron/runtime.cjs
+++ b/launcher/electron/runtime.cjs
@@ -475,6 +475,7 @@ class RuntimeHost {
       path.join(coreHome, "codex", "integration-journal.json"),
       path.join(coreHome, "codex", "integration-journal.recovery.json"),
       path.join(this.codexHome, "config.toml"),
+      path.join(this.codexHome, "hooks.json"),
       path.join(this.codexHome, "models_cache.json"),
       path.join(coreHome, "secrets", "tunnel-runtime.key"),
       path.join(coreHome, "secrets", "tunnel-runtime-automatic.key"),
@@ -506,7 +507,8 @@ class RuntimeHost {
       }
     }
     return [...paths].map(filePath => captureRegularFile(filePath, {
-      followSymlink: filePath === path.join(this.codexHome, "config.toml"),
+      followSymlink: filePath === path.join(this.codexHome, "config.toml")
+        || filePath === path.join(this.codexHome, "hooks.json"),
     }));
   }
 
--- a/launcher/tests/runtime-host.test.cjs
+++ b/launcher/tests/runtime-host.test.cjs
@@ -785,9 +785,13 @@ test("failed first-time setup removes its route before restoring the unconfigure
   const recoveryJournalPath = path.join(coreHome, "codex", "integration-journal.recovery.json");
   const configPath = path.join(root, "config.json");
   const codexConfigPath = path.join(codexHome, "config.toml");
+  const codexHooksPath = path.join(codexHome, "hooks.json");
+  const sharedHooksPath = path.join(root, "shared-hooks.json");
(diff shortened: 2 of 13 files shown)


miuuyy:
Thanks for the detailed work and validation. I am declining this implementation: removing the mixed hook-source warning brings in a much broader change to hook storage, journal migration, trust state, recovery and uninstall across 13 files. The reported hooks already work, so that scope and migration risk do not fit the current bug-fix patch.

I am keeping the narrower TOML ownership fixes in my local patch separate from this proposal. Your later confirmation that native Agent v2 is already supported is also noted; that does not require this migration. I may revisit storage consistency separately, but please check with me before rebuilding the PR, since the local implementation is already diverging.

PR #442, opened 2026-09-11
Scrub all six broker handle kinds, escape-aware and length-exact

zm2231:
## What this fixes

The retired-handle scrubber in `prompt.ts` was letting broker handles through in three ways, and any one of them puts a live `call_`/`control_` token back into the context we forward to ChatGPT web.

Upstream `main`:

```ts
const RETIRED_TURN_HANDLE = /\b(turn|request|binding)_[A-Za-z0-9_-]{24,}/g;
```

Three gaps:

1. **Missing kinds.** The broker mints six handle kinds. This regex only knows `turn|request|binding`, so `call_`, `control_`, and `handoff_` handles are never scrubbed and ride straight into the outbound context.
2. **`\b` breaks on escaped prefixes.** In serialized context JSON a handle often sits right behind a JSON escape. A `\b` word boundary does not fire between the escape character and the letter, so those handles leak:

   ```
   "\ncall_0123456789abcdefghijklmnopqrstuv"
   "control_0123456789abcdef"
   ```
3. **`{24,}` is looser than the real ID.** Broker IDs are exactly 32 body chars (24 base64url bytes for turn/binding/call/request, 16 hex for control/handoff). The open-ended `{24,}` can chew into adjacent data.

## The change

```ts
const RETIRED_HANDLE_KIND = "turn|binding|call|request|control|handoff";
const RETIRED_TURN_HANDLE = new RegExp(
  `(?:(?<![A-Za-z0-9_-])|(?<=\\[bfnrt])|(?<=\\u00[01][0-9a-fA-F]))(${RETIRED_HANDLE_KIND})_[A-Za-z0-9_-]{32}(?![A-Za-z0-9_-])`,
  "g",
);
```

All six kinds, exact 32-char body, and a left boundary that accepts a real token gap, a JSON single-char escape (`\n`, `\t`), or a `\u00XX` control escape. The `withoutRetiredTurnHandles` signature and its three call sites are untouched, so the per-string scrub behavior is identical for the handles it already caught.

## Why regex-only

Keeping it inside the existing per-string pass means no new allocation, no structural walk of the context tree, and the diff stays reviewable at a glance. The scrubber was already the right layer. It was just under-specified.

## Verification

- `bunx tsc --noEmit`: exit 0.
- `bun test`: 989 pass, 2 skip, 0 fail. Added a regression case that scrubs every handle kind and preserves the near-misses (a 31-char and a 33-char body, an unescaped-prefix false neighbor). Removing any escape branch or the `call` alternative now fails that test, so it has teeth.
- ReDoS check: 1M-char adversarial input scans in 0.13ms. No backtracking blowup from the alternation.

## Scope

Two files, +25/-1. `src/adapters/chatgpt-web/prompt.ts` and `tests/prompt-contract.test.ts`. Nothing else moves.

### DIFF
--- a/src/adapters/chatgpt-web/prompt.ts
+++ b/src/adapters/chatgpt-web/prompt.ts
@@ -143,7 +143,11 @@ export function formatChatGptWebMultipartCommit(
   ].join("\n");
 }
 
-const RETIRED_TURN_HANDLE = /\b(turn|request|binding)_[A-Za-z0-9_-]{24,}/g;
+const RETIRED_HANDLE_KIND = "turn|binding|call|request|control|handoff";
+const RETIRED_TURN_HANDLE = new RegExp(
+  `(?:(?<![A-Za-z0-9_-])|(?<=\\\\[bfnrt])|(?<=\\\\u00[01][0-9a-fA-F]))(${RETIRED_HANDLE_KIND})_[A-Za-z0-9_-]{32}(?![A-Za-z0-9_-])`,
+  "g",
+);
 
 /**
  * The accumulated Codex context replays earlier turns, including the broker handles those turns
--- a/tests/prompt-contract.test.ts
+++ b/tests/prompt-contract.test.ts
@@ -7,6 +7,7 @@ import {
   compileChatGptWebPrompt,
   formatChatGptWebMultipartCommit,
   formatChatGptWebMultipartStage,
+  withoutRetiredTurnHandles,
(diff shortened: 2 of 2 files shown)


miuuyy:
Thanks. I reproduced the narrower escaped-newline cleanup defect and fixed it locally using decoded JSON string values. I'm keeping native tool-call IDs intact instead of extending the generic scrubber to every handle-looking identifier. Closing this implementation; the verified scope and limitations are tracked in #501.

PR #459, opened 2026-09-12
Add support for up to ten concurrent agent-controlled ChatGPT browser tabs

derekszen:
## What this changes

Adds focused support for up to ten concurrent agent-controlled ChatGPT Web tabs.

- Keeps automatic agent tabs on independent leases, trace IDs, surfaces, and helper ownership.
- Adds a user-managed New chat tab without replacing the selected tab.
- Makes tab sizing responsive when many tabs are visible.
- Documents that all tabs share the same authenticated ChatGPT account, cookies, site storage, account limits, and profile settings.
- Records multi-account support as separate future work requiring distinct authenticated browser partitions.

Fixes #

## Evidence

Before: the launcher and worker/session capacity were limited to five browser tabs/turns, and opening a new user tab could replace the selected tab.

After: launcher and worker limits are ten; user-created tabs preserve the selected tab; the ten-turn and eleventh-rejection paths have focused regression coverage.

## Scope and invariants

- [x] Small, focused change with no unrelated cleanup or generated rewrite.
- [x] Same authenticated ChatGPT account remains the explicit behavior; this is not multi-account support or a quota bypass.
- [x] Automatic turns retain independent ownership and fail-closed capacity handling.

## Verification

- [x] Focused browser-worker and turn-broker tests: 138 passed, 0 failed.
- [x] Launcher tests and build passed before the final test-fixture-only commit.
- [ ] Full `bun run verify`: not run.
- [ ] Manual live multi-agent account test: not run.
- [ ] Packaged AppImage validation: not run.
- [x] No browser state, credentials, tunnel IDs, raw logs, or private paths committed.

Not run. The implementation uses the existing authenticated launcher partition and has not been installed into a packaged runtime.


miuuyy:
Thanks, but I'm keeping the five-tab ceiling. Raising it to ten increases parallel account traffic and may increase cooldowns and instability; the current tests do not establish that this is a safe default. I'm closing the PR. The measurement-first concurrency proposal remains in #382.

PR #460, opened 2026-09-12
feat: upload large ChatGPT Web context as TXT and rebind answers

jamie950315:
## Summary
- Bigger Context sends large automatic Instant, Medium, High, Extra High, and Pro tasks as one complete TXT attachment. Small tasks stay inline; manual Zero Risk and legacy Luna behavior is unchanged.
- The TXT contains the canonical task context, preserves message roles, and excludes the current broker capability. No additional filesystem scan or client-side model switch is introduced.
- File-based native turns start a fresh web conversation with complete history so local tools remain available. Tool-result rounds within a turn share the same active runtime.
- Count file contents, instructions, image reserves, and platform reserve. Enforce the selected budget and a separate per-mode ceiling; reject excess size or upload failure without silently trimming the file. One TXT leaves nine image slots.
- Negotiate helper support and validate the attachment name, hash, envelope, and size. Retain the latest-answer rebinding fix and bounded packaging/process-cleanup checks.
- Core plan-aware budget plumbing is included here; #461 exposes the launcher and CLI plan selector.

## Validation and limits
- All four updated stack heads passed `bun run verify` locally.
- The clean PR stack passed a real Medium web TXT test at approximately 309,747 estimated total input tokens: 24/24 distributed markers recovered.
- The TXT implementation also passed Instant file recall near 120k total tokens and Medium/High/Extra High near 310k, plus two-turn local-tool tests on Instant, Medium, and Pro.
- These are bounded file-assisted results, not guarantees of full internal context retention. ChatGPT may process files with tools or Thinking steps even when the client selects Instant. Actual Plus-account behavior and live native compaction remain unverified.
- No native Codex inference or separately billed API fallback is used by the live tests.

First PR in the four-PR stack. Later PRs also target main and include preceding changes.

PR #461, opened 2026-09-12
feat: configure Plus and Pro budgets for TXT context

jamie950315:
## Summary
- Expose Plus/Pro budget selection in the launcher and CLI for the TXT transport in #460.
- Plus budgets: Instant 123k; other available automatic modes 270k. Pro budgets: Instant 123k; other available automatic modes 400k, with compaction at 90%.
- A Plus budget selection stays conservative even when the authenticated account exposes Pro modes. Selecting a budget does not unlock models unavailable to that account.
- Budgets include file text, instructions, images, and platform reserve. They are configured limits, not measured internal model windows or guarantees of full recall.
- Settings text explicitly states that actual Plus-account file behavior still needs live verification. Restart Codex after changing the budget.

## Validation
- Updated stack heads passed `bun run verify` locally.
- Tests cover plan selection, catalog limits, TXT accounting, per-mode ceilings, and conservative Plus selection on a Pro-capable account.
- Existing real web TXT tests were performed on a Pro account; they must not be presented as actual Plus-account verification.

Stacked after #460; please review the TXT transport first.


miuuyy:
I'm closing this together with #460. It carries the same TXT transport stack, and the larger file budgets still need account-specific measurements. The feature is preserved in #492; these closed PRs will be linked there as design history, not an approved implementation.

PR #468, opened 2026-09-12
fix: preserve retry turn handoff

Safa675:
A retryable ChatGPT Web 503 can cause Codex to retry the same user instruction under a new native `turn_id`. The browser adapter currently rejects that replay as stale, so the task cannot resume after model capacity recovers.

This change records a process-local handoff only for retryable 503 failures emitted by this daemon, binds the exact failed user revision to a single successor turn, and clears the handoff when a fresh instruction arrives. It also adds lifecycle regressions for the accepted retry path and for rejecting a different stale instruction.

Validation:
- `bun test tests/server-lifecycle.test.ts` — 37/37 passed during the original verification run
- `./node_modules/.bin/tsc --noEmit` — passed after reopening the execution session
- `git diff --check` — passed

The repository-wide `verify` remains blocked by pre-existing dependency audit findings in the pinned `hono` and `js-yaml` dependencies.

### DIFF
--- a/src/adapters/chatgpt-web/environment.ts
+++ b/src/adapters/chatgpt-web/environment.ts
@@ -3,6 +3,7 @@ import { isAbsolute, join, relative, resolve, sep } from "node:path";
 import { isReadableCompactionSummaryText, OPAQUE_COMPACTION_NOTE } from "../../responses/compaction";
 import type { CodexContentPart, CodexParsedRequest, CodexTool } from "../../types";
 import { isAcceptedCompactionContinuation } from "./compaction-continuation";
+import { clearRetryableTurnHandoff, isAcceptedRetryContinuation } from "./retry-continuation";
 
 export type ChatGptSandboxPolicy =
   | { type: "dangerFullAccess" }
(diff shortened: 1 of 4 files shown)


miuuyy (changes requested):
A targeted lifecycle sequence exposes a handoff-clearing race: after failed revision A is accepted under successor turn B, a delayed request from original A clears the handoff; B's next tool round then fails turn-ID validation.

Revision validation should be side-effect free, and handoffs should be retired only on a proven newer instruction or lifecycle event. Please cover delayed/replayed original requests interleaved with the accepted successor. The native 503-to-new-turn transition also still needs actual lifecycle evidence beyond the constructed request fixtures.

miuuyy:
Thank you for the proposed retry handling. Closing this implementation because a delayed request from failed turn A can clear the handoff already accepted by successor B, making B's next round fail turn-ID validation. The reproduced sequence in the existing review still applies to this head. The motivating native 503-to-successor transition also lacks captured lifecycle evidence, so the added state cannot currently be accepted as a reliable fix.

Issue #438, opened 2026-09-11
[Bug]Unable to complete "Install to Codex"


EricZhu-KYO:
### Codex Web GPT version

v5.0.6

### Codex version

26.903.71938

### Platform

Windows x64

### OS version

Windows11

### ChatGPT account tier

Pro

### Integration mode

Setup / sign-in (no model turn yet)

### Selected model

not applicable

### What happened?

When clicking "Install to Codex", a repeated reminder appears: Please exit Codex completely (including background processes) and then reopen it to refresh the model list. Simply exiting and logging back into the account or just closing the window does not constitute a restart. Please keep this launcher open.

The computer has completely exited the Codex process, and even after restarting, it still cannot complete the model installation.
Please help to solve this. Thank you

<img width="752" height="170" alt="Image" src="https://github.com/user-attachments/assets/1e38d3d3-6faf-4770-afd2-8d4d7b8d2f63" />

### Minimal reproduction

 click "Install to Codex",

### Complete final error

```text
install failed
```

### Privacy-safe diagnostics

[codex-web-gpt-diagnostics-2026-09-11.txt](https://github.com/user-attachments/files/32096057/codex-web-gpt-diagnostics-2026-09-11.txt)

### Other Codex wrappers or route owners

none

### Preflight checklist

- [x] I tested the latest release and, for catalog or turn failures, fully restarted Codex including its background process.
- [x] I stated whether the failure also occurs in a fresh Codex task, or explained why that does not apply.
- [x] I attached only a privacy-safe export, or explained why the exporter could not produce one; I removed any remaining private content.
- [x] I searched existing issues for this exact final error.

miuuyy:
Thanks for the log. It records successful core setup repeatedly, and the final Doctor check says the Codex route is installed. I do not see a successful Codex model-catalog verification. There is also a separate earlier MCP setup failure: Windows refused to launch the tunnel-client executable with `EACCES`; that does not explain the later Browser-only catalog result by itself.

Could you provide these two details?

1. With the current installation left in place and the launcher running, open the Codex model picker, then run this read-only PowerShell command and paste the output:

```powershell
Invoke-RestMethod http://127.0.0.1:17841/healthz | Select-Object version, mode, successful_model_catalog_requests, last_successful_model_catalog_request_at
```

[...]

PR #469, opened 2026-09-13
fix: apply Electron PAC to managed native processes

liu-dongfang:
## What this changes

When Electron uses a system PAC, the embedded ChatGPT page follows that proxy decision, but launcher-owned native children do not. On affected macOS networks, the browser remains usable while `tunnel-client` tries to reach the OpenAI tunnel control plane directly, times out, and Full-harness tool calls end as `Session terminated` before reaching the local broker.

This change resolves Electron's PAC for the native Codex endpoint and OpenAI tunnel control plane, then passes the resulting environment only to the managed Responses proxy and tunnel-client children. Existing explicit proxy variables retain priority, `DIRECT` remains direct, and loopback addresses remain in `NO_PROXY`.

## Evidence

Before the change, direct access to the tunnel control plane timed out while the same request through the PAC-selected HTTP proxy returned promptly. The installed launcher repeatedly logged control-plane dial timeouts, and a real Full-harness Web turn failed on its first local tool call without reaching the broker.

After the change:

- tunnel health returned `healthz=200`, `readyz=200`, and a successful control-plane poll;
- a real installed Full-harness `chatgpt-web/extra-high` task completed 36 separate file reads, image inspection, file writes, explicit context compaction, post-compaction cwd recovery, and final artifact validation;
- the native Codex passthrough returned the exact probe marker;
- focused clean-branch tests passed: 67/67;
- the packaged candidate's full launcher suite passed: 307 passed, 1 skipped, 0 failed;
- the root runtime suite passed: 1082 passed, 2 skipped, 0 failed.

## Scope and invariants

## Verification

- [ ] I ran `bun install --frozen-lockfile` in the repository root and `launcher/`.
- [x] If this changes ChatGPT browser UI handling, I included observed DOM evidence and a reproducible fixture instead of broadening selectors speculatively. (No browser DOM handling changed.)

The two unchecked commands were not rerun in the clean PR worktree because it intentionally has no installed dependencies. The focused tests run there passed. The complete suites, typecheck, renderer build, runtime bundle build, package signing check, and runtime manifest validation were run on the same five-file candidate before extracting this clean commit.

- macOS arm64: real installed Full harness, ChatGPT Web Extra High, passed.
- Native Codex passthrough through the installed bridge: passed.
- macOS arm64 ZIP and DMG: built and validated.
- Windows packages: not run.
- Linux packages: not run.
- Browser-only: not rerun; this change does not expose MCP or broker capabilities there.


miuuyy:
Thanks for the real-world proxy evidence. The network mismatch is worth investigating, but the native proxy setting added here is not consumed by the bridge, and the PAC/NO_PROXY handling leaves compatibility gaps. I'm closing this implementation; the related proxy case in #438 remains open for a focused fix.

PR #472, opened 2026-09-13
feat: transport large Full-mode context through the turn-bound MCP bridge

### LINKED ISSUE #471: Proposal: keep large Full-mode Codex context out of the ChatGPT composer
## Problem

A short human Codex request can expand into a very large canonical Full-harness context once Codex adds system/developer instructions, conversation history, environment context, tool results, and code/file excerpts.

Today that complete role-preserving JSON envelope is inserted into the ChatGPT Web composer. With large contexts this makes the browser-automation phase disproportionately expensive: the Lexical editor has to accept, reconcile, render, and submit a very large DOM-backed message even though most of that data is machine-to-machine transport rather than something the user needs to see in the ChatGPT UI.

The goal is specifically to reduce browser/composer overhead without changing the semantic Codex task, dropping history, summarizing tool results, or weakening the Full-harness capability boundary.

## Proposed design

For automatic Full mode only:

1. Compile the canonical Codex JSON envelope exactly as today.
2. If the canonical envelope is at least 32,768 characters, do **not** paste that envelope into the ChatGPT composer.
3. Compute SHA-256 metadata and an opaque `ctx_...` context ID, split the exact string into bounded 32,768-character chunks, and install it into the already turn-bound `TurnBroker` before browser submission.
4. Send ChatGPT only a small manifest containing the context ID, hash, character/byte count, chunk count, chunk size, and the existing `turn_token`.
5. Read the context back through the existing **read-only `codex_tool_inventory` MCP operation** using a bridge-reserved query of the form `codex_web_context_read:<context_id>` and `offset=<chunk>`.
6. Require every chunk to report the same context ID and SHA-256 and require the complete chunk sequence before executing the task. Any missing reader, failed chunk, hash/ID mismatch, or incomplete sequence fails closed instead of executing partial context.

Small Full-mode requests remain inline. Browser-only, Zero Risk/manual, compaction, Bigger Context, and image-attachment behavior remain unchanged.

## Why use the existing turn-bound MCP bridge?

The Full harness already has the right security/lifecycle boundary for local Codex capabilities:

- an opaque turn token identifies one live broker channel;
- the broker already owns the exact trusted Codex environment for that turn;
- capabilities retire with the turn;
- stale or revoked handles fail closed;
- no filesystem path or extra local HTTP/file-upload transport has to be exposed to ChatGPT.

Keeping the large payload in the broker therefore reuses the existing trust boundary instead of adding another transport or credential surface.

The context becomes immutable once MCP binding has started and is retired with the turn. Input-token accounting continues to use the canonical context, not the small manifest, so moving bytes out of the composer does not make usage/accounting pretend the context disappeared.

## Why the context reader uses `codex_tool_inventory`, not `codex_tool_call`

The first implementation exposed a bridge-reserved `codex_web_context_read` wire name through inventory and invoked it through the generic `codex_tool_call` dispatcher. A real installed Full-harness test showed that this is the wrong safety shape: ChatGPT correctly sees the generic dispatcher as potentially mutating/open-world, so the first required context read can be blocked by tool safety.

Marking the generic dispatcher read-only would be incorrect because it can legitimately dispatch mutating native tools. Adding a new top-level public MCP tool would also change the connector tool schema/identity for what is fundamentally a private transport primitive.

The revised design routes only the reserved context-read query through the already read-only `codex_tool_inventory` operation. This preserves the generic `codex_tool_call` safety annotations and the public connector tool list while giving the transport operation accurate read-only semantics.

## Why not summarize/compress/truncate instead?

This payload is the canonical Codex task context. It contains role ordering, instruction priority, prior assistant/user messages, environment context, and tool results. Lossy summarization or truncation can change the task. The proposed transport is deliberately lossless: chunks reconstruct the exact canonical JSON string and are verified against the same SHA-256.

## Validation performed in a fork

The implementation has been exercised in `EmpRider/codex-chatgpt-web` with focused regression tests for prompt compilation, broker lifecycle/context immutability, exact chunk reconstruction, adapter installation, and the read-only safety path.

A real installed Full-harness end-to-end run used a small human request while the generated canonical context was **70,010 characters / 3 chunks**. The visible ChatGPT user message contained only the MCP manifest, the hidden context was reconstructed through the read-only inventory path, and the model then completed the original repository-analysis task successfully. This directly tests the behavior this proposal targets: keeping the generated large transport payload out of the browser composer while preserving the complete task.

The fork's cross-platform CI/release validation also passed on Windows, Ubuntu, macOS arm64, and macOS Intel, including package/smoke checks.

## Scope for the PR

The contribution will be intentionally narrow:

- context transport primitive;
- prompt compilation / token accounting integration;
- TurnBroker context installation/read lifecycle;
- MCP read-only inventory routing;
- focused regression tests;
- a small architecture-doc update.

It will **not** include fork version bumps, release workflow changes, launcher/UI customizations, Chat Clean changes, dependency updates, or unrelated cleanup.

I can open the implementation as a draft PR against current `main` for review if this direction is acceptable.

EmpRider:
## What this changes

Closes #471.

Large automatic Full-harness turns can produce a very large canonical Codex JSON envelope even when the human request itself is small. The current path pastes that complete machine-generated envelope into the ChatGPT Web Lexical composer, so browser automation pays the cost of inserting, reconciling, rendering, and submitting tens or hundreds of kilobytes of transport data.

This PR keeps that large canonical context out of the visible browser message while preserving it losslessly:

- Full-mode canonical contexts below 32,768 characters keep the existing inline path.
- At 32,768 characters or above, the exact canonical JSON is hashed, assigned an opaque `ctx_...` identity, split into bounded 32,768-character chunks, and installed in the turn-bound `TurnBroker` before browser submission.
- ChatGPT receives only a compact manifest containing the context identity/hash, character and byte counts, chunk metadata, and the existing turn capability.
- The model retrieves every chunk through the existing **read-only `codex_tool_inventory`** MCP operation using the reserved query `codex_web_context_read:<context_id>` and `offset=<chunk>`.
- Every response carries matching context ID/SHA-256 metadata and `next_chunk`; the prompt requires the complete sequence before task execution and fails closed on missing/failed/mismatched context.
- The context becomes immutable after MCP binding begins and retires with the turn capability.
- Input-token estimation continues to account for the canonical MCP-delivered context rather than the small browser manifest.

Browser-only, Zero Risk/manual, compaction, Bigger Context, and image attachment behavior are unchanged.

### Why use the existing MCP/TurnBroker path?

This data belongs to the same Codex turn that already owns a narrow, opaque broker capability. Reusing that boundary avoids introducing a second local HTTP endpoint, temporary context file, filesystem path, upload mechanism, or credential surface. Context installation happens before browser submission, reads are bound to the live turn, stale capabilities fail closed, and the payload retires with the turn.

This is intentionally **lossless** rather than a summarization/truncation optimization. The canonical envelope contains role ordering, instruction priority, conversation history, environment context, and tool results; changing those bytes can change task semantics. Chunking only changes transport, not content.

### Why read through `codex_tool_inventory` instead of `codex_tool_call`?

The first implementation exposed a bridge-reserved `codex_web_context_read` wire name and dispatched it through generic `codex_tool_call`. A real installed Full-harness test surfaced an important safety mismatch: the generic dispatcher is correctly annotated as potentially mutating/open-world, so ChatGPT's tool safety can block a context read before the task starts.

Changing `codex_tool_call` to read-only would be incorrect because it legitimately dispatches mutating native tools. Adding a new top-level MCP tool would also change the connector's public tool schema/identity for what is only a private transport primitive.

The final implementation therefore routes only the exact reserved context query through `codex_tool_inventory`, which is already the read-only discovery surface. The generic dispatcher and its safety annotations remain unchanged, the context reader is not exposed as a normal inventory descriptor, and no new public MCP tool is registered.

### Before

A short Codex request can expand into a generated Full-harness browser message containing the complete role-preserving JSON context. With large repository/tool histories, the browser automation phase becomes the bottleneck because the huge transport payload is pasted into the ChatGPT composer DOM.

### After

In a real installed Full-harness test, the human request was a normal repository-analysis task while the generated canonical context was **70,010 characters across 3 chunks**. The visible ChatGPT user message contained only the MCP context manifest, not the 70K canonical envelope. The read-only inventory transport reconstructed the hidden context and the model then completed the original repository-analysis task successfully.

An earlier installed test using the generic `codex_tool_call` reader reproduced the safety failure described above. The final focused regression explicitly proves the old generic dispatcher path is rejected while the reserved read-only inventory query succeeds without queuing a native broker tool.

The clean contribution branch is based directly on upstream `main` commit `e85e3693fdb4e3e033348c08df0298c20fcdb612`, contains two focused implementation commits, and changes only the context-transport implementation, its tests, and the architecture paragraph. It contains no fork version/release/launcher/Chat Clean changes.

## Scope and invariants

- [x] This is a focused change with no unrelated cleanup or generated rewrite.
- [x] Full-harness Web efforts retain the same turn-bound capability and Browser-only gains no broker or connector.
- [x] Terms and trademark claims remain factual; this is a browser/composer performance fix, not a quota or rate-limit bypass.

## Verification

- [x] Ran `bun install --frozen-lockfile` in both the repository root and `launcher/` with Bun 1.4.0 on the clean upstream-port branch.
- [ ] `bun run verify` was attempted. It stops at dependency audit findings already present on the upstream base; this PR changes no package manifest or lockfile. Those audit findings were recorded rather than suppressed or fixed in this unrelated PR.
- [x] Added/updated focused regression tests for prompt externalization, exact chunk reconstruction, broker context lifecycle/immutability, adapter installation, and the read-only safety route.
- [x] Manually tested the affected behavior through a real installed Full-harness integration.
- [x] Because this changes MCP execution / the outer Codex loop, validation was not limited to DEV mode.
- [x] No ChatGPT DOM selector or browser UI parsing behavior is changed; the change reduces the text inserted into the existing composer path.
- [x] No launcher source is changed by this PR.
- [x] No browser state, credentials, Tunnel IDs, raw logs, generated artifacts, or private paths are committed.
- [x] No dependency updates, release artifacts, workflow changes, or version changes are included.

### TDD / focused checks

For the safety refinement, the regression test was tightened **before** removing the generic `codex_tool_call` compatibility path:

1. **RED:** the tightened test failed while the generic dispatcher still accepted `codex_web_context_read`.
2. The compatibility path was removed so exact context reads are served only by the reserved read-only inventory query.
3. **GREEN:** `mcp-context-transport`, `mcp-context-read-safety`, `mcp-context-adapter`, and `prompt-contract` passed: **28 passed, 0 failed**; root typecheck also passed.

The earlier clean-port focused suite covering prompt, transport, adapter, safety, harness, and DEV chat passed **123 tests, 0 failures**.

### Broader clean-branch verification

After recording the unchanged upstream dependency-audit findings, the remaining repository gates were run on the final contribution branch and passed:

- exact 13-file changed-file allowlist and `git diff --check`;
- version alignment (`5.0.6` remains unchanged);
- `bun run check-version`;
- root typecheck and full root test suite;
- launcher typecheck and full launcher test suite;
- launcher renderer build;
- runtime bundle build;
- third-party-notice generation;
- release/runtime smoke validation.

The separate verification job containing those post-audit gates completed successfully.

- Real installed Full harness / Codex Native2: passed with a 70,010-character / 3-chunk generated context.
- ChatGPT Web account used for the installed test: authenticated account with the Full harness connector available.
- Windows x64 fork release validation: passed.
- Ubuntu x64 fork release validation: passed.
- macOS arm64 fork release validation: passed, including packaged signature checks.
- macOS Intel fork release validation: passed, including packaged signature checks.
- Browser-only: not rerun for this feature; no Browser-only capability or transport path is changed.
- Zero Risk/manual: not rerun for this feature; it explicitly retains the existing transport path.


miuuyy:
Thanks, but I'm not taking this approach. I don't consider fetching the task context through MCP reliable enough for this bridge. Sorry, this is not planned.

EmpRider:
Actually, I'm using this in my own fork, and it works super fine. The reason I used this approach is that when our chat groves web page got hung when we sent a huge user message, the Codex session went timeout before the web page got a response back. In this method, I convert a huge prompt into small chunks and give a super small user message in the web page chat to tell it to get the actual prompt via MCP call that helps to avoid the web page hanging and continue the session without interruption. Before this modification, I could not work in a single chat for more than 10 minutes because the webpage got hung and the session terminated every time. After this modification, I can use a single session for more than 2 days. I'll attach an SS for reference; it has a 9-hour running time. In a single chat,

BTW, I respect your decision. Thank you.
image

PR #478, opened 2026-09-14
docs: clarify Codex catalog restart

RRiiiccckkk:
## What this changes

Related to #438.

The setup warning currently tells users to restart Codex, but the affected desktop host can appear as **ChatGPT** and may keep a bundled `codex app-server` alive after the visible window closes. The verification is request-driven: opening the model picker after the restart is what sends the catalog request that lets the launcher complete step 3.

This update carries those details into the troubleshooting guide and the English, Chinese, and Japanese launcher copy. It also tells users where to look when the safe health data still reports zero catalog requests, so repeated reinstalls do not hide the useful diagnostic.

## Evidence

Issue #438 reports the same state after re-login and a full desktop restart: the launcher keeps showing the restart warning and ChatGPT Web models remain absent. The existing discussion identifies the stale desktop/app-server catalog as a relevant failure mode and uses `successful_model_catalog_requests` to distinguish an installation problem from a catalog request that never reached the bridge.

## Scope and invariants

- [x] This is a small, focused documentation and UI-copy change.
- [x] It preserves Codex's native catalog and the explicit ChatGPT Web model entries.
- [x] It does not change Full/MCP behavior, tool routing, credentials, browser state, or private paths.
- [x] All factual terms match the current implementation and troubleshooting flow.

## Verification

- [x] I ran `bun install --frozen-lockfile` in the repository root and `launcher/` with Bun 1.4.0.
- [ ] I added or updated a focused regression test for behavior changes.
- [x] I manually reviewed the affected troubleshooting steps and all three locale strings.
- [ ] If this changes local tools, MCP execution, or the outer Codex agent loop, I tested it through a real installed Codex integration.
- [ ] If this changes browser UI or DOM assumptions, I captured fresh DOM evidence.
- [ ] If this changes launcher packaging or release behavior, I ran the launcher packaging checks.
- [x] No secrets, cookies, browser storage, private paths, or raw prompts are included.
- [x] No unrelated dependency, release, or version files were changed.

`bun run --cwd launcher typecheck` passes. This PR changes only documentation and localized copy, so the behavior-specific and packaging checks do not apply. The root verification audit remains blocked by the existing Hono advisories documented in the earlier setup PR.

Documentation and copy only. Reviewed for the macOS and Windows restart paths described in #438; no account, browser, or model-turn behavior changes.


RRiiiccckkk:
Addressed all Copilot review points. The troubleshooting guide now separates the macOS menu-bar quit action from Activity Monitor process confirmation, repeats the `ChatGPT` host alias in the step-3 section, and provides direct curl/PowerShell `/healthz` queries because Activity safe-log export does not include the catalog counters.

miuuyy:
Thanks. I'm closing this version because it does not distinguish a failed catalog request from no request: a zero successful-request counter proves neither restart failure nor absence of traffic. The separate cases in #438 still need their own diagnosis, so this wording would be misleading.

PR #488, opened 2026-09-15
fix: Reject unsupported ChatGPT Web input images at HTTP boundary

ksjxixi-gif:
What this changes

Web-routed Responses turns carrying an input image that is not an inline base64 data URL (png, jpeg, gif, webp) are now rejected with an explicit HTTP 400 at the HTTP boundary, before the browser adapter is constructed. Previously the request was accepted and the turn died mid-flight inside the browser worker (chatGptImageFilePayloads threw ChatGPT web input image <ref> must be an inline base64 data URL after browser/adapter work had already started).
Fixes # (no linked issue; found during local code review)
Evidence

New regression tests in tests/server-web-image-validation.test.ts:
- { model: "chatgpt-web/light", input: [user message with input_image image_url "https://example.com/cat.png"] } → HTTP 400 invalid_request_error, message contains "must be an inline base64 data URL", adapter factory invoked 0 times.
- Same request with data:image/bmp;base64,... → HTTP 400, message contains "unsupported media type: image/bmp", adapter factory invoked 0 times.
- Same request with an inline data:image/png;base64,... → HTTP 200, adapter constructed exactly once (valid images still pass).
No browser UI handling changed, so no DOM evidence applies. The replaced mid-turn throw is covered by existing browser-worker-contract assertions on the same message text.
Scope and invariants
- I read and followed CONTRIBUTING.md.
- This is a small, focused change with no unrelated cleanup or generated rewrite.
- The change stays focused on ChatGPT web-backed Codex models; it does not add a generic provider or unrelated product surface.
- Model, route, effort, connector, and capability selection remain explicit with no silent fallback or false-success path.
- If this touches Full harness or MCP, every available Web effort retains the same turn-bound capability and Browser-only gains no broker or connector. (Not touched.)
- Terms and trademark claims remain factual; this change is not marketed as a quota or rate-limit bypass.
Verification
- I ran bun install --frozen-lockfile in the repository root and launcher/. (Ran plain bun install in root only.)
- I ran bun run verify with the Bun version pinned by package.json. (Ran root bun run typecheck — clean — and bun test ./tests instead. Local Bun is 1.3.14, package pins 1.4.0.)
- I added or updated a focused regression test for behavior changes.
- I manually tested the affected behavior. (Via focused tests driving responseRequest directly.)
- If this changes local tools, MCP execution, or the outer Codex agent loop, I tested it through a real installed Codex integration; DEV mode alone is acceptable only when execution is not affected. (Not applicable — none of those changed.)
- If this changes ChatGPT browser UI handling, I included observed DOM evidence and a reproducible fixture instead of broadening selectors speculatively. (No UI handling changed.)
- If this changes the launcher, I preserved macOS, Windows, and Linux packaging and named the platform packages actually built below. (No launcher change.)
- I did not commit browser state, credentials, Tunnel IDs, raw logs, generated artifacts, or private paths.
- I did not include an unrelated dependency update, release artifact, or version change.
Platform or account validation

- Windows, Browser-only mode, root suite + typecheck: ran (typecheck clean; new tests 3/3 pass).
- Full root suite: ran, with 1 pre-existing failure in browser-worker-contract.test.ts ("Bigger Context fits mixed-density…", 30s+ timeout) that also fails on main without this change.
- macOS: not run. Linux: not run. Packaged builds: not run. Paid-tier browser flows: not run. Full mode / MCP: not run.


miuuyy:
Thanks. The bounded checks did not establish a new failure in the supported image path: unsupported formats already fail explicitly. This change also rejects old images that the compiler has correctly omitted from the outgoing attachments. I'm closing this implementation; the validation idea and that constraint are preserved in #498.

PR #534, opened 2026-09-16
fix: preserve explicit Codex skill provenance

Subaru486desuwa:
## What this changes

Preserve the provenance of explicit Codex skill injections when translating a Responses request into the ChatGPT Web prompt.

Codex appends selected skill instructions as a server-owned `role=user` item. Current Codex identifies that item with `internal_chat_message_metadata_passthrough.content_item_kinds = ["skills.selected_skill_instructions"]`. The bridge previously flattened it into an ordinary human `user` message, which could make the Web model treat the skill payload as the latest human request and invent a generic `Skill` tool that is not exposed by the harness.

This change:

- marks verified selected-skill payloads as `origin: "codex_skill"` during Responses parsing;
- emits them as a distinct `codex_skill` role in the encoded Web context;
- tells the Web model to preserve user precedence and not invent a generic `Skill` tool;
- retains a narrow compatibility path for older Codex transcripts where the human explicitly selected `$skill-name` but `content_item_kinds` is absent;
- keeps literal human-authored `<skill>...</skill>` text as a normal user message.

## Evidence

Observed current Codex wire shape in a real Codex integration:

- human instruction is a `role=user` message with `content_item_kinds=["user.text"]`;
- the selected skill is appended as another `role=user` message with `content_item_kinds=["skills.selected_skill_instructions"]`;
- the skill payload uses the structured `<skill><name>...<path>.../SKILL.md ...</skill>` envelope.

Before this change, parser-to-prompt transport produced `user, user` for the human instruction and selected skill. The regression test first failed on that behavior. After the change it produces `user, codex_skill` while a literal user-authored skill envelope remains `user`.

## Scope and invariants

## Verification

- [x] I ran `bun run verify` with Bun 1.4.0 pinned by `package.json`.
- [x] I added focused regression tests for current Codex metadata, legacy explicit `$skill` selection, and literal human-authored skill XML.
- [x] I manually inspected the affected real Codex wire shape and reproduced the provenance loss before the change.
- [ ] I did not replace the currently running production Launcher with this patched build while it was carrying the active repair session, so final installed-runtime E2E validation of the patched branch is still pending.
- [x] This does not broaden ChatGPT browser UI selectors.
- [x] No launcher source is changed by this PR.

`bun run verify` result: 710 pass, 1 skip, 0 fail in the core suite; launcher tests/build and relocatable runtime smoke also passed.

- macOS arm64: source tests, launcher tests/build, runtime bundle validation passed.
- Real Codex integration: affected pre-fix wire shape observed; patched installed-runtime E2E not run yet to avoid interrupting the active production session.
- Browser-only: parser/prompt regression coverage passed.
- Full mode: parser/prompt regression coverage passed; installed-runtime E2E pending as noted above.


miuuyy:
Thanks for identifying the native selected-skill metadata. I implemented a different, opt-in approach in [v5.0.8](https://github.com/miuuyy/codex-chatgpt-web/releases/tag/v5.0.8): **Settings → Skills as files (experimental)**. Explicitly selected skills are uploaded as named UTF-8 text files, with their original path/resource envelope preserved. The experiment is off by default; normal skill discovery/tool reads stay unchanged.

The parser recognizes only the explicit native `skills.selected_skill_instructions` metadata. I did not adopt the `$skill-name`/XML-text compatibility heuristic, which can misclassify ordinary user text across turns. This is an independent implementation, so I am closing this PR without merging it. File delivery, retained-chat behavior, token accounting and helper compatibility are tested; better model adherence is not yet established. Please try the experimental switch on 5.0.8 and report a fresh safe log if delivery fails.

PR #540, opened 2026-09-16
[codex] Clarify Native2 dynamic-tool invocation and local capabilities

JulianZJN:
## What this changes

Make the existing Native2 tool invocation contract explicit, without adding tools or changing dispatch/approval behavior.

- Add an additive `invocation` field to `codex_tool_inventory`: direct function and tool-search entries use `codex_tool_call.arguments`, while freeform entries use `codex_tool_call.input`. Gateway-only entries explicitly defer to the nested declaration rather than guessing its schema.
- Before calling an unfamiliar discovered tool, instruct the Full-mode model to read its exact inventory entry with `include_schema=true`, follow the returned invocation/schema, and respect declared timeout bounds.
- Explain the existing local entrypoints: `codex_exec` can read text such as TeX/CSV and run installed parsers/compilers; `codex_apply_patch` creates or edits text. This does not add a TeX/CSV reader, install a compiler, or grant filesystem access that the outer Codex turn does not have.

The motivating symptom is a Web-backed model treating generic local capabilities as missing, or guessing the wrong payload shape for a discovered tool. The bridge already distinguishes structured arguments from freeform input; this PR exposes that distinction instead of changing execution.

## Evidence

The previous inventory described tool `kind` and optional `parameters`, but did not state the corresponding `codex_tool_call` field. The prompt only referred generically to declared descriptions/schemas.

Regression coverage now checks:

- Real stdio MCP inventory and broker dispatch for `tool_search` and a namespaced function parsed from Responses Lite `tool_search_output`.
- Calling the exact returned `wire_name` with structured arguments, without routing a direct function through the freeform exec gateway.
- Rejecting freeform `input` for the discovered function; the rejected call is not queued at the broker.
- Freeform and gateway inventory hints, plus Responses Lite serialization as a client `tool_search_call`.
- Prompt guidance while preserving existing turn-token, completion, and Browser-only/compaction boundaries.

These are protocol/fixture tests, not evidence that every live ChatGPT model will always follow the guidance.

## Scope and invariants

- Five files; production changes are three prompt lines and two additive inventory fields.
- Existing tool registration, dispatch, validation, outer-Codex approvals, and turn ownership are unchanged.
- All available Full-mode efforts retain the same turn-bound capability. Browser-only gains no broker or connector.
- No model/effort fallback, routing changes, UI changes, dependency updates, or version bump.
- No local usage-dashboard, compaction-policy, or machine-specific router patches are included.
- No browser state, credentials, tunnel IDs, private paths, or raw session logs are committed. Test identifiers are synthetic.

## Verification

Based on upstream `main` at `2f5f0b9e6526415973ff104840c3b78e972792f9`, using pinned Bun 1.4.0 on macOS arm64:

- `bun install --frozen-lockfile` in root and `launcher/`: passed; lockfiles unchanged.
- Root and launcher dependency audits: no vulnerabilities found.
- Root and launcher TypeScript checks: passed.
- `bun test ./tests --timeout 15000`: **711 passed, 1 Windows-only skip, 0 failed**.
- `bun run launcher:test`: **308 passed, 1 platform-conditional skip, 0 failed**.
- `bun run launcher:build`, `bun run build` (including third-party notices), and `bun run smoke`: passed, including `RELOCATABLE_RUNTIME_SMOKE_OK`.
- `git diff --check`: passed.

**The unmodified `bun run verify` command is not green on this machine.** It stops at the pre-existing `native fetch reaches a proxy-only target, refreshes routing, and never retries a failed route` test after its default 5000 ms timeout. A clean checkout of upstream `2f5f0b9` reproduces the same failure. That test includes multiple 2000 ms negative network waits and passes here in about 6.4 seconds with `--timeout 15000`; the full core suite also passes with that test-runner setting. No network test or product timeout is changed in this PR. The remaining verify steps were run separately and passed.

The same production guidance/inventory additions are present in a locally installed enhanced 5.0.7 build. Its packaged startup, authentication/health checks, and installed prompt-string inspection passed, but that build includes unrelated local features and is not a substitute for isolated-branch acceptance evidence.

**Not run for this PR:** a fresh paid ChatGPT task through the installed Codex integration, cross-platform installed MCP sessions, or platform installer packaging for this isolated branch. Keeping this as a draft makes that acceptance boundary explicit; no claim is made that the stdio fixture is a live ChatGPT/Codex end-to-end test.


miuuyy:
Thanks for the contribution and the detailed protocol tests. I compared this with the native Codex tool descriptions and the instructions already transported by the bridge: command execution, file editing, and tool schemas are already exposed. Our MCP wrappers need their own transport contract, so the instructions are not literally identical to native Codex, but I want to keep them close to it.

This PR adds invocation hints and prompt guidance without changing dispatch or fixing a demonstrated dispatch failure. The tests validate the existing protocol; there is no isolated live before/after reproduction showing that these extra instructions resolve the reported behavior. I am therefore closing this rather than adding more prompt rules to the hotfix. A reproducible failing call or captured missing capability would be useful for a targeted fix.

PR #544, opened 2026-09-17
chore: route native Codex traffic to prism.openai.com

ty1563:
Routes native Codex backend requests from `chatgpt.com/backend-api/codex` to `prism.openai.com` via the `CODEX_BACKEND` constant in `src/native-passthrough.ts`. All endpoint suffix logic (`/${endpoint}`) is preserved. Test assertions updated accordingly.


miuuyy:
This changes the destination of all native Codex passthrough traffic, including responses, compaction, model discovery, search and images, while preserving the same authentication and endpoint suffixes. Updating mocked URL assertions only proves the string substitution.

I cannot include this in the compatibility patch without evidence that Prism provides those authenticated Codex contracts and preserves the native routes. A separate scoped integration proposal with real protocol evidence is needed; the current native backend must remain intact.

miuuyy:
Closing this proposal in its current form. Replacing the shared native Codex backend with the Prism hostname changes responses, compaction, model discovery, search and image routes for existing users, without establishing their authenticated protocol compatibility. The changed mocked URL assertions demonstrate the substitution, not a working integration. Any future Prism proposal needs to be a separately scoped integration supported by actual protocol evidence, while preserving existing native routes. Thank you for the proposal.

Issue #549, opened 2026-09-17
Full Harness setup stops Launcher-owned MCP runtime after readiness


whatcall:
### Summary

The full Harness setup can fail after the managed MCP runtime has already started and become healthy. The setup flow starts a Launcher-owned runtime to validate the tunnel, then unconditionally stops it before the Launcher supervisor takes ownership. When the MCP stdio child does not exit promptly after SIGTERM, setup reports a failure and the runtime profile may be removed or left stopped.

This was reproduced on macOS arm64 with the 5.0.7 and 5.0.8 runtime flows and tunnel-client 0.0.12. The exact failure is reported as:

```text
process <pid> did not exit after SIGTERM
```

The failure is independent of the Codex account session once the tunnel has authenticated and reached readiness.

### Reproduction

1. Install the Launcher and configure a full Harness with valid tunnel credentials.
2. Start the native MCP setup from the Launcher.
3. Let the managed runtime become healthy and ready.
4. Observe the setup cleanup/handoff step.

### Expected behavior

After the runtime becomes healthy and ready, the Launcher-owned runtime should remain running and be adopted by the Launcher supervisor. Setup should complete with the MCP profile and health state intact.

### Actual behavior

Setup stops the runtime after validation. If the MCP stdio child does not exit within the stop operation's deadline, setup fails with a SIGTERM-related error. A subsequent status check reports a stopped runtime, `ready: false`, and may report that the recorded profile is missing.

### Evidence

The setup implementation currently routes both Launcher-owned and external-service setup through the same bootstrap helper. That helper calls `connect`, waits for readiness, and then always calls `stop`. The Launcher path subsequently attempts to start or recover a runtime that has just been stopped.

### Suggested direction

Treat runtime ownership explicitly:

- For Launcher-owned setup, keep the validated runtime alive and let the supervisor adopt it.
- For an external system service, stop the temporary validation runtime before installing or starting the service.
- Preserve cleanup on failed bootstrap so an incomplete runtime is not leaked.

The issue is intended to document the lifecycle problem and provide a reproducible case. A small reference implementation is proposed separately in a pull request.

whatcall:
A reference implementation is available in [PR #550](https://github.com/miuuyy/codex-chatgpt-web/pull/550). The PR keeps a successfully validated Launcher-owned MCP runtime alive during setup while preserving temporary-runtime cleanup for external service installation. It is provided for discussion and reference; it is not intended to require merging as-is.

[...]

PR #550, opened 2026-09-17
Keep Launcher-owned MCP runtime alive after setup

whatcall:
### Problem

Full Harness setup starts a managed MCP runtime, waits for it to become healthy, and then unconditionally stops it. This is appropriate only for a temporary validation runtime. In the Launcher-owned path, the runtime is the process that the Launcher supervisor is expected to adopt.

When the MCP stdio child does not exit promptly after SIGTERM, setup fails with a message such as `process <pid> did not exit after SIGTERM`, even though the runtime had already become ready.

### Change

Make the bootstrap helper ownership-aware:

- Keep the runtime alive after successful validation when setup is Launcher-owned.
- Continue stopping the temporary runtime for external system-service setup.
- Keep the existing cleanup behavior when bootstrap fails.

This is deliberately a small reference fix. It does not change tunnel authentication, timeout values, or the external service lifecycle.

### Validation

- Typecheck passes.
- Tunnel status and diagnostic tests pass: 7/7.
- The patched runtime bundle was tested locally: full Harness setup completed and the managed runtime remained ready.

### Scope

This PR is submitted as a reference implementation for discussion. It does not assume that the patch should be merged as-is; the project may prefer a different supervisor handoff design.


miuuyy:
Thank you — the redundant setup/cleanup ownership problem is confirmed. I implemented a local variant: launcher and DEV setup commit configuration without acquiring the tunnel; the supervisor creates the profile and owns startup, readiness and failed-start cleanup. External-service bootstrap retains its existing stop.

Keeping the setup-started runtime alive alone leaves ownership unclear if a later configuration/integration step fails before commit. The local tests cover that gap and first-time profile creation as well as a real local manager subprocess. This is unreleased and has not been validated against the reporter's live tunnel.

miuuyy:
Thank you for the concrete setup reproduction and the reference implementation. Closing this PR in favor of the local supervisor-owned startup correction described above: setup commits configuration first, and the supervisor owns runtime startup, readiness and failed-start cleanup. This also covers failures between configuration and runtime acquisition. Your diagnosis is credited. The local correction is tested but remains uncommitted and unreleased; acceptance on the original live tunnel is still outstanding in #549.

Issue #428, opened 2026-09-10
[Bug] Follow-up turns omit the Codex Native2 connector mention


BrannonSD:
### Codex Web GPT version

5.0.6

### Codex version

Codex Desktop package 26.903.8094.0; codex-cli 0.153.4

### Platform

Windows x64

### OS version

Windows 11 Enterprise 24H2, build 26200 (10.0.26200)

### ChatGPT account tier

Plus

### Integration mode

Full harness (MCP)

### Selected model

ChatGPT Web — High

### What happened?

When replying to another message in an existing agent chat, the browser automation does not prepend/select the `Codex Native2` connector again before submitting the follow-up turn.

The browser turn can still be accepted and completed, but the Web agent then says it does not have access to the local tooling and cannot perform the requested local operation. No local tool call is made for that follow-up.

Expected behavior: every Full harness browser-submitted turn that needs local access should restore/select the verified `Codex Native2` connector as needed, including replies and follow-ups in an existing agent chat.

Actual behavior: a retained/reused chat follow-up is submitted without the Native2 connector mention/selection, so the model loses access to local tooling.

### Minimal reproduction

1. Use Codex Web GPT 5.0.6 on Windows x64 with the Plus plan, Full harness (MCP), a verified `Codex Native2` connector, and `ChatGPT Web — High`.
2. Start an agent chat/task and let it produce at least one response.
3. In that same existing chat, reply to another message or send a follow-up that requires a local Codex operation.
4. Observe the retained/reused browser turn and its submitted composer state.
5. The follow-up is accepted, but the agent reports that it does not have access to local tooling and does not issue the local tool call.
6. As a comparison, a fresh/new agent turn prepends/selects `Codex Native2` and can execute local `exec_command` calls.

This is specifically the existing-chat follow-up path; it is not a global connector setup failure.

### Complete final error

Model-visible failure (wording varies): the agent states that it does not have access to the local tooling and therefore cannot perform the requested local operation.

### Privacy-safe diagnostics

[codex-web-gpt-diagnostics-2026-09-10.json](https://github.com/user-attachments/files/32067006/codex-web-gpt-diagnostics-2026-09-10.json)

### Other Codex wrappers or route owners

None identified. Codex Web GPT owns the route and the verified connector is `Codex Native2`.

### Preflight checklist

- [x] I tested the current release (5.0.6); this is a retained-chat turn-path failure rather than a catalog-refresh report.
[...]

Issue #304, opened 2026-09-03
[Bug] Codex native executor disabled


hankok:
### Codex Web GPT version

5.0.0

### Codex version

26.831.20005

### Platform

Windows x64

### OS version

Windows 11 pro

### ChatGPT account tier

Plus

### Integration mode

Full harness (MCP)

### Selected model

High

### What happened?

Chat goes on normally, but execution cannot be done. 

The messages seen in the chats are:
- Codex Native 本地工具调用返回 FORBIDDEN: This conversation does not support developer MCPs
- I couldn’t complete the requested task update in this turn because the local Codex executor became unavailable.
- 这个task本身看起来是对的，但当前这次会话里的 Codex Native 执行器仍然被宿主禁用了，所以我没法实际替你运行它。

### Minimal reproduction

This issue happens often after a few rounds of chat. Meanwhile there is another session that run executions normally

### Complete final error

```text
added in "What happened?"
```

### Privacy-safe diagnostics

[codex-web-gpt-diagnostics-2026-09-03.json](https://github.com/user-attachments/files/31812331/codex-web-gpt-diagnostics-2026-09-03.json)

### Other Codex wrappers or route owners

N/A

### Preflight checklist

- [x] I tested the latest release and, for catalog or turn failures, fully restarted Codex including its background process.
- [x] I stated whether the failure also occurs in a fresh Codex task, or explained why that does not apply.
- [x] I attached only a privacy-safe export, or explained why the exporter could not produce one; I removed any remaining private content.
- [x] I searched existing issues for this exact final error.

acrbaran:
I have this problem too. I'll be following the thread.

miuuyy:
Thanks for the detailed report and diagnostics. The logs show that the local runtime stayed healthy and another conversation continued executing tools, so the rejection appears to be scoped to this retained ChatGPT conversation.

Could you please try the following?

1. In ChatGPT settings, turn **Developer Mode** off and back on, then retry.
2. Let us know whether anything unusual happened while setting up or using the `Codex Native2` plugin—for example an initial setup failure, recreating or renaming the plugin, reconnecting/changing the tunnel, an unexpected Chat/Work mode switch, or selecting another plugin in the same conversation.

If toggling Developer Mode does not help, please also tell us whether a completely fresh Codex task works.

[...]

PR #553, opened 2026-09-17
feat(chatgpt-web): add opt-in fresh-conversation-per-turn isolation

MeroZemory:
## What this changes

Adds an opt-in escape hatch from retained-conversation reuse.

A retained epoch sends only `messages.slice(lastAssistant + 1)` because the browser conversation is assumed to still hold the earlier turns (`src/adapters/chatgpt-web/conversation-key.ts:45-58`). When that assumption does not hold, the follow-up turn loses the context it needs — and, on this installation, the Codex Native connector with it, so local tools stop working after the first turn of a task.

`experimentalFreshConversationPerTurn` (CLI `--fresh-conversation`) withholds the conversation key, so the Launcher cannot match a retained tab: every Codex turn opens a fresh Temporary Chat and receives the complete compiled prompt. `--retained-conversation` restores the default explicitly.

This is the optional isolation mode described in #304 ("an optional experimental isolation mode where every Codex turn starts a fresh ChatGPT conversation"). It does not change the retained path, does not reselect the connector, and does not alter behaviour when unset.

Related: #428

## Evidence

**Before** (v5.0.6 and v5.0.8, Windows, Full mode, `chatgpt-web/high`): the first turn of a task ran local tools; every following turn answered from text only, reporting that the repository or `exec_command` was unavailable. One thread produced 121 goal continuations whose replies were near-identical, 24 of them byte-identical, with 0 tool calls after the first turn.

**After** (same installation, retained reuse disabled): one thread ran **85 tool calls over 25 minutes** across many turns — `exec_command` 67, `apply_patch` 11, `view_image` 6, `write_stdin` 1 — surviving two context compactions. It edited source, added a regression test, ran the project's test suites and a Playwright browser check, and correctly reported the one requirement it could not verify.

I reached that state first by neutralising the conversation-key call in an installed 5.0.8 bundle, then reimplemented it here as the opt-in flag.

**Unit evidence** (`tests/chatgpt-web-harness.test.ts`): with the flag on, turn two's `codex_context_json` carries all three messages

```json
{"version":3,"system":[],"messages":[
  {"role":"user","content":"Inspect the project"},
  {"role":"assistant","content":[{"type":"text","text":"First retained answer"}]},
  {"role":"user","content":"Continue in the same repository"}]}
```

where the retained path sends only the trailing user message. The existing retained test directly above it is unchanged and still asserts the suffix behaviour.

**One thing I could not verify, and a reviewer should weigh:** `connectorIdentity` is gated on `turn.conversationKey` (`src/adapters/chatgpt-web/browser-worker.ts:4269-4272`). With the flag on, turn start no longer carries it. Empirically the connector still attached and tools ran for 25 minutes, but the Launcher is outside this repository and I have no source-level proof of that contract. `tests/browser-worker-contract.test.ts:533-541` pins the current coupling, so if you consider the two to be intentionally bound, this flag may need `connectorIdentity` decoupled from retention instead.

Two further behaviours worth naming: with the flag on, every compaction takes `runFreshCompactionFallback("source_unavailable_before_handoff")` (working, already covered by `tests/retained-compaction.test.ts`, but it logs a warning each time), and each turn resends the full context instead of a suffix, so turns are larger.

## Scope and invariants

## Verification

- [x] I ran `bun run verify` with the Bun version pinned by `package.json` (1.4.0).
- [ ] If this changes ChatGPT browser UI handling, I included observed DOM evidence and a reproducible fixture instead of broadening selectors speculatively. — not applicable; no selector or DOM handling is touched.
- [ ] If this changes the launcher, I preserved macOS, Windows, and Linux packaging and named the platform packages actually built below. — not applicable; no launcher change. No launcher UI toggle was added, matching the CLI-only scope.

`bun test ./tests` on Bun 1.4.0: **720 pass / 2 fail**. Baseline on unmodified `main` in the same checkout: 718 pass / the **same** 2 failures (`reversible native Codex route integration`, pre-existing Windows `EPERM: symlink` at `tests/codex-integration.test.ts:108`). The two added tests account for the difference. `bun run typecheck` is clean.

- Windows 11 x64, Bun 1.4.0, Full mode with a real installed Codex integration (v0.154.0), `chatgpt-web/high`, Pro account, automatic browser interaction — exercised with the flag on for 85 tool calls across 25 minutes, and with the flag off for the default retained path.
- macOS: not run.
- Linux: not run.
- Browser-only mode: not run.
- Zero Risk / manual interaction: not run. The flag is deliberately not zeroed for manual mode (Zero Risk already tolerates an absent conversation key), but I have not exercised that combination.
- Packaged launcher builds: not run; no launcher code is touched.


miuuyy:
Thank you for the concrete before/after evidence. I am keeping this as an urgent experimental feature proposal associated with #428, rather than treating it as proof that every retained-turn failure has the same cause.

Source review clarifies the connectorIdentity concern: launcher/electron/browser-host.cjs uses that field for retained-surface identity, while a fresh automatic Full-mode turn independently takes the connector mention/selection path in the browser worker. Withholding the conversation key therefore does not itself disable fresh connector selection.

The remaining acceptance work is the full installed lifecycle with repeated turns and compaction, including cancellation and the explicit manual/browser-only mode boundaries. Resending the complete context on every turn and always taking fresh compaction are material tradeoffs. The feature remains default-off in this proposal; it is not merged or promised for a particular release.

miuuyy:
Thanks for the reproduction and proposal. Fresh browser conversations were supported previously and later removed; I have now restored an opt-in fresh-conversation-per-turn mode locally for the next update. Closing this PR as superseded by that restoration. It is not released yet.

PR #554, opened 2026-09-17
fix: dismiss the Korean ChatGPT rate-limit dialog

twoimo:
## Problem

When the ChatGPT web UI is in Korean, the "too many requests" rate-limit dialog is not recognized, so the turn stalls on the modal instead of dismissing it and returning the structured 429. The handler already covers EN, zh-TW, zh-CN, and ja.

Observed Korean dialog:
- body: `요청을 너무 빠르게 보내고 있습니다...` (also `요청이 너무 많습니다`)
- button: `알겠습니다`

## Fix

Add the Korean strings to the existing rate-limit dialog filters and to the acknowledge button name, mirroring the other locales. No behavior change for the other languages.

## Tests

- Added `the Korean ChatGPT rate-limit dialog is acknowledged and returns a structured 429` to `tests/browser-worker-contract.test.ts` (same fixture as the Japanese case).
- `bun run typecheck` clean.
- `bun run test`: 720 pass, 1 skip, 0 fail (root and `launcher/` dependencies installed per CONTRIBUTING).


miuuyy:
Thank you for the Korean dialog evidence. I included those observed strings in the local rate-limit fix and extended the existing language regression coverage.

There is one intentional difference from this PR: #547 shows that this failure must be non-retryable automatically, so reconnects do not repeatedly submit during the same cooldown. The modal is still dismissed and a structured 429 is returned. Local checks pass; this is not released yet.

miuuyy:
Thank you for the observed Korean dialog strings and regression case. Closing this PR because the Korean coverage has been incorporated into the local rate-limit correction, with credit to your report. The local implementation also makes the structured cooldown failure non-retryable automatically, covering the repeated-submission problem in #547. This is locally verified, still uncommitted and unreleased.

PR #559, opened 2026-09-17
Fix ChatGPT continuation rebinding and bound native polling

Ahmet1991:
Ports two generic stability fixes proven in downstream use: allow assistant-turn rebinding during an active MCP continuation, and cap forwarded write_stdin poll waits so they remain inside the bridge invocation deadline. Also closes the client write half after ordinary broker responses so Windows named-pipe calls can settle cleanly. Includes regression coverage for continuation rebinding and long write_stdin polls.


miuuyy (changes requested):
The DOM rebinding hunk cannot use `activeToolCalls > 0` as ownership proof for a newly observed user message. A targeted fixture with an unrelated new user message and one replacement assistant is accepted with this flag enabled. Separately, updated `acceptedTurnIdentities` are discarded when no assistant-identity replacement is returned.

Keep the new-user guard until the browser continuation is bound to the specific conversation/native invocation. Please separate the write_stdin budget and Windows socket-close changes: they need their own transport evidence and do not establish connector attachment, broker claim or DOM ownership. The Windows close hypothesis still needs a trace showing a valid response frame followed by stalled physical close.

miuuyy:
Thank you for the investigation. Closing this combined proposal because the rebinding change accepts an unrelated new browser user turn whenever a native tool is in flight, which is not proof of ownership. The same-identity reconciliation update is also discarded by the caller. The polling-budget concern is real and separable; the Windows close change still needs the response-frame/physical-close evidence described in the review. These transport concerns do not justify relaxing the browser ownership boundary.

PR #560, opened 2026-09-17
Make Windows package smoke isolated and deterministic

Ahmet1991:
Changes Windows package smoke so it extracts the NSIS application payload into an isolated temp directory instead of installing over the registered live launcher. Smoke mode also skips embedded ChatGPT browser readiness, uses a dedicated 120-second Windows launch budget for runtime materialization, and retries transient Windows cleanup locks. Includes Windows extraction, launcher-startup, and packaging-contract regression tests plus release-validation guidance.\n\nVerification: targeted Windows smoke/browser/packaging tests pass; launcher typecheck passes. The full launcher suite reached 309 pass / 1 fail / 2 skipped on this Windows host; the lone failure is an unrelated EPERM creating a symlink in runtime-host.test.cjs due to local Windows symlink privileges.


miuuyy:
Keeping this open as a non-urgent tooling correction: the current Windows package-smoke path really does execute the NSIS installer and then read the registered installation, so isolating the extracted application payload addresses a concrete problem. This is not a fix for end-user setup or connector failures.

Before acceptance, the staged path needs evidence against the actual distributed NSIS artifact on Windows, with package-only smoke distinguished from browser readiness and installer upgrade/uninstall coverage. The new extraction regression uses a synthetic nested archive, skipping browserHost.ready() also changes smoke coverage on other platforms, and exhausted cleanup failures should remain visible. I have not claimed Windows package acceptance from this macOS review.

miuuyy:
Thanks for working on package-smoke isolation. I am closing this proposal for now because it is not a current priority. The isolation concern is valid, but this patch also relaxes readiness verification beyond Windows, and the archive fixture does not validate the real packaged executable.

PR #565, opened 2026-09-18
Stamp runtime builds and clarify doctor diagnostics

Ahmet1991:
Adds build identity to runtime manifests (commit, dirty flag, build time) and reports it first in doctor output so broken installs can still identify the exact build. Doctor reports now also distinguish locally healthy checks from connector attachment that cannot be proven from the machine, instead of collapsing that state to a plain ready result. Launcher types are kept compatible with the extended report shape.\n\nVerification:\n- build-stamp + doctor-report regression tests: 10/10 pass\n- root TypeScript typecheck: pass\n- launcher TypeScript typecheck: pass\n- runtime bundle build: pass; manifest build stamp verified\n- full root suite: 728 pass / 2 fail; both failures are pre-existing Windows symlink EPERM errors on this host in codex-integration tests, unrelated to these changes


miuuyy:
Thank you for the diagnostics work. I am closing this combined proposal because it expands the build-identity and doctor contracts without repairing the reported setup/connector failures, and I do not want to include that separate diagnostic redesign in the current compatibility work.

There are unresolved semantics in the implementation: a failed git-status lookup becomes dirty: false rather than unknown, and build-stamp metadata is outside the content-derived bundle identity, so reuse of identical runtime contents preserves the previously installed stamp. I confirmed the latter through the actual runtime installer with isolated bundle fixtures. These details need an explicit diagnostic contract before the metadata can be relied on. This closure does not dismiss the value of better diagnostics.

PR #567, opened 2026-09-18
fix: reuse same-thread environment when envelopes are only historical

### LINKED ISSUE #557: [Bug] 5.0.8: mid-turn follow-up with a path-less/unrecognized <environment_context> fails closed ("missing cwd in trusted Codex environment context") despite a valid cached thread environment
## Summary

On 5.0.8 (latest), a **normal** `chatgpt-web/*` turn intermittently aborts mid-turn with:

```
```

The turn runs for minutes, then fails on a **follow-up round** of the same turn. Retrying repeats the failure. This is the same user-visible error as #288/#314 (both closed with a released fix), but it reproduces on 5.0.8 in a case those fixes don't cover: the follow-up request carries an `<environment_context>` from which no cwd/workspace authority can be extracted, and the bridge **fails closed** instead of reusing the thread's cached environment — even though a valid cached authority exists.

## Environment

- Codex Web GPT app: **5.0.8** (latest release; manifest verified `bundleId` intact)
- macOS **26.6.2** (25G83), arm64, ChatGPT **Pro**
- Codex CLI observed in affected rollouts: `0.154.0` and `0.155.0-alpha.2.6`
- Harness: Full
- Model: `chatgpt-web/pro`

## Observed trigger (no model switch, no compaction, no subagent spawn)

Both failing turns are ordinary multi-round turns:

- **Session A**: turn ran ~6 min, many tool rounds. The abort came on the round immediately after two `apply_patch` calls.
- **Session B**: turn ran ~11 min. The abort came on the round immediately after the agent read files **outside** its declared workspace roots (a skills path under the Codex home). A retry of the turn failed identically ~4 min later.

No `context_compacted` event, no in-thread model switch, and no `spawn_agent`/subagent call occurred in either rollout. Affected turn ids: `01a0b094-e057-7092-9108-61825d14b7f5`, `01a0b09c-e8a2-7c22-8d0c-d2981377d2da`.

## Key evidence: cached thread environment existed but was bypassed

`~/.codex-chatgpt-web/runtime/thread-environments.json` contains valid entries for every affected thread (correct `cwd`, `roots`, `writableRoots`, `sandboxPolicy`, fresh `updatedAt`). Example (redacted home):

```json
{
  "cwd": "<worktree>/tzudong",
  "roots": ["<worktree>/tzudong", "<codex-home>/visualizations/.../01a0b09b-..."],
  "writableRoots": ["<worktree>/tzudong", "<codex-home>/visualizations/.../01a0b09b-..."],
  "sandboxPolicy": { "type": "dangerFullAccess" }
}
```

So the store *had* the authority for the thread; it simply was not consulted.

## Suspected root cause (code paths)

1. `src/adapters/chatgpt-web/environment.ts:687` — `uniqueAbsolutePaths()` throws `MissingTrustedCodexEnvironmentError("cwd")` whenever `environmentCwdMatches()` produced zero candidates.
2. `environment.ts:630-672` — `environmentCwdMatches()` is deliberately fail-closed: it returns `[]` for any shape it doesn't recognize (e.g. a path-less environment diff, more than one `<workspace_roots>` section, unmatched open/close tags, >1 `<environment>` without a resolvable primary, etc.). A follow-up round could therefore carry a *legitimate* environment update that yields no cwd.
3. `src/adapters/chatgpt-web/thread-environment.ts:189` — the cached-thread fallback at `:190` is only reached when `hasRawChatGptEnvironmentContext(parsed)` is **false**. That helper (`environment.ts:105-112`) returns true for **any** input message containing `</?environment_context\b`, valid or not. So a malformed / path-less / unsupported current envelope re-throws the original error and the cached per-thread environment is never used.

Net effect: a follow-up environment envelope that carries no usable path authority does not degrade gracefully to the thread's cached authority — it aborts the whole turn.

## Expected behavior

When the current request's `<environment_context>` yields no path authority (or is otherwise an unrecognized shape), don't abort an in-flight turn whose thread already has a trusted cached environment. Options:

- Fall back to `ChatGptThreadEnvironmentStore` authority for that `threadId` instead of re-throwing at `thread-environment.ts:189`; or
- Treat a path-less environment diff as a no-op update and keep the cached authority; or
- At minimum, when failing closed, log the offending envelope shape (or a redacted excerpt) so this is diagnosable without a repro.

The fail-closed *security* intent is understood; the problem is that an update containing no filesystem claims cannot lower authority, yet currently it discards a valid cached one.

## Prior art / relation

- #288 — "Codex 0.150.0-alpha.8 omits `<cwd>` from `<environment_context>` when resuming a thread after a model switch" — **closed, fix released**.
- #314 — "[Agent Spawn] … missing cwd in trusted Codex environment context" — **closed, fix released**.

This report is a **different trigger** on 5.0.8: mid-turn follow-up rounds of ordinary turns (no model switch, no agent spawn), where the cached thread environment already exists.

## Diagnostics

A privacy-safe `Activity → Export safe log` export cannot be produced on this machine right now. If you add a temporary dump of `trustedEnvironmentText(parsed)` (or a redacted form) when `MissingTrustedCodexEnvironmentError` is thrown, I can reproduce and attach the exact offending envelope, which should make the shape unambiguous.

felixlark:
## What this changes

`ChatGptThreadEnvironmentStore.resolve()` now reuses the same thread's cached trusted authority when the request carries environment XML from **earlier turns only** and the native rollout cannot prove anything newer. It changes one fail-closed branch; everything else stays as it was:

- A **current** envelope that fails to parse still throws (`hasRawContext && hasCurrentContext`).
- History never becomes authority itself. It only stops blocking the cache that the same thread already earned.
- History-only requests still cannot use thread-spawn parent inheritance (`if (hasRawContext) throw error` after the same-thread lookup).

Fixes #557, the "follow-up turn fails closed despite a valid cached thread environment" case.

## Evidence

**Reproduction.** The bridge runs on another host than Codex CLI 0.155.0: a Linux server running the packaged 5.0.8 launcher (Full harness), reached from macOS Codex over the network. Codex resends full history each turn, so from the second turn on, `input` contains turn 1's `<environment_context>`. That item carries `internal_chat_message_metadata_passthrough.turn_id` = turn 1, and assistant output follows it, so `hasCurrentChatGptEnvironmentContext` is correctly `false`.

`extractChatGptTurnEnvironment` then throws `MissingTrustedCodexEnvironmentError`, and `resolveCurrentCodexRolloutEnvironment` finds nothing because the rollout lives on the Codex host. `hasRawChatGptEnvironmentContext` is `true` because of the historical item, so the turn fails with `ChatGPT web turn is missing cwd in trusted Codex environment context`. It fails even though `runtime/thread-environments.json` holds the correct entry for that thread (checked).

**Before:** every second turn failed, in both the interactive TUI and `codex exec resume --last`. This reproduced on two independent bridge hosts.

**After** (the same logic applied to the 5.0.8 runtime on both hosts):
- A new turn after the previous one completed: passes.
- A follow-up queued mid-turn: passes.
- `codex exec resume --last`: passes.
- A three-message recall check: the model correctly reports the edits from the earlier turns.

## Tests

The new regression test in `tests/environment.test.ts` covers both a tagged and an untagged historical envelope followed by assistant output:
- It throws with no cached authority.
- It resolves to exactly the cached authority for the same thread.
- It still throws for an unrelated thread.

The test fails on `main` and passes with this change. The existing `does not borrow authority across threads or hide an invalid trusted update` test passes unchanged.

## Scope and invariants

## Verification

- [x] I ran `bun run verify` with the Bun version pinned by `package.json` (1.4.0): 720 pass, 0 fail, relocatable runtime smoke OK.
- [x] I added a focused regression test for this behavior change.
- [x] I manually tested the affected behavior through a real installed Codex integration (Codex CLI 0.155.0 → packaged 5.0.8 runtime on Linux x64, Full harness).

- **Bridge:** Linux x64 (Ubuntu), packaged 5.0.8 launcher, Full harness, on two separate hosts. Tested by applying the equivalent change to the shipped runtime bundle.
- **Codex:** CLI 0.155.0 on macOS arm64, connecting to the remote bridge. Tested with the interactive TUI and `exec resume`.


miuuyy (changes requested):
The historical-cache path needs changes before this can be accepted. In a targeted fixture, historical XML plus cached full access and current read-only metadata returns the old full-access authority. Absence of a local rollout on a remote bridge does not authenticate unchanged permissions.

Recovery needs a current authenticated authority contract. Do not bypass the current-envelope guard by reusing cached authority solely because the XML appears historical. The local steering fix being investigated instead requires equality with the exact current native rollout.

miuuyy:
Thank you for the remote-bridge reproduction. Closing this implementation because it can return cached full-access authority even when the current request declares read-only access; the counterexample in the existing review still applies to this head. Absence of a local rollout cannot authenticate unchanged permissions. The remote-bridge case remains valid to investigate and is not claimed fixed by the separate local native-steering correction.

PR #568, opened 2026-09-18
fix: recover mid-turn environment deltas that omit cwd

### LINKED ISSUE #557: [Bug] 5.0.8: mid-turn follow-up with a path-less/unrecognized <environment_context> fails closed ("missing cwd in trusted Codex environment context") despite a valid cached thread environment

lhz1076277065-netizen:
## Problem

A `chatgpt-web/*` turn dies with:

```
```

Codex retries the **identical** request 5/5 times (`codex_core::responses_retry`), the rejection is deterministic, and the turn is lost. Reproduced on 5.0.8 (latest release), macOS arm64, Full harness, `chatgpt-web/extra-high`. Same user-visible error as #288/#314 but a trigger neither fix covers, and the failure mode described in #557: a valid cached thread authority is bypassed instead of used.

## Root cause

Codex appends an `<environment_context>` delta to the **end** of `input` when the calendar date (or timezone) changes *while a turn is already running*, and that delta omits `<cwd>` and `<shell>`:

```
<environment_context>
  <current_date>2026-09-19</current_date>
  <timezone>Asia/Shanghai</timezone>
  <filesystem><permission_profile type="disabled"><file_system type="unrestricted" /></permission_profile></filesystem>
</environment_context>
```

Real field data — thread `01a0b523-5218-74b3-a033-a55c2eb15db3`, turn `01a0b535-0acd-7423-95c3-eb0ba751395f`. The turn started at 23:49 local and crossed midnight; the delta is item 191 of that turn, i.e. it arrives **after** the turn's tool rounds.

Two consequences chain:

1. **Positional blind spot.** Every claim resolver in `rawEnvironmentText` (`environmentBeforeUser`, `canonicalMetadataEnvironmentBeforeUser` and the backwards skill loop) searches *backwards from the active instruction*, so none can see a trailing delta. Control reaches `if (hasCurrentChatGptEnvironmentContext(parsed)) return undefined;` — the "attempted current update beats older authority" guard — which returns no text. `trustedEnvironmentText` then yields `""` for a raw request, `environmentCwdMatches("")` returns `[]`, and `uniqueAbsolutePaths` raises `MissingTrustedCodexEnvironmentError("cwd")`.
2. **The cached authority is never consulted.** In `ChatGptThreadEnvironmentStore.resolve()`, `historicalMessages` is only computed when a subagent `lineage` exists, so for a **root user thread** `if (hasCurrentContext && !currentCompaction && !historicalMessages) throw error;` re-throws before the rollout lookup and before the same-thread cache at the bottom of the `catch`. `runtime/thread-environments.json` on this machine held a valid, non-expired entry for that exact thread (`cwd`, `roots`, `writableRoots`, `sandboxPolicy`) whose `updatedAt` was **7.06 s before** the failing turn.

## Change

Two independent layers:

1. `rawEnvironmentText` also accepts a current-turn delta that **trails** the active instruction — but only after `environmentMatchesCanonicalMetadata`, so turn/sandbox/workspace metadata must corroborate it. Position no longer hides a genuine update, and user-authored XML still cannot become filesystem authority.
2. `resolve()` answers a delta that names **no filesystem path at all** with the authority the thread already holds. Naming no path means such a delta cannot widen, move or lower authority, so this *reuses* authority rather than discarding it; when no cached entry exists the existing fail-closed paths still throw.

Layer 1 handles deltas that do declare roots; layer 2 covers the path-less shape above, which no root ordering could resolve.

## Safety

Deliberately stricter than the unconditional `return this.get(threadId)` sketched in #557, because that version would also resurrect authority for updates the maintainers reject on purpose. Both reuse points are gated on "the delta names no `cwd` and no `root` element", so all three existing fail-closed expectations still throw, verified by name:

- `does not borrow authority across threads or hide an invalid trusted update`
- `a malformed current update is not replaced by a valid older transcript envelope`
- `rollout recovery rejects outside paths and never repairs malformed raw authority`

The second layer additionally declines to reuse when the delta's declared sandbox type disagrees with the cached policy, so it cannot mask a genuine downgrade.

## Testing

New cases in `tests/environment.test.ts`, inside `describe("trusted Codex task environment continuity")`, using the existing `currentWire()` idiom:

| test | pristine `v5.0.8` | with this change |
| --- | --- | --- |
| `reuses the thread's cached authority for a date-rollover delta appended mid-turn` | **RED** — raises the production `MissingTrustedCodexEnvironmentError("cwd")` | GREEN (layer 1) |
| `recovers a path-less mid-turn delta from the thread's cached authority` | **RED** | GREEN (layer 2) |
| `fails closed for a path-less mid-turn delta when the thread has no cached authority` | GREEN | GREEN (guard, unchanged) |

- `bun test ./tests/environment.test.ts` — **57 pass, 1 skip, 0 fail** (pristine `v5.0.8`: 54 pass, 1 skip, 0 fail).
- `bun test ./tests` — **719 pass, 1 skip, 3 fail**; the 3 failures are identical to the pristine `v5.0.8` baseline (`Cannot find package 'electron'` from `launcher/electron/browser-host.cjs` via `tests/zero-risk-adapter.test.ts`, because a root `bun install` does not install the `launcher/` workspace). Verified by failing-test name set, not counts.
- `bun x tsc --noEmit` — clean.
- Bundled Bun 1.4.0 (the repo-pinned version) used throughout; no system toolchain changes.

**Live end-to-end A/B** on the reporter's machine, driving `POST /v1/responses` against a real 5.0.8 install with a request reconstructed from the failing turn's geometry, for a thread holding a valid cached authority:

| runtime | result |
| --- | --- |
| official 5.0.8 (`app/cli.js` sha256 `0697c474…`) | `response.failed` — `ChatGPT web turn is missing cwd in trusted Codex environment context` (2 independent runs) |
| this patch | `response.completed` — real ChatGPT web turn streamed `E2E OK`, usage 9006 in / 16 out |

## Notes for review

- I did not make `environmentCwdMatches` treat `<environment_context>` (singular) as recoverable by taking `roots[0]` as the cwd: that infers a working directory from root ordering, whereas cached/rollout authority is already proven. Worth flagging separately though: that fallback exists today for the adjacent case, so an envelope whose only root lies outside the real workspace gets that root adopted as `cwd` — and my first two synthetic attempts at this bug passed on unmodified code precisely because of it. I left that behaviour untouched as out of scope here; happy to open a separate issue if useful.
- Adds a deterministic trigger (midnight / date-change deltas) that #557 did not identify, and fixes the bypass #557 describes.

Closes #557


miuuyy (changes requested):
The pathless-delta acceptance is too broad. Targeted fixtures with contradictory sandbox declarations or an external permission profile still reuse cached full access. Missing cwd/roots does not establish that permissions are unchanged.

Please preserve fail-closed behavior unless the current delta is authenticated and its complete permission semantics are understood. The trailing-delta fixtures also use an active instruction from an older turn than the current metadata, which does not prove the same-turn native lifecycle claimed here.

miuuyy:
Thank you for investigating the midnight/delta case. Closing this implementation because its pathless-delta classification reuses cached full access for contradictory sandbox declarations and an external permission profile. A delta without cwd or roots can still change permission semantics. The counterexamples in the existing review remain applicable. The reported midnight case stays distinct from the separately verified native-steering fix; this closure does not declare that report resolved.

lhz1076277065-netizen:
Both review points were correct — thank you. Pushed `9e3dee8` addressing them.

**1. The pathless classification was too broad, and it was exploitable exactly as you described.** My predicate returned `sandboxType: undefined` when no recognized sandbox mode was stated, and `resolve()` then guarded the comparison with `if (delta.sandboxType && ...)`, so the check was skipped entirely. Confirmed by running the new fixtures against `17646ab`:

- a delta declaring `<permission_profile type="external">` **did** reuse cached `dangerFullAccess`
- a delta stating no sandbox at all **did** reuse it

Both now fail closed: `pathlessChatGptEnvironmentDelta` returns `undefined` unless every current envelope states one recognized sandbox mode, and `resolve()` requires `delta.sandboxType === cached.sandboxPolicy.type`. Absent, unrecognised or differing declarations take the original throw path. A third fixture (explicit narrower `read-only` profile) already failed closed before this commit, so I am not claiming it as a second hole.

**2. The trailing-delta fixtures were not same-turn.** They reused an active instruction tagged with an older turn than the current metadata, which does not match the lifecycle I claimed. Rewritten to the observed geometry: the turn's own instruction and tool rounds carry the current `turn_id`, and the delta is appended after them — matching the field case, where the session contains exactly two `environment_context` messages (turn one's, and the midnight delta) and the delta is item 191 of its own turn. Added `keeps using the turn's own start envelope when a delta trails its tool rounds` to pin the unaffected adjacent path.

`tests/environment.test.ts` 61 pass / 0 fail; full suite unchanged apart from the new cases; `tsc --noEmit` clean.

The midnight trigger itself is still unreleased on your side, so I am holding a local build of this branch to keep my own Codex usable. Happy to reopen here or resubmit as a fresh PR against whatever shape you prefer — including a version restricted to layer 1 only (trailing-delta acceptance behind canonical-metadata corroboration), which needs no cached-authority reuse at all for the field case where the delta still declares its roots.

PR #569, opened 2026-09-18
fix: expose deferred tool search in inventory

Safa675:
`codex_tool_inventory` returned an empty result when a query matched only tools that Codex exposes through deferred `tool_search`. ChatGPT could then conclude that subagents or extra MCP tools were unavailable even though Codex could load them.

This change returns the visible deferred `tool_search` entry as a fallback when a filtered inventory query has no direct or gateway matches. Existing direct and nested matches keep their current behavior, and unfiltered inventory remains unchanged.

Validation:

- `bun test tests/chatgpt-web-harness.test.ts` — 83 pass, 0 fail
- `bun run verify` — 719 pass, 1 skip; launcher 308 pass; typechecks, build, and relocatable runtime smoke passed


miuuyy:
Thank you — I reproduced the deferred-discovery gap. I implemented a local variant that keeps `tools`, `total` and pagination truthful: when a filtered search has no matches, actual visible native discovery entries are returned separately as `discovery_tools`. No tool is called automatically.

The MCP stdio regression checks the empty match set, discovery metadata, invocation through the broker, and unchanged matching results. This is local and unreleased; your report/proposal is credited in the investigation record. I have not merged this PR as-is.

miuuyy:
Thank you for identifying and testing the deferred-discovery gap. Closing this PR because the correction is already implemented locally, with your contribution acknowledged. The local implementation returns native discovery entries separately in `discovery_tools`, preserving the meaning of query matches, totals, and pagination. It has regression coverage but is still uncommitted and unreleased; this is not a claim that users already have the fix.

PR #577, opened 2026-09-19
fix: retry a timed-out same-page rebind before failing an accepted turn

felixlark:
## What this changes

When a stalled DOM probe triggers a same-page rebind and that rebind **times out** (`ChatGPT browser stage timed out: response_page_rebind_N`, or `Could not connect Playwright to the launcher browser: … connectOverCDP: Timeout`), the accepted turn now gets one more bounded connection attempt before it fails.

- The retry happens only after the timed-out attempt has **fully settled**. It uses the existing `runStage(..., awaitAbortedActionSettlement = true)` path, and any connection that attempt produced is closed first. Two CDP connections never contend for the same stalled page, which the surrounding code comments warn against.
- It is bounded by `MAX_CHATGPT_PAGE_REBIND_CONNECT_ATTEMPTS = 2` per rebind. The existing `MAX_CHATGPT_BROWSER_PAGE_REBINDS` budget is unchanged.
- Every other rebind failure stays terminal: viewport failures, duplicate ownership, aborted turns, and so on.
- The retry is logged (`is retrying its same-page rebind (attempt 2 of 2) after: …`) so it can be diagnosed.

## Evidence

This is from launcher logs on two Linux hosts running the packaged 5.0.8 launcher with the Full harness:

- **Successful rebinds** completed in 3.6 s, 3.8 s, 12.0 s, 12.5 s and 35.7 s. A stalled page usually comes back.
- **Failed turns** all timed out at exactly 60 s inside `response_page_rebind_1`, with no second chance. Each of those turns had already been accepted (`submission accepted evidence=generation_running`), and ChatGPT kept working throughout the rebind. For example, the broker received valid claims and even delivered a tool call during the 60 s window. Codex then shows `ChatGPT stopped responding after the task started` or `ChatGPT did not confirm that the prompt was sent`.
- The failures are not explained by prompt size. On the same host, inline single messages of 158K, 219K, 280K and 360K characters all completed in 17–21 s, including a larger message than one of the failing turns (310K).

This does not claim to fix pages that stay frozen for many minutes. It adds one more bounded chance for the common case where the page recovers shortly after the first budget.

## Tests

New tests in `tests/browser-worker-contract.test.ts` cover `retryTimedOutChatGptPageRebind` and `isRetryableChatGptPageRebindFailure`. They check that:
- a timed-out rebind is retried once and then succeeds;
- a CDP connect timeout is retryable;
- other failures are never retried;
- the retry budget is bounded;
- an aborted turn is never retried.

## Scope and invariants

- [x] Model, route, effort, connector, and capability selection remain explicit with no silent fallback or false-success path: a turn still fails when the page never comes back.

## Verification

- [x] I ran `bun run verify` with the Bun version pinned by `package.json` (1.4.0): 722 pass, 0 fail, relocatable runtime smoke OK.
- [x] I added focused regression tests for this behavior change.
- [x] I manually tested through a real installed Codex integration: Codex CLI 0.155.0 and Codex Desktop, connected to the packaged 5.0.8 runtime on Linux x64 with the equivalent change applied.

- **Bridge:** Linux x64 (Ubuntu), packaged 5.0.8 launcher, Full harness, Pro account.
- **Not yet observed:** a live retry in production. The failure is intermittent; the patched runtime is deployed and logs every retry.


miuuyy:
Thanks for investigating the rebind failures. I am closing this implementation: its retry predicate also accepts connection-refused and protocol failures under the same message prefix, while timing out does not cancel the underlying operation. Another retry can therefore leave the original work unresolved. Any follow-up fix needs a bounded connection lifecycle and evidence for which failures are actually transient.

PR #578, opened 2026-09-19
fix: keep Codex plan block markers verbatim in converted Markdown

felixlark:
## What this changes

In plan mode, Codex renders a plan card with its **Implement plan** action only when the assistant text contains `<proposed_plan>` and `</proposed_plan>` verbatim on their own lines. Codex's `TaggedLineParser` trims the line and then compares it exactly.

ChatGPT renders those markers as literal text (`<p>&lt;proposed_plan&gt;</p>`). `chatGptHtmlToMarkdown` then passes them through Turndown, which escapes the underscore. Codex therefore receives `<proposed\_plan>` … `</proposed\_plan>`, shows the tags as plain text, and never offers Implement plan.

The fix adds `preserveCodexPlanBlockTags`, which restores the underscore only on standalone marker lines, with optional surrounding whitespace to match Codex's trimming. It follows the precedent of `preserveObsidianWikiLinks`, which already undoes a Turndown escape that Codex misreads. Nothing else changes:
- Other escapes such as `snake\_case` stay escaped.
- A marker mentioned mid-sentence is not rewritten.

## Evidence

The report came from a real Codex Desktop plan-mode turn (`collaboration_mode = plan`, model `chatgpt-web/pro`) through the packaged 5.0.8 runtime. The rollout's final `task_complete.last_agent_message` began with `<proposed\_plan>` and contained `</proposed\_plan>`. The desktop app showed the tags as text and offered no Implement action.

With the equivalent change applied to the packaged runtime, the same marker block comes back as `<proposed_plan>` … `</proposed_plan>`. The body is unchanged: `snake\_case` stays escaped, and an indented closing marker after a list item is accepted because Codex trims the line.

Note for packaging: the conversion runs in `browser-helper.cjs` (the browser helper bundle). Both bundles are built from `src/adapters/chatgpt-web/markdown.ts`, so the source change covers both.

## Tests

`tests/chatgpt-web-markdown.test.ts` has two new tests:
- Standalone opening and closing markers are restored, while a `snake_case` word in the plan body keeps its escape.
- A marker mentioned inside a sentence is left escaped.

The first test fails on `main` and passes with this change.

## Scope and invariants

## Verification

- [x] I ran `bun run verify` with the Bun version pinned by `package.json` (1.4.0): 739 pass, 0 fail, relocatable runtime smoke OK.
- [x] I added focused regression tests for this behavior change.
- [x] I manually tested through a real installed Codex integration: Codex CLI 0.155.0 connected to the packaged 5.0.8 runtime on Linux x64, with the equivalent change applied.

- **Bridge:** Linux x64 (Ubuntu), packaged 5.0.8 launcher, Full harness, Pro account.
- **Codex:** CLI 0.155.0 and Codex Desktop on macOS arm64.


miuuyy:
Thanks for identifying the plan-marker problem. An independent fix is already implemented locally for the next update: it preserves the markers during paragraph conversion without rewriting literal examples inside fenced code blocks. Closing this PR as superseded; the local fix is not released yet.

PR #579, opened 2026-09-19
fix: do not rebind a stalled DOM probe while the accepted turn is provably alive

felixlark:
## What this changes

When a DOM observation probe times out after a submission was accepted, the worker currently rebinds at once. It closes the turn's CDP transport and reconnects the whole launcher browser (`waitForNewAssistantTurn` and the response-observation loop). The existing liveness fallback (`chatGptExternalProgressIsLive` → `waitForTurnDomOrExternalProgress`) only handles other errors, so broker activity never prevents that teardown.

With this change, a probe timeout does **not** rebind while the turn is provably alive, meaning either:
- broker activity suppresses DOM health (`chatGptExternalProgressSuppressesDomHealth`), or
- the submission was accepted within the DOM grace window. ChatGPT already reported `generation_running`, and the page is often busy rendering a large prompt right after that.

Instead the worker waits for new external progress or a short backoff (1 s, 2 s, 5 s) and re-probes the same page. This wait does not touch the page, so no further evaluates pile up in a busy renderer. Rebinding still happens, unchanged, once the page stays unresponsive past the grace window with no broker activity.

## Evidence

Production launcher logs (packaged 5.0.8, Linux x64, Full harness) show turns failing with *"ChatGPT stopped responding after the task started"* in two cases:
- A 5 s probe timeout came about 5 s after `submission accepted evidence=generation_running`, and the forced rebind's `connectOverCDP` then timed out at 60 s.
- The failing rebind happened while the broker was still receiving **valid** claims and even delivering tool calls for that same turn.

Rebinds that succeeded historically recovered within 3.6–35.7 s, so these pages were busy, not dead.

## Tests

In `tests/browser-worker-contract.test.ts`:
- **New:** a probe timeout while a tool batch is in flight waits and re-probes the same page with 0 rebinds.
- **New:** a probe timeout right after acceptance, with no broker activity yet, does the same.
- **Adjusted:** the existing "rebinds the missing assistant observation" test now passes `graceMs = 0`, so it still covers the case where a rebind is required.

## Scope and invariants

## Verification

- [x] I ran `bun run verify` with the Bun version pinned by `package.json` (1.4.0): 739 pass, 0 fail, relocatable runtime smoke OK.


miuuyy:
Thanks for the investigation. I am closing this implementation because a probe timeout does not cancel the underlying evaluate call; continuing to issue new probes can leave multiple evaluations pending on the same transport. The liveness distinction is useful, but this patch does not establish bounded probe ownership.

PR #580, opened 2026-09-19
fix: bound per-page ownership inspection during launcher page selection

felixlark:
## What this changes

`selectLauncherPage` reads native target ownership by attaching a CDP session to **every** page in the launcher browser inside one `Promise.all`, with no per-page bound. If another turn's renderer is busy, for example while inserting a very large prompt, its `newCDPSession` can take a long time. That holds up selection for an unrelated, responsive tab and makes its rebind stall.

Each candidate now gets its own inspection bound (`LAUNCHER_PAGE_INSPECTION_TIMEOUT_MS`, 2 s):
- A page that cannot answer in time is treated as *not owned on this pass*, and the existing loop retries until the overall deadline. Ownership rules are unchanged.
- A session that arrives after its inspection expired is detached immediately.

A follow-up that splits the rebind connect budget depends on #577 and is not included here.

## Evidence

Stalled rebinds in production coincided with a second tab of the same launcher loading a roughly 290K-character prompt.

## Tests

`tests/launcher-browser-host.test.ts`: an unrelated page whose CDP session only resolves after 200 ms no longer delays selection of the owned page (done in under 150 ms with a 30 ms bound), and the late session is detached.

## Scope and invariants

## Verification

- [x] I ran `bun run verify` with the Bun version pinned by `package.json` (1.4.0): 738 pass, 0 fail, relocatable runtime smoke OK.


miuuyy:
Thanks for investigating the cross-tab stall. I am not adopting automatic per-page inspection timeouts, and have also removed my separate local deadline experiment. The existing user-controlled recovery is to close the affected task tab in the launcher, which requests cancellation of its owned turn. Closing this PR as a design decision; this does not claim that the underlying cross-tab contention has been fixed.

PR #581, opened 2026-09-19
fix: retire browser turns whose tool batch Codex never answered

felixlark:
## What this changes

An automatic browser turn deliberately survives a client disconnect so that Codex can reconnect (`index.ts`, the AbortError branch). But once a tool batch is outstanding and Codex never returns the results:
- `externalToolCallsInFlight` blocks completion with no age limit;
- `turnTimeoutMs` is unset by default;
- the launcher heartbeat is a plain interval;
- `ChatGptTurnSessions.prune()` skips active sessions.

The turn can therefore keep a tab and heartbeat forever.

This change records when a tool batch was handed to Codex (`ChatGptTurnSession.outstandingSince()`). `ChatGptTurnSessions.reapDetachedToolTurns()` retires an active turn only when both of these hold:
1. The batch has been outstanding longer than `CHATGPT_DETACHED_TOOL_RESULT_TIMEOUT_MS` (30 min).
2. The session has seen no client activity for that long. Every tool-result request touches the session.

Retirement goes through the existing `retire()` path, which cancels the runtime and releases the tab. The reaper runs from `prune()` and from a one-minute, unref'd interval.

A turn whose client is still around is never reaped, because any request resets its activity.

## Evidence

A production turn kept emitting `browser.turn_heartbeat` for 2+ hours after checkpoint `response-stalled-60s`. It occupied a launcher tab until the service was restarted.

## Tests

`tests/chatgpt-web-harness.test.ts`: an abandoned active turn is reaped after the timeout. A turn with recent client activity and an idle turn are not reaped. A batch whose results were delivered clears `outstandingSince()` and is never reaped.

## Scope and invariants

## Verification

- [x] I ran `bun run verify` with the Bun version pinned by `package.json` (1.4.0): 738 pass, 0 fail, relocatable runtime smoke OK.


miuuyy:
Thanks for investigating abandoned turns. I am closing this implementation: 30 minutes without a new tool-result request does not establish that Codex detached. A legitimate long-running tool or approval can be silent for that period and would be cancelled by this reaper. Retirement needs explicit lifecycle evidence rather than elapsed client inactivity.

PR #582, opened 2026-09-19
fix: prevent Windows titlebar overlap

berkeserce:
Fixes the Windows launcher browser chrome overlapping the native titlebar controls.

- Aligns the browser tab strip with the 46px Windows titlebar overlay
- Reserves space for minimize, maximize, and close controls
- Extends the separator cleanly across the toolbar
- Adds a focused renderer regression test

Validated on Windows. Renderer wiring tests: 23/23 passing.


miuuyy:
Thanks for reporting this. I have not reproduced the overlap on a Windows runtime, and I am closing this patch for now. There is a real height mismatch in the source, but the proposed fixed padding does not cover narrow-window tab overflow or scaling between the native titlebar and CSS. A future focused fix should include native Windows validation at those sizes and zoom levels.

PR #589, opened 2026-09-19
fix: defer GFM email history scans when rendering large messages

GordeyTsy:
## What this changes

Prevents a reproduced post-send renderer stall caused by repeated history scans in ChatGPT’s bundled GFM email-autolink tokenizer.

The tokenizer scans accumulated Markdown events before determining whether a candidate contains `@`. Large escaped link labels can make these unsuccessful attempts quadratic. This patch defers the scan until `@` is encountered, preserving the event prefix from before the attempt.

The worker applies the change through an owned CDP breakpoint before the relevant module initializes, then detaches the debugger. It reinstalls the hook before a connector-refresh reload and releases it during cleanup. Unknown, ambiguous, or already-patched signatures are left unchanged.

## Evidence

Earlier tests used synthetic messages on an installed Linux launcher with Electron 41.10.7 / Chromium 146:

- **Before:** an approximately 198 KB code-rich JSON envelope left the renderer unresponsive for over 60 seconds after submission. A separate CPU profile attributed approximately 8.7 of 11.36 seconds to the email-autolink scanning path.
- **After:** with this patch and the original single native insertion, the user-turn detection wait took 488 ms, the following responsiveness probe took 20 ms, and an answer arrived.
- The stall also reproduced with single native insertion, establishing that it was independent of chunking.

These are individual observations, not statistical estimates. They concern local rendering, not service response latency.

The observed editor was ProseMirror; the profiled tokenizer belonged to the site’s `conversation-small` asset. This patch does not change composer selectors or insertion behavior. Separate experimental insertion changes are excluded because their broader integrity tests remain unresolved.

## Scope and invariants

- [x] Read and followed `CONTRIBUTING.md`.
- [x] Focused change without unrelated cleanup, release artifacts, or version changes.
- [x] Prompt content, model, route, effort, connector, and capability selection are unchanged.
- [x] No silent model fallback, quota bypass, sandbox change, or web-security change.
- [x] No browser state, credentials, Tunnel IDs, raw logs, generated artifacts, or private paths are committed.

The two added development dependencies provide the actual micromark implementation used by the regression tests. They are not runtime dependencies.

## Verification

- [x] Frozen dependency installation in the repository root and `launcher/`.
- [x] `bun run verify` with project-pinned Bun 1.4.0.
- [x] Focused regression tests covering complete rendered-output equality for email, URL, Unicode, nested-label and long-label cases, plus 250 deterministic generated cases.
- [x] Tests reject unknown, ambiguous, and already-patched bundle signatures.
- [x] Repeat live-account validation of the complete PR branch against current upstream.

Local verification against upstream `eaf4f09` passed:

- 739 core tests; 2 skipped.
- 310 launcher tests.
- Both TypeScript checks and dependency audits.
- Launcher renderer build, runtime bundle, license generation, and relocatable-runtime smoke test.

The focused parser tests require no ChatGPT account:

```bash
bun test tests/autolink-render-compat.test.ts
```


miuuyy:
Thanks for the profiling and reproduction details. Rewriting ChatGPT's bundled JavaScript through debugger hooks depends on private site internals and is too fragile for us to ship and maintain. The performance finding is useful, but we will not adopt this patching mechanism. Closing the PR.

PR #590, opened 2026-09-19
Opt-in local Grok CLI compaction, so checkpoints stop costing a ChatGPT message

ioio101:
## What this changes

Adds an opt-in backend where a local Grok CLI writes the Codex compaction checkpoint in place of the retained ChatGPT chat. It's off by default, so nothing changes unless you turn it on.

### Why

1. Compaction currently costs a message. When a task hits the threshold, the bridge types a summarize request into the live chat. That's one message from the account (a Pro message for us) every time, spent on bookkeeping.
2. Codex already has the full history locally, so any LLM can write that summary. Plenty of people have another LLM CLI lying around, and using it seems better to us than wasting a ChatGPT message. ChatGPT Web is still the task model. Grok only writes the checkpoint.
3. The checkpoint no longer goes through the ChatGPT composer, so a task doesn't have to stay under what one message can carry. With the second commit we ran a single chat to about 240k tokens without having to worry about it.

### How it works

- Turn it on with `grokCompaction.enabled` in config or `CODEX_CHATGPT_WEB_GROK_COMPACTION=1`.
- On a compaction request (v1 and v2) the history is written to one prompt file under the same COMPACT_PROMPT contract. `grok` runs headless with tools off, and its stdout goes back to Codex as the usual native compaction item.
- A Web response that is still running is settled the way it is today: pending tool results are delivered and the next tool call gets a stop instruction. That doesn't cost a message.
- The old chat is retired and the next turn opens a fresh Temporary Chat from the compacted history, same as after a ChatGPT checkpoint. No summarize message is typed into ChatGPT.
- It fails closed. The CLI is looked up before the retained chat is touched, so a machine without `grok` keeps its live chat. A non-zero exit, a timeout, or an empty or very short reply is a terminal compaction error. The same-chat ChatGPT handoff is used only if `allowChatGptFallback` is set.
- The 104k/95k send preflight is unchanged.

### Two commits

1. The Grok backend. This is the part we care about most.
2. Optional `autoCompactTokenLimit`. It moves the point where Codex compacts the task on automatic Sol reasoning efforts and nothing else. The one-message boundary and its preflight stay where they are, and context is not split into extra messages. Instant keeps its measured window (as far as we could find, Pro gives Instant 128K total context and reasoning chats 400K, but we couldn't load OpenAI's own page to confirm). Config rejects it without `grokCompaction`, with Bigger Context, or in Zero Risk. If you only want the first commit, drop this one.

### Other CLIs

The approach isn't specific to Grok. Any CLI that takes a prompt and prints a reply would do: Claude Code, Gemini CLI, Codex itself, or a local model behind Ollama. `command` and `args` are already configurable with a `{prompt_file}` placeholder, so a CLI that accepts a prompt file can be set up through config today. One that wants the prompt on stdin would need a few lines to pipe the file in. We've only tested Grok, which is why the names and defaults are Grok's. We're happy to rename it to something neutral like `localCompaction`.

We know CONTRIBUTING asks for a discussion before a feature like this. We built it for ourselves first and wanted to show up with working code and numbers. If you'd prefer it go through Ideas first, that's fine with us.

## Evidence

Before: every automatic compaction typed a summarize message into the retained chat and used one message from our Pro account.

After, on our real signed-in Pro session with a real installed Codex:

Short check. Codex showed "Context compacted" after 17s, and a follow-up asking for the first message of the thread was answered correctly. Launcher diagnostics show no browser turn during the compaction and a fresh Temporary Chat for the next turn.

Long run. `codex exec -m chatgpt-web/high` with a read-only sandbox on a private reverse engineering repo, `autoCompactTokenLimit` set to 240000:

| | |
|---|---|
| Tool calls | 112 |
| Chat 1 | grew from 45k to 239k tokens inside one ChatGPT response, no send errors |
| Compaction | requested at 239,074 tokens, next tool call stopped 13s later, response ended normally |
| Grok | 755,460 characters in with nothing trimmed, 15.9k-character checkpoint out, 3m25s |
| Chat 2 | opened with a 21k-token message, ran to about 227k, finished the task |
| Total | 2 ChatGPT messages, 1 compaction |

By our estimate the stock 95k threshold would have compacted around five times on the same task, each time with its own summarize message.

We then asked chat 2 for details that only existed before the compaction: a build string, a PE timestamp, an md5, and one line quoted verbatim. All four came back correct.

The long run turned up three problems, fixed in this PR:

- The default timeout was too tight for a history that size. It's now 15 minutes, and the adapter deadline follows the configured timeout.
- Grok mentioned its own working directory in the checkpoint. It now runs in an empty temp directory and the prompt tells it not to.
- The raised threshold was being applied to Instant. It now applies to reasoning efforts only.

Those fixes are covered by unit tests. We haven't repeated the long live run since making them.

Not tested: recall from deep inside a single 200k+ chat with no checkpoint in between, and a cold rebuild above the one-message boundary (it still fails with the existing "run /compact" error and doesn't split).

## Scope and invariants

- [ ] This is a small, focused change with no unrelated cleanup or generated rewrite. (One feature and no unrelated cleanup, but it isn't small: about 1,400 lines, of which roughly 460 are tests and 40 are docs.)
- [x] The change stays focused on ChatGPT web-backed Codex models; it does not add a generic provider or unrelated product surface. (Grok is not exposed as a model or provider. It only writes the checkpoint.)
- [x] Model, route, effort, connector, and capability selection remain explicit with no silent fallback or false-success path. (Fails closed. The ChatGPT fallback needs an explicit flag and logs a warning when used.)
- [x] Terms and trademark claims remain factual; this change is not marketed as a quota or rate-limit bypass. (Task messages still go through ChatGPT as before and the send preflight is untouched. The only thing that moves is who writes the summary.)

## Verification

- [ ] I ran `bun run verify` with the Bun version pinned by `package.json`. (We ran every verify step separately with Bun 1.4.0. `check-version`, both audits, both typechecks and the launcher build pass. `bun run test` is 749 pass, 2 fail and `launcher:test` is 306 pass, 1 fail. All three failures are `EPERM` on `symlinkSync` in tests this PR doesn't touch, because our Windows account can't create symlinks. That makes `bun run verify` stop early on our machine, so we didn't get to the runtime bundle and release smoke steps and are leaving this box unchecked.)
- [x] I added or updated a focused regression test for behavior changes. (`tests/grok-compaction.test.ts`, 14 tests, including an adapter test that fails if any browser message is sent during a Grok compact.)
- [ ] If this changes ChatGPT browser UI handling, I included observed DOM evidence and a reproducible fixture instead of broadening selectors speculatively. (Not applicable, no browser UI handling changed.)
- [ ] If this changes the launcher, I preserved macOS, Windows, and Linux packaging and named the platform packages actually built below. (Not applicable, launcher untouched.)

- Windows 11 Pro, ChatGPT Pro account, Full mode, automatic interaction.
- Launcher run from source (unpackaged Electron from this branch) on our normal signed-in profile, with the Codex CLI pointed at the bridge.
- Grok CLI on its default model.
- Efforts exercised live: `chatgpt-web/high`. Others: not run.
- Zero Risk: not run live, and the local Zero Risk settle path has no dedicated test yet.
- Browser-only mode: not run.
- DEV chat `compaction-lab`: not run.
- macOS and Linux: not run.
- Packaged builds: not built.


ioio101:
I added this locally since the current compaction method spends an entire message to compact, I'm having a great experience with it compared to previously. I now have 170 messages a day with 5.6 sol pro on web instead of spending half of it on compactions. (Assuming that what OpenAI claims is factual; $200 plan getting 170 5.6 sol pro messages/day) and my chats lasts a lot longer  before compacting with no problems(240k context!). Haven't done a specific 'needle' test, but it has continued properly and completed tasks for me, so I'm deciding to share with you in case you want to take the methodology and apply it to your project! I had Fable 5.1 generate a PR in case you wanted the idea.

This could be optimized further and other CLI's can easily be added.

miuuyy:
Thanks for sharing your implementation and experience. An additional Grok CLI backend for compaction is outside the scope of this project, so I am closing this proposal.

Issue #546, opened 2026-09-17
[Bug] Selected model is at capacity. Please try a different model.


mikebilly:
### Codex Web GPT version

Windows · v5.0.8

### Codex version

I use Codex extension in Vscode

### Platform

Windows x64

### OS version

Windows 11

### ChatGPT account tier

Free

### Integration mode

Browser-only

### Selected model

ChatGPT Web - Luna Light

### What happened?

<img width="2559" height="1384" alt="Image" src="https://github.com/user-attachments/assets/ac79f76e-3350-4a57-9a06-09f9dcb33c20" />

### Minimal reproduction

I run from source.

### Complete final error

```text
Selected model is at capacity. Please try a different model.

ChatGPT web current user message conflicts with native Codex turn_id metadata
```

### Privacy-safe diagnostics

[safe_log.txt](https://github.com/user-attachments/files/32325327/safe_log.txt)

### Other Codex wrappers or route owners

none

### Preflight checklist

- [x] I tested the latest release and, for catalog or turn failures, fully restarted Codex including its background process.
- [x] I stated whether the failure also occurs in a fresh Codex task, or explained why that does not apply.
- [x] I attached only a privacy-safe export, or explained why the exporter could not produce one; I removed any remaining private content.
- [x] I searched existing issues for this exact final error.

Kazanovasgh:
The same

devopsaptlife:
the same

ChatGPT web current user message conflicts with native Codex turn_id metadata

lvtong199881:
THE SAME!!! 
FIX!!! PLEASE!!!
GOD BLESS YOU!!!

miuuyy:
The actionable error here is `current user message conflicts with native Codex turn_id metadata`; Codex's "at capacity" banner does not establish a ChatGPT quota problem.

Please give the exact Codex VS Code extension version and source commit you ran, then reproduce once after fully restarting VS Code/Codex. State whether it also happens in a brand-new task. Attach a fresh Activity → Export safe log with the failure time/timezone and the matching detailed Codex Output/error entry. We already have the original safe_log.txt; the needed addition is the native error/turn correlation, not the same file again. Please redact message contents, tokens, credentials and private paths.

lvtong199881:
你好，我已收到您的邮件，并会尽快给您回复。

devopsaptlife:
i have the latest version of codex and your app for now

browser · turn ended
[...]

Issue #410, opened 2026-09-09
[Bug] codex with /goal never work


jessecmc:
### Codex Web GPT version

lastest

### Codex version

lastest

### Platform

macOS arm64

### OS version

mac os

### ChatGPT account tier

Business / Enterprise / Edu

### Integration mode

Full harness (MCP)

### Selected model

chat gpt web high

### What happened?

[Bug] codex with /goal never work

### Minimal reproduction

[Bug] codex with /goal never work

### Complete final error

```text
[Bug] codex with /goal never work
```

### Privacy-safe diagnostics

[Bug] codex with /goal never work

### Other Codex wrappers or route owners

[Bug] codex with /goal never work

### Preflight checklist

- [x] I tested the latest release and, for catalog or turn failures, fully restarted Codex including its background process.
- [x] I stated whether the failure also occurs in a fresh Codex task, or explained why that does not apply.
- [x] I attached only a privacy-safe export, or explained why the exporter could not produce one; I removed any remaining private content.
- [x] I searched existing issues for this exact final error.

miuuyy:
Could you add one concrete example of the `/goal` failure? At the moment the report repeats the title but contains no error or diagnostic export.

Please include the exact launcher and Codex versions, the selected Web model, what you expected the goal to do, what actually happened, and the complete final error or response. Say whether the same simple request works in a fresh task without `/goal`.

Immediately after the failure, use **Launcher → Activity → Export safe log** and drag the export into a reply here; ZIP it if needed. Include the time/timezone, and redact private prompt content from screenshots. I need that to distinguish a goal-lifecycle failure from missing tools or a browser error.

miuuyy:
Closing this report as inactive while waiting for a concrete goal failure with exact versions and a diagnostic export, requested on September 11. The evidence already supplied remains recorded here; this closure does not mean the reported failure was fixed or dismissed.

Please reopen this issue, or open a new one linking here, when that specific evidence is available. Include the exact running versions, failure time/timezone and Launcher → Activity → Export safe log. There is no need to resend unrelated full task histories, credentials or configuration files.

miuuyy:
Closing pending a concrete reproduction. The September 11 request for exact versions, the actual goal failure and an Activity safe export still has no reply; the report currently repeats its title without an error or diagnostic evidence. This is not being marked fixed.

Please reopen or create a new issue linking here when those details are available. Include the failure time/timezone and whether the same request works without /goal.

PR #593, opened 2026-09-20
fix: accept authenticated cross-thread follow-ups

nofihq:
## What

Recognize the exact current-turn `send_message_to_thread` result that Codex Desktop uses for cross-thread follow-ups as the active instruction revision.

The accepted shape remains narrow:

- `type` is `function_call_output`
- `name` is `send_message_to_thread`
- the item has an ID
- its native `turn_id` equals the active turn
- its output is a non-empty, well-formed `codex_delegation` envelope with a source thread ID

Malformed envelopes, another tool name, another turn ID, and context-only restart recovery continue to fail closed.

## Why

I reproduced the P1 `ChatGPT web current user message conflicts with native Codex turn_id metadata` failure on Linux with v5.0.8 Full harness and ChatGPT Web Pro.

After an ordinary task completed, Codex Desktop's `send_message_to_thread` follow-up created a new native turn. The follow-up instruction was present in the canonical rollout as a current-turn `function_call_output` named `send_message_to_thread`, wrapped in `<codex_delegation>`. It was not serialized as a user message. Revision selection ignored it, selected the older human prompt, observed the old prompt's `turn_id`, and returned the conflict before creating a browser turn.

This is a concrete reproduction of the error reported in #546 and also explains why an external goal monitor could not continue a completed task. It does not weaken the existing protection against replaying an old prompt after an unauthenticated restart.

## Testing

- Added a focused revision test covering the observed wire shape plus wrong-tool, malformed-output, and wrong-turn rejection.
- Added a server lifecycle regression proving an authenticated cross-thread follow-up reaches the adapter while the existing context-only restart test still returns HTTP 400.
- Ran the full pinned Bun 1.4.0 verification gate:
  - 738 runtime tests passed, 2 platform skips
  - 310 launcher tests passed
  - root and launcher dependency audits passed
  - root and launcher typechecks passed
  - launcher renderer build passed
  - runtime bundle, notices, and relocatable release smoke passed
- Built and installed the patched runtime locally under the real v5.0.8 Full harness, then retried the previously failing `send_message_to_thread` path. The new turn completed, executed one native `exec_command` (`pwd && git status --short`) with exit code 0, returned its result, and made no repository edits.

## Scope

This patch addresses the authenticated cross-thread delegation reproduction. It does not claim to resolve unrelated capacity, compaction-stall, or token-lifecycle reports that may surface similar user-facing symptoms.

Related: #546, #410


nofihq:
@codex review

miuuyy:
Thanks for identifying the native `send_message_to_thread` payload. I reproduced the turn-ID conflict with the installed Codex app-server and implemented an independent fix locally on the 6.0 branch.

The implementation uses the same instruction handling for task identity, environment resolution and compaction, validates the native `codex_app` envelope, and preserves the checks against replaying an unrelated older turn. The relevant suites pass, and an offline native Codex/bridge HTTP round-trip accepts the delegated task with the correct workspace.

Closing this PR in favor of that implementation. The fix is local and unreleased; live ChatGPT validation is still separate. Thanks for the useful report and investigation.

PR #596, opened 2026-09-20
fix: recognize the client's .codex visualization root on a bridge running on another host

felixlark:
## Problem

When the bridge runs on a different machine than Codex (remote `base_url`), any thread that has used the visualization feature is rejected with:

```
ChatGPT web turn is missing cwd in trusted Codex environment context
```

The environment block of such a thread declares an extra sandbox root, e.g. `/Users/me/.codex/visualizations/2026/09/16/<threadId>`. `environment.ts` recognizes that root by rebuilding the expected path from the **bridge's own** `CODEX_HOME`. On a remote bridge that is a different path (`/home/user/.codex-web-gpt-home`), so the root never matches, it is treated as an unknown root, and the whole environment block is discarded — which then surfaces as the missing-cwd error above.

This is deterministic: every desktop thread that has produced a visualization fails on a remote bridge.

## Change

When checking a declared root, also accept the client-side shape `<some .codex dir>/visualizations/<YYYY>/<MM>/<DD>/<threadId>`, matched structurally instead of against the bridge's own home. The two existing checks are unchanged:

- the date segments must be a well-formed date directory, and
- the last segment must be the current thread id or the parent thread's id.

So a root can still only be accepted for the thread that owns it; this does not widen which threads a root authorizes, only where the `.codex` directory may live.

## Tests

`tests/environment.test.ts`, three new cases:

- a client-side visualization root for the current thread is accepted,
- the same shape naming a *different* thread is rejected,
- the same date/thread shape outside any `.codex` directory is rejected.

Reverting the change makes the first case fail with the original missing-cwd rejection. `bun test tests/environment.test.ts` on top of current `main`: 64 pass, 1 skip, 0 fail.

## Notes

Rebased onto `main` (`eaf4f09`); the only conflict was the test file's import line.


miuuyy:
Thanks for the clear reproduction. We are declining this change because matching an arbitrary `.codex/visualizations/YYYY/MM/DD/<thread-id>` path does not establish that it belongs to the actual trusted Codex home. An isolated control also accepts a matching path below an unrelated temporary directory after this patch; checking the thread ID does not resolve that ownership gap.

The remote-bridge layout in this report is outside the supported local integration. Supporting it would require an explicit way to validate the client's native environment, rather than broadening trusted roots based on their spelling.

PR #597, opened 2026-09-20
fix: stabilize browser turns, observation recovery, and launcher lifecycle

WoofahRayetCode:
## Summary

This PR hardens ChatGPT Web turns against the failure modes that show up in long Full-harness sessions: temporary CDP/DOM drops, collapsed **Thinking failed** headers, offscreen effort-picker and layout clipping, completed-tab retention, account-wide tab fan-out, and narrow native setup retries. Recovery is deliberately **read-only** after Send — reconnect and rebind the same launcher tab; never resubmit the original prompt or replay tools.

### What changed

1. **Launcher offscreen reliability**
   - Hidden/automatic viewport floor is now **1120×720** (was 800×600); cramped visible windows under **800×600** keep the same emulation so the connector pill and Instant–Pro effort menu are not clipped.
   - Chromium background throttling is disabled via Electron prefs **and** process flags (`disable-renderer-backgrounding`, timer throttling, occluded-window backgrounding, `CalculateNativeWinOcclusion`).
   - Helper heartbeat timeout is **120s** so a 60s CDP rebind cannot look like a dead helper.
   - Completed temporary-chat tabs close after a **30s** grace (was a long retained TTL); running tasks, tool calls, and manual input keep their tabs.

2. **Observation / CDP recovery (no resend)**
   - Shared surface-rebind budget with full-jitter backoff; Pro gets a larger budget (up to 6) for long thinking.
   - Transient connection errors and stalled DOM probes retry and reattach to the same tab.
   - `selectLauncherPage` bounds each `Target.getTargetInfo` inspect (~250ms) so a hung peer cannot burn the whole acquisition window.
   - Owned-target pings during observation fail fast on CDP drop.

3. **Thinking failed + effort fallback**
   - Collapsed assistant **Thinking failed** headers are treated as structured retryable failures.
   - In-place regenerate/retry up to 3 times, then step down to the next cheaper selectable Sol effort and regenerate — still without resubmitting the Codex prompt.
   - Effort picker open/select/close hardened for offscreen Radix menus (pointerdown+up, tick labels, Escape/trigger fallback).

4. **Concurrency**
   - Process-wide queue at the **5-tab** account limit before exclusive page work, so ordinary fan-out waits instead of looking like spam upstream.

5. **Failed-turn continuation**
   - Narrow Codex rollout recovery: only an exact, unexecuted user instruction after native capacity / model-control / turn_id-conflict failures may continue. Abort, assistant output, or tool side effects still fail closed.

6. **Long-turn / multipart / packaging**
   - First-response grace and composer budgets scale with effort and prompt size; outstanding tool calls suppress false DOM-health stalls.
   - Instant multipart acks may be short paraphrases; six-part Bigger Context is preferred when two-part stages remain huge (~180k+ chars).
   - Compaction handoff deadlines reset on heartbeats/submission.
   - Linux AppImage packaging requires a prepared libnotify toolset and smoke-tests the required symbol.
   - Doctor treats “browser busy on a Codex turn” as a warning when the host is still live.
   - README documents observation retry and the 30s completed-tab cleanup.

### How things improve in practice

| Before | After |
| --- | --- |
| Temporary CDP/browser attach loss mid-turn often failed the whole Responses turn | Helper retries/rebinds the same tab (up to effort budget) and continues reading; prompt is not resent |
| ChatGPT showed **Thinking failed** while Stop still looked live; helper could heartbeat for a very long time | Header is detected, regenerate is attempted in-place, then cheaper effort fallback; structured retryable error if recovery is exhausted |
| Offscreen / small launcher windows clipped the connector mention and effort slider (e.g. ~710×522) | 1120×720 floor + layout emulation keeps Instant–Pro and connectors usable |
| Background Chromium throttling dropped painting/CDP targets for hidden tabs | Flags + `backgroundThrottling=false` keep targets attachable without opening the Browser tab |
| Completed temporary chats could linger a long time | Close after 30s grace; immediate follow-up can reuse, then a fresh chat opens with Codex task context |
| Many concurrent Codex requests could open more browser work than the account limit | Queue at 5 tabs; waiters abort cleanly |
| Capacity / model-control setup failures forced a hard stop even when Codex retried the same unexecuted instruction | Exact native failure + identical user item can continue; anything with assistant/tool output cannot |
| Submitted-turn failures could look retryable to Codex | Mapped to non-retryable `chatgpt_submitted_turn_failed` / `chatgpt_submission_ambiguous` so tools are not replayed |

### Example scenarios these fixes address

- **Pro / Extra High thinking past the old ingest window**: first-response grace and surface rebinds scale with effort; a brief CDP blip reconnects instead of aborting a long think.
- **“Connection interrupted” on chatgpt.com while the helper still owns the tab**: observation recovery treats attach/DOM stalls as transient reads; it does not treat the ChatGPT.com banner as a reason to resend.
- **Thinking failed after a huge multipart stage**: regenerate in the same chat; if Instant keeps collapsing, fall back Medium → High, etc., only when that effort is actually available on the account.
- **Doctor during an active turn**: no longer reports a hard browser-host error solely because session verification is deferred while a Codex turn is running.

## Test plan

- [x] Unit/contract coverage added or extended for observation retry, concurrency, failed-turn continuation, effort session helpers, browser-worker Thinking failed recovery, connector mention, debug logging, launcher host viewport/heartbeat/grace, and AppImage tool prep
- [ ] Run focused suites: observation-retry, concurrency, failed-turn-continuation, chatgpt-session, browser-worker-contract, launcher browser-host tests
- [ ] Manual: start a Pro turn, briefly disrupt CDP/attach, confirm rebind without prompt resend
- [ ] Manual: induce or wait for Thinking failed, confirm in-place regenerate / effort fallback
- [ ] Manual: complete a temporary-chat turn, confirm tab closes ~30s later and the next turn opens cleanly
- [ ] Manual: cramped launcher window — connector pill + Instant–Pro picker still usable
- [ ] Linux: prepare AppImage tools, package, confirm symbol smoke passes


miuuyy:
Thanks for the contribution. We are declining this PR because it combines too many behavioral changes and unverified recovery assumptions to establish that it improves reliability without regressions.

There are concrete blockers: it automatically lowers the user's selected reasoning effort, and the multipart acknowledgement check accepts short failure responses (including "I could not load any of the supplied context.") as successful context receipt. Retries, tab lifetime changes, concurrency limits, and compaction deadline changes are also bundled together, while the focused and live verification items remain unchecked.

Any future fix needs a specific reproduced failure, a narrow change that preserves the selected model/effort and context integrity, and validation of that exact behavior.

PR #598, opened 2026-09-20
feat(launcher): add an opt-in browser-host-only mode that hides bridge setup

felixlark:
## Problem

A launcher whose core home has no `config.json` still shows the full setup wizard. If the bridge for that ChatGPT profile is already running elsewhere — a `cli.js serve` started by another launcher, a service, or a second machine — the wizard's third step, **Install into Codex**, tries to start one here as well. It fails every time it is clicked:

```
Error invoking remote method 'launcher:setup-core': Error: codex-chatgpt-web:
Cannot bind 127.0.0.1:17841: listen EADDRINUSE: address already in use 127.0.0.1:17841
```

The window is otherwise doing useful work — hosting the ChatGPT browser for turns — but it permanently shows a "required" badge, a wizard it cannot finish, and a button whose only outcome is that error.

## Change

Opt in with `CODEX_WEB_GPT_BROWSER_HOST_ONLY=1`. `resolveLauncherProfile()` exposes `browserHostOnly`, the snapshot carries it, and the renderer then hides what this launcher cannot own:

- the Install into Codex step,
- the "restart Codex" notice,
- both MCP entry points (sidebar item and setup section),
- the "required" badge on Setup.

It opens on the browser surface rather than an unfinishable setup page, and the setup page is titled **Browser host** with one line saying the bridge runs as a separate service. Signing in to ChatGPT and the browser smoke test stay exactly as they are — signing in is what such a window is for.

Nothing is inferred: without the variable the launcher behaves exactly as before, including with a missing config (`"0"` is not opt-in either).

## Tests

- `launcher/tests/profile.test.cjs`: the flag is off by default, on for `"1"`, off for `"0"`, and it changes no path (core home and userData stay identical).
- `launcher/tests/renderer-wiring.test.cjs`: main exposes the flag, and each of the four hidden elements plus the two kept ones is asserted.

`bun run typecheck` and `node --test tests/*.test.cjs` on top of `main`: 311 pass, 1 skip, 0 fail.

## Context

I run one bridge with several launchers as browser hosts, so turns in one Chromium cannot stall another. Those hosts hit this wizard every time their window is opened. The mode is deliberately a plain environment flag so it stays out of the way of the normal single-launcher install.


miuuyy:
Thanks for the proposal. A browser-host-only launcher with a separately managed bridge is outside the deployment model we support. We are keeping the launcher responsible for its integrated runtime and setup, so I am closing this feature PR.

PR #600, opened 2026-09-20
Fixes Windows MCP Setup failures when installing `tunnel-client`.

xingxingshese:
## Summary

Fixes Windows MCP Setup failures when installing `tunnel-client`.

- Add bounded retries for Windows `EBUSY`/`EPERM` cleanup failures.
- Ensure staged tunnel-client files are closed before install and cleanup.
- Preserve the original setup error when rollback cleanup also fails.
- Reuse an existing valid `tunnel-client.exe` when possible.
- Clean up stale `.install-*.exe` files safely.
- Add regression tests for retry, rollback, cleanup, and existing binary reuse.

## Root Cause

Windows may briefly keep a lock on the downloaded executable after it is executed for version verification. The immediate cleanup then fails with `EBUSY`. Previously, the rollback cleanup error could replace the original setup failure.

## Verification

- `bun run verify`
- Windows package built successfully with electron-builder.
- Verified `tunnel-client.exe --version`.
- Verified no leftover `tunnel-client.exe.install-*.exe` files remain.


miuuyy:
Thanks for identifying the Windows cleanup failure. I have implemented a narrower fix locally for 6.0 and am closing this PR in favor of that implementation.

The stale-file sweep introduces another setup failure: an unrelated old `.install-*` file that cannot be deleted aborts the installation before downloading anything. I reproduced this with a protected fixture file on macOS, where the existing installer succeeds alongside that file. The sweep and adoption of untracked binaries also expand the installer behavior beyond the reported cleanup problem.

Our local fix bounds retries for Windows `EBUSY`/`EPERM` on the install's explicit cleanup paths and preserves the original error if cleanup or rollback also fails. It does not scan other staging files or change binary adoption. The targeted setup/runtime checks pass; native Windows validation and release are still pending. For similar isolated failures, an issue with logs and a reproduction is a useful starting point. Thanks for the report.

PR #602, opened 2026-09-20
fix: classify unavailable Pro effort explicitly

nofihq:
## Summary

- distinguish a missing Pro slider step from a generic model-control failure
- return a structured, non-retryable `chatgpt_pro_unavailable` rate-limit error
- preserve explicit model selection and state that no lower-effort fallback was used
- retain the localized ChatGPT tooltip detail when the disabled Pro row exposes one

## Verification

- `bun run verify` (737 core tests passed, 2 skipped; 310 launcher tests passed; audits, typechecks, builds, notices, and relocatable runtime smoke passed)
- installed-runtime reproduction with an exhausted Pro allowance returned the new actionable error and the outer supervisor classified it as an account cooldown

This keeps the existing fail-closed model-selection invariant while preventing callers from treating a temporarily absent Pro option as an unknown UI breakage.


nofihq:
@codex review

miuuyy:
Thanks. This branch classifies a missing Pro slider position as a quota error without checking the current account plan or requiring an explicit quota explanation from ChatGPT. That does not establish the narrower case of a confirmed Pro subscriber exhausting their allowance. Closing this classification change; 6.0 will retain the picker failure and ChatGPT's own explanation when available.

PR #603, opened 2026-09-20
fix: classify rate-limit dialogs between multipart stages

nofihq:
## Summary

- detect ChatGPT requests-too-quickly dialogs before mutating the next multipart composer
- return the existing structured 429 instead of a misleading prompt-attachment integrity failure
- cover both the classification boundary and multipart recovery fixture

## Observed failure

A six-part Extra High context accepted part 1, then ChatGPT opened its requests-too-quickly dialog before part 2. Because the modal owned focus, `execCommand("insertText")` returned false and the adapter surfaced `ChatGPT composer rejected the plain-text editing command`. The safe diagnostic showed one open dialog and focus on its button in two independent repositories.

## Verification

- `bun test tests/browser-worker-contract.test.ts` (132 pass)
- `bun test tests/compaction-browser-recovery.test.ts` (4 pass)
- full `scripts/verify.ts` on the same patch stack: 742 core tests passed, 2 skipped; 310 launcher tests passed; audits, typechecks, builds, and relocatable runtime smoke passed
- installed local runtime now reports ready in `doctor`


nofihq:
@codex review

nofihq:
Live Extra High reproduction found a second multipart boundary failure: after part 1 was acknowledged, ChatGPT rewrote the closed effort-control label (safe diagnostics changed from 7 to 10 characters) while the semantic control remained present. The part-2 send fence compared the stale label and aborted with `ChatGPT model controls are unavailable`.

Commit `5958581` now re-runs the existing semantic slider selection/proof between staged parts, preserving the strict model check instead of weakening it. Focused multipart/browser suite: 135 pass, 0 fail. @codex review

miuuyy:
Thanks for identifying both multipart failures. We have addressed them in the local 6.0 worker: rate-limit dialogs are checked before shared prompt attachment, and each subsequent part gets fresh model/effort proof while preserving model-family selection and Limits accounting. Closing this PR in favor of that implementation; it is not released yet.

PR #604, opened 2026-09-20
fix: preserve replacement broker sockets during shutdown

nofihq:
## Summary
- record the Unix socket device/inode owned by each turn broker
- protect a newer replacement socket while an older Bun server closes
- only unlink the exact socket instance owned by the closing broker
- add a regression test that replaces a live broker pathname before closing the old server

## Why
A daemon restart can briefly overlap old-process shutdown with a replacement listener. Bun removes the Unix socket pathname from `Server.close()`, so the old process could unlink the new daemon's `turn-broker.sock` while both process states looked healthy. Native tool calls then failed until another provider restart.

## Verification
- `bun test tests/turn-broker-lifecycle.test.ts` (14 pass)
- full integration-branch `bun run verify` with the same patch (743 core pass, 2 platform skips; 310 launcher pass; audits, typechecks, builds, and relocatable smoke pass)
- installed runtime doctor healthy and broker protocol v5 reachable after controlled restart


nofihq:
@codex review

miuuyy:
Thanks. The replacement-socket bug is real, and this patch preserves the replacement path after shutdown. However, our socket reproduction also shows that the replacement listener is unreachable with ENOENT while an old connection keeps shutdown pending: its live pathname has been renamed away. We cannot take a shutdown strategy that makes another runtime temporarily unreachable. Closing this implementation; the underlying ownership problem still needs a different fix.

Issue #443, opened 2026-09-11
[Bug] Error invoking remote method 'launcher:setup-core': Error: codex-chatgpt-web: Codex interrupt lifecycle hook changed after setup; refusing to overwrite it


guberm:
### Codex Web GPT version

5.0.6

### Codex version

26.908.40834

### Platform

Windows x64

### OS version

Windows 11

### ChatGPT account tier

Pro

### Integration mode

Browser-only

### Selected model

Setup

### What happened?

[codex-web-gpt-diagnostics-2026-09-11.json](https://github.com/user-attachments/files/32132022/codex-web-gpt-diagnostics-2026-09-11.json)

### Minimal reproduction

...

### Complete final error

```text
Error invoking remote method 'launcher:setup-core': Error: codex-chatgpt-web: Codex interrupt lifecycle hook changed after setup; refusing to overwrite it
```

### Privacy-safe diagnostics

<img width="431" height="106" alt="Image" src="https://github.com/user-attachments/assets/adf03a15-2bfc-4fc6-af75-d8655988ca96" />

### Other Codex wrappers or route owners

...

### Preflight checklist

- [x] I tested the latest release and, for catalog or turn failures, fully restarted Codex including its background process.
- [x] I stated whether the failure also occurs in a fresh Codex task, or explained why that does not apply.
- [x] I attached only a privacy-safe export, or explained why the exporter could not produce one; I removed any remaining private content.
- [x] I searched existing issues for this exact final error.

springerphillip16-hash:
Having the same issue rn

miuuyy:
I read the attached log: setup previously succeeded, and the hook verification starts failing on September 11. The export does not include the hook configuration itself, so it cannot distinguish a harmless TOML rewrite from an actual change to the managed command.

I already have a tested local fix for the harmless interleaved-table case in #431. It preserves unrelated Codex settings while still rejecting modified managed hooks. I cannot yet confirm that your configuration is that same case. Please share a small redacted excerpt around the `codex-chatgpt-web` hook markers and `[hooks...]` tables in `config.toml`, preserving the order of table headers and comments. Omit private values, credentials and authentication files. Please do not delete the hook or disable the integrity check as a workaround.

26OKbe:
I can reproduce this on Windows x64 with codex-chatgpt-web 5.0.8 and Codex CLI 0.154.0.

The integration journal and current Codex config appear semantically equivalent, but the TOML representation has been rewritten.

[...]

PR #613, opened 2026-09-21
fix: tolerate reordered Codex interrupt hook state

### LINKED ISSUE #612: doctor rejects semantically unchanged interrupt hook after Codex reorders trust state
## Summary

`doctor` and setup restoration reject an installed Codex interrupt hook after Codex rewrites `config.toml`, even though the managed hook command, timeout, trust key, and trusted hash are unchanged.

This is distinct from #443. That fix accepts equivalent quoted trust-state keys; it still rejects this ordering and boundary-whitespace rewrite.

## Environment

- macOS arm64
- released app: v5.0.8
- also reproduced against `main` at `eaf4f09`

## Reproduction

1. Install the managed interrupt hook with setup.
2. Let Codex rewrite `config.toml` so that the matching `[hooks.state."...:interrupt:0:0"]` table is grouped before `[[hooks.Interrupt]]`.
3. Let the writer normalize the blank line between `timeout = 3` and the managed end marker from two line endings to one.
4. Run `doctor` or verify the recorded hook against the rewritten config.

Minimal shape after the rewrite:

```toml
[hooks.state."/Users/example/.codex/config.toml:interrupt:0:0"]
trusted_hash = "sha256:<unchanged>"

# Managed by codex-chatgpt-web: release the exact Responses request when its Codex turn is interrupted.
[[hooks.Interrupt]]

[[hooks.Interrupt.hooks]]
type = "command"
command = "<unchanged installed command>"
timeout = 3
# End codex-chatgpt-web interrupt lifecycle hook.
```

The rewritten document parses to the same owned `hooks.Interrupt[0]` and `hooks.state[stateKey]` values as the installation journal.

## Actual result

```text
Codex interrupt lifecycle hook changed after setup; refusing to overwrite it
```

## Expected result

Verification succeeds when:

- the owned hook and trust-state definitions are unique,
- their parsed values exactly match the journal,
- the command hash and group index match,
- only serialization order and definition-boundary line endings changed.

Actual command, timeout, hash, duplicate definitions, or nested extensions must continue to fail closed.

## Proposed fix

Locate the hook and trust-state definitions independently, tolerate either serialization order, and use a boundary-normalized textual match only as a fallback. Keep the existing exact parsed-definition, marker, group-index, hash, uniqueness, and restoration checks.

A regression test covers LF and CRLF, the reordered trust state, normalized boundary blank line, successful restore, and negative command/hash/duplicate cases.

guanbear:
Fixes #612.

## What changed

- locate the managed Interrupt hook and its trust state independently
- accept either serialization order after Codex regroups `hooks.state` tables
- fall back to a boundary-normalized match when the native writer changes only surrounding line endings
- retain exact parsed-definition, marker, group-index, command-hash, uniqueness, and restoration checks
- add LF/CRLF regression coverage for the real rewrite shape and negative mutation cases

This complements #443: quoted trust-state key aliases remain supported, while this PR covers a semantically unchanged state table moved before the hook plus normalized boundary whitespace.

## Verification

- `bun run typecheck`
- `bun test tests/codex-interrupt-hook.test.ts tests/setup-lifecycle.test.ts` (17 pass)
- `bun test tests/zero-risk-adapter.test.ts` (11 pass)
- read-only verification against the affected real-world `config.toml` and integration journal succeeds; the restored document parses and contains no managed hook


miuuyy:
Thanks for the report and patch. This was already fixed independently in the local 6.0 work, including reordered hook trust state and strict ownership checks. Closing as covered by that unreleased change.

